PR #6, opened 2022-05-18
Cosmos-sdk v0.45.4 + wasmd

pr0n00gler:


foxpy (commented):
To be honest I cannot say anything other than "well, it seems to look good".

I don't even know how to review it.

I can only say that task seems to be completed? I mean, I can see a downgrade in `go.sum`, code compiles and tests complete successfully.

PR #9, opened 2022-06-06
fix: prepend consensus state

zavgorodnii:
this PR makes it possible to submit normal queries by fixing the requested consensus state height & fixing the revision number.

please note that the revision number should not be hardcoded, but this is out of the scope  of this PR


pr0n00gler (approved):
Would be nice to leave a comment why we should do +1

foxpy:
> Would be nice to leave a comment why we should do +1

I sign under these words

swelf19 (approved):

PR #10, opened 2022-06-09
sudo() messages to pass data from Go code to CosmWasm #LSC-31

ratik:
https://p2pvalidator.atlassian.net/browse/LSC-31

Make sure you are testing this one with proper branch of contracts (feat/sudo)

https://github.com/lidofinance/lido-interchain-staking-contracts/pull/26

Please note that as for now ibc returns only one error for interchain tx errors: `ABCI code: 5: error handling packet on host chain: see events for details` so we may want to retrieve actual error from events bc acknowledge data just don't have any in it itself


register-interchain-account.sh
```diff
@@ -0,0 +1,5 @@
+export DEMOWALLET_1=$(gaia-wasm-zoned keys show demowallet1 -a --keyring-backend test --home ./data/test-1) && echo $DEMOWALLET_1;
+export DEMOWALLET_2=$(gaia-wasm-zoned keys show demowallet2 -a --keyring-backend test --home ./data/test-2) && echo $DEMOWALLET_2;
+
+gaia-wasm-zoned tx interchaintxs register-interchain-account connection-0 cosmos1mjk79fjjgpplak5wq838w0yd982gzkyfrk07am \
```
foxpy: I would move `cosmos1mjk79fjjgpplak5wq838w0yd982gzkyfrk07am` into a named constant to provide better understanding on purpose of this account

x/interchaintxs/client/cli/tx.go
```diff
@@ -91,8 +92,20 @@ func SubmitTxCmd() *cobra.Command {
...
 				}
 
-				if err := cdc.UnmarshalInterfaceJSON(contents, &txMsgs); err != nil {
-					return fmt.Errorf("error unmarshalling sdk msgs file: %w", err)
+				var rawTxMsgs struct {
+					Msgs []json.RawMessage `json:"msgs"`
+				}
+
+				if err := json.Unmarshal(contents, &rawTxMsgs); err != nil {
+					panic(err)
```
foxpy: Is it really okay to panic here (I doubt)?
If it is okay to panic, then feel free to resolve this thread.
If it is not okay, please leave a TODO comment to deal with it later.

x/interchaintxs/client/cli/tx.go
```diff
@@ -91,8 +92,20 @@ func SubmitTxCmd() *cobra.Command {
...
+				}
+
+				for _, txMsg := range rawTxMsgs.Msgs {
+					var sdkMsg sdk.Msg
+					if err := cdc.UnmarshalInterfaceJSON(txMsg, &sdkMsg); err != nil {
+						panic(err)
```
foxpy: and same here

foxpy (changes requested):

register-interchain-account.sh
```diff
@@ -0,0 +1,6 @@
+export DEMOWALLET_1=$(gaia-wasm-zoned keys show demowallet1 -a --keyring-backend test --home ./data/test-1) && echo $DEMOWALLET_1;
+export DEMOWALLET_2=$(gaia-wasm-zoned keys show demowallet2 -a --keyring-backend test --home ./data/test-2) && echo $DEMOWALLET_2;
+export OWNER=$(gaia-wasm-zoned keys list  --keyring-backend test --home ./data/test-1 --output json | jq -r -c '.[] | select(.name == "rly1") | .address') && echo $OWNER;
```
mmikev: Why not just `gaia-wasm-zoned keys show rly1 -a --keyring-backend test --home ./data/test-1`?

send_test_tx.sh
```diff
@@ -0,0 +1,5 @@
+export OWNER=$(gaia-wasm-zoned keys list  --keyring-backend test --home ./data/test-1 --output json | jq -r -c '.[] | select(.name == "rly1") | .address') && echo $OWNER;
```
mmikev: `gaia-wasm-zoned keys show rly1 -a --keyring-backend test --home ./data/test-1`?

x/interchaintxs/keeper/ibc_handlers.go
```diff
@@ -93,7 +108,29 @@ func (k *Keeper) HandleChanOpenAck(
...
+	messageType SudoMessageType,
+	contractAddress sdk.AccAddress,
+	request channeltypes.Packet,
+	msg []byte,
+	error []byte,
+) ([]byte, error) {
+
+	k.Logger(ctx).Info("Sudo", "contractAddress", contractAddress, "request", request, "msg", msg)
+
+	m, err := json.Marshal(SudoMessage{
```
NeverHappened: Not sure if it belongs here, but question to everybody:
How can we then find in smartcontracts what zone did we get response for?
I believe that matching by connection_id is not the easy way)

zavgorodnii: That's a good question (not just the connection id, connection id + port id, i guess?..). 

But:

1. Is it really that hard?
2. What's a better way to do it? Put the zone id to tx memo?

foxpy (approved):

pr0n00gler (approved):
Would be nice to have couple words in readme about how to test/see in work this PR (if it works of course)

x/interchaintxs/client/cli/tx.go
```diff
@@ -91,8 +92,20 @@ func SubmitTxCmd() *cobra.Command {
 					return fmt.Errorf("json input was not provided; failed to read file with tx messages: %w", err)
 				}
 
-				if err := cdc.UnmarshalInterfaceJSON(contents, &txMsgs); err != nil {
```
sotnikov-s, line 94: I know it's out of scope, but could you please fix the short description of this cobra command above?

x/interchaintxs/keeper/ibc_handlers.go
```diff
@@ -1,11 +1,29 @@
...
 )
 
+type SudoMessageType uint
+
+const (
+	OpenAck SudoMessageType = iota
+	Response
+	Timeout
+	Error
+)
```
sotnikov-s: please add a `SudoMsgType...` prefix to the constants. otherwise, for example, a standalone `Error` looks strange in the code. it's a good widespread practice — to prefix enums-like constants with the type they represent.
```golang
_, err = k.Sudo(ctx, Error, hubContractAddress, packet, []byte{}, []byte(errorText))
```

x/interchaintxs/types/tx.go
```diff
@@ -101,3 +101,15 @@ func PackTxMsgAny(sdkMsg sdk.Msg) (*codectypes.Any, error) {
 
 	return any, nil
 }
+
+func (msg MsgSubmitTx) UnpackInterfaces(unpacker codectypes.AnyUnpacker) error {
```
sotnikov-s, line 110: unused func?

x/interchaintxs/keeper/ibc_handlers.go
```diff
@@ -93,7 +108,29 @@ func (k *Keeper) HandleChanOpenAck(
...
-func (k *Keeper) Sudo(ctx sdk.Context, contractAddress sdk.AccAddress, msg []byte) ([]byte, error) {
-	//return k.wasmKeeper.Sudo(ctx, contractAddress, msg)
-	return nil, nil
+func (k *Keeper) Sudo(
+	ctx sdk.Context,
+	messageType SudoMessageType,
+	contractAddress sdk.AccAddress,
+	request channeltypes.Packet,
+	msg []byte,
+	error []byte,
```
sotnikov-s: please rename the parameter to err because it looks confusing in the code editor. it's a common practice to shorten error variables to err
image

x/interchaintxs/keeper/ibc_handlers.go
```diff
@@ -14,16 +32,20 @@ func (k *Keeper) HandleAcknowledgement(ctx sdk.Context, packet channeltypes.Pack
 		return sdkerrors.Wrap(err, "failed to GetHubAddress")
 	}
 
-	acknowledgementPacketBz, err := packet.Marshal()
-	if err != nil {
-		k.Logger(ctx).Error("failed to marshal acknowledgement packet", "error", err, "sequence", packet.Sequence, "dst_channel",
-			packet.DestinationChannel, "dst_port", packet.DestinationPort, "height", packet.TimeoutHeight)
-		return sdkerrors.Wrap(err, "failed to marshal Timeout packet")
+	var ack channeltypes.Acknowledgement
+	if err := channeltypes.SubModuleCdc.UnmarshalJSON(acknowledgement, &ack); err != nil {
```
sotnikov-s, line 50: the SubModuleCdc description claims:
SubModuleCdc references the global x/ibc/core/04-channel module codec. Note, **the codec should ONLY be used in certain instances of tests** and for JSON encoding.

can we use it actually? is there a production ready alternative?

x/interchaintxs/keeper/ibc_handlers.go
```diff
@@ -14,16 +32,20 @@ func (k *Keeper) HandleAcknowledgement(ctx sdk.Context, packet channeltypes.Pack
...
+		return sdkerrors.Wrapf(sdkerrors.ErrUnknownRequest, "cannot unmarshal ICS-27 packet acknowledgement: %v", err)
```
sotnikov-s, line 51: please use %w for returned errors

x/interchaintxs/keeper/ibc_handlers.go
```diff
@@ -77,7 +92,7 @@ func (k *Keeper) HandleChanOpenAck(
 	//  and it should be easy for the contract to parse it. I don't want to use JSON (it's super ugly
 	//  in this context); maybe we should generate a separate proto-message that will reference the
 	//  channeltypes.Packet?
-	_, err = k.Sudo(ctx, hubContractAddress, []byte(portID+channelID+counterPartyChannelId+counterpartyVersion))
+	_, err = k.Sudo(ctx, OpenAck, hubContractAddress, channeltypes.Packet{}, []byte(portID+channelID+counterPartyChannelId+counterpartyVersion), []byte{})
```
sotnikov-s: I don't really like the way the Sudo func is designed: not all parameters are used which results in empty arguments passed. what do you think about replacing the keeper's Sudo method with four methods — one on each SudoMessageType and with only the required parameters? e.g. execSudoOpenAck, execSudoError etc. it will also become possible to make logging more precise and concise in these methods

x/interchaintxs/keeper/ibc_handlers.go
```diff
@@ -1,11 +1,29 @@
...
 
 import (
+	"encoding/json"
+
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
 	channeltypes "github.com/cosmos/ibc-go/v3/modules/core/04-channel/types"
 )
 
+type SudoMessageType uint
```
sotnikov-s: how do you find making the SudoMessageType of type string with more detailed values? it will make it possible to use this type in logging which will result in a better UX for us

sotnikov-s (changes requested):

sotnikov-s (commented):
_ignore_

test_tx.json
```diff
@@ -0,0 +1,17 @@
...
+      "@type": "/cosmos.staking.v1beta1.MsgDelegate",
+      "delegator_address": "cosmos1awzs6zxuyyjkqutvmlj8f9ygjjgpwnhfsnz359c23xlg5crye3xsj62589",
+      "validator_address": "cosmosvaloper1qnk2n4nlkpw9xfqntladh74w6ujtulwnmxnh3k",
+      "amount": {
+        "denom": "stake",
+        "amount": "1000"
+      }
+    }
+  ]
+}
```
sotnikov-s: if I understand it correctly, the .sh files you've introduced and this json are made for tests. I think it would be more clean to move them to a dedicated folder and maybe even add a makefile in the root of the project that executes the .sh files in the order you need. what do you think?

x/interchaintxs/types/tx.go
mmikev: It's actually implements [`UnpackInterfacesMessage.UnpackInterfaces`](https://github.com/cosmos/cosmos-sdk/blob/d07d35f29e0a0824b489c552753e8798710ff5a8/codec/types/interface_registry.go#L60).

Would be good to add a comment about this, I agree.

And add 
```go
var (
	_ codectypes.UnpackInterfacesMessage = MsgSubmitTx{}
)

```

at the top of the file to check that the message satisfies the interface

sotnikov-s: ... and add a comment describing this idea to the empty interface check variable (they are sometimes confusing)

x/interchaintxs/keeper/ibc_handlers.go
ratik: not sure we really need to do it bc we already have `err` var in fn scope which is returned. so me personally prefer to keep err var name for returned var

x/interchaintxs/keeper/ibc_handlers.go
ratik: this peace of code been borrowed from interchain-demo, also I checked other ways and hadn't found any alternatives

x/interchaintxs/keeper/ibc_handlers.go
ratik: can't agree, this way we gonna have several copypasted functions which do almost the same stuff

x/interchaintxs/types/tx.go
ratik: fixed

ratik:
> Would be nice to have couple words in readme about how to test/see in work this PR (if it works of course)

sure will do, when we have it ready on contract side. As for now you have it's kinda complex to actually test

test_tx.json
ratik: sure, I think it would be nice to have separate task for it bc there are other test-related stuff to move

x/interchaintxs/keeper/ibc_handlers.go
mmikev: Actually, `sdkerrors.Wrapf()` does not support `%w`

ratik: oh nice

x/interchaintxs/keeper/ibc_handlers.go
ratik: so we had a conversation with @swelf19 and @NeverHappened and figured out that this info stored in hub storage so we could find proper zone on contract side

x/interchaintxs/keeper/ibc_handlers.go
sotnikov-s: yeah, my bad!

x/interchaintxs/keeper/ibc_handlers.go
sotnikov-s: okay, errorMsg then?
and also: the parameter type is []byte whereas you
1) pass string arguments to this func calls;
2) cast parameters back to string.
so I'd make the parameters string and rename error to errorMsg

x/interchaintxs/keeper/ibc_handlers.go
sotnikov-s: matter of taste :) but okay, I don't insist ofc

x/interchaintxs/keeper/ibc_handlers.go
sotnikov-s: moreover, matching on the uint type on the smart contracts side that possibly may be messed in the future is definitely worth switching to string type imo

foxpy (approved):

x/interchaintxs/keeper/ibc_handlers.go
```diff
@@ -8,20 +8,34 @@ import (
...
-	Message     string              `json:"result,omitempty"`
-	Error       string              `json:"error,omitempty"`
+type SudoMessageOpenAck struct {
+	OpenAck OpenAckDetails `json:"open_ack"`
+}
+
+type OpenAckDetails struct {
+	PortID                string `json:"port_id"`
+	ChannelID             string `json:"channel_id"`
+	CounterPartyChannelId string `json:"counter_party_channel_id"`
```
sotnikov-s: CounterPartyChannelId -> CounterpartyChannelId
counter_party_channel_id -> counterparty_channel_id
right?

sotnikov-s (approved):
a little misprint. the rest LGTM

demo/init_contracts.sh
```diff
@@ -0,0 +1,61 @@
...
+
+cd $(dirname "$0")
+
+OWNER=$(gaia-wasm-zoned keys show rly1 -a --keyring-backend test --home ../data/test-1);
+
+echo "Owner" $OWNER
+echo "Depoy Hub:" 
+TX_HASH=`../build/gaia-wasm-zoned tx wasm store ./../../lido-interchain-staking-contracts/artifacts/lido_interchain_hub.wasm --chain-id test-1 --from demowallet1 --gas 20000000 --gas-adjustment 1.4 --gas-prices 0.5stake --keyring-backend test --home ../data/test-1 --node tcp://127.0.0.1:16657 -y | grep txhash | cut -d " " -f 2`
+echo "Tx hash:" $TX_HASH
+sleep 5;
```
mmikev, line 11: I think we can avoid `sleep` calls if commands to send transactions contain `broadcast-mode block` flag?

mmikev: And `gaia-wasm-zoned` has `--output json` flag to print output in json, so you can use `jq` instead of greps, which can be more convenient

mmikev: Ah, I see you use that in other commands

pr0n00gler (approved):

zavgorodnii (approved):

NeverHappened (approved):

foxpy (approved):

PR #11, opened 2022-06-15
Feat/tx memo

NeverHappened:
https://p2pvalidator.atlassian.net/browse/LSC-39

See the https://github.com/lidofinance/lido-interchain-staking-contracts/pull/29 for contracts side of memo addition.

Added memo to SubmitTx message and to acknowledgement packet. I've tested it and it returns the memo back.

Test steps:
Use the ./gaia-wasm-zone/demo/readme.md to run test steps, but also need to add logging on acknowledgement to see memo returned back


proto/interchaintxs/v1/tx.proto
```diff
@@ -52,7 +52,8 @@ message MsgSubmitTx {
   // capability for.
   string owner = 2;
   string connection_id = 3;
-  repeated google.protobuf.Any msgs = 4;
+  string memo = 4;
+  repeated google.protobuf.Any msgs = 5;
```
foxpy: Please don't reorder protobuf message fields, they are sequentially enumerated for a good reason.

This is probably not going to cause any issues for our project (though I cannot prove it), but still.

Better safe than sorry.

foxpy (changes requested):

proto/interchaintxs/v1/tx.proto
NeverHappened: Good point, actually It didn't work that way, and I redid it

zavgorodnii:
only one consideration — maybe we shouldn't make the memo a required command line argument?.. i mean, not everyone needs a memo

x/interchaintxs/client/cli/tx.go
```diff
@@ -81,6 +81,7 @@ func SubmitTxCmd() *cobra.Command {
 			connectionID := args[0]
 			owner := args[1]
 			pathToMsgs := args[2]
+			memo := args[3]
```
zavgorodnii: I don't get it. You set MinimumNArgs(3), and here you always take the 4-th element of an array that might not have 4 elements.

ratik: btw, should we check memo length here or somewhere else?

NeverHappened: probably should in the message server handler. I also want to find where the real limit on memo is set in code

ratik: https://github.com/cosmos/cosmos-sdk/blob/6f070623741fe0d6851d79ada41e6e2b1c67e236/x/auth/types/params.go#L13

zavgorodnii (approved):

pr0n00gler (approved):

ratik (approved):

x/interchaintxs/client/cli/tx.go
```diff
@@ -67,10 +67,10 @@ func RegisterInterchainAccountCmd() *cobra.Command {
 
 func SubmitTxCmd() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:     "submit-tx [connection-id] [owner] [path/to/sdk_msgs.json]",
+		Use:     "submit-tx [connection-id] [owner] [path/to/sdk_msgs.json] [memo]",
```
foxpy: Hmmm, this help message implies (people usually use square brackets to mark optional arguments) that connection-id, owner, path and memo are all optional arguments, but they are not. I am providing my own suggestion but I am not really sure about it neither.

```suggestion
		Use:     "submit-tx connection-id owner path/to/sdk_msgs.json [memo]",
```

foxpy (changes requested):

x/interchaintxs/client/cli/tx.go
mmikev: It's a common practice in Cosmos-SDK to wrap arguments in square brackets. So I'd stick with current version

mmikev: But I'd personally pass memo via flag, not argument. This shows that memo is optional

foxpy: Well, if it is a common practice, then I am leaving an approving review. And at the same time anyone is free to change arguments hint if desirable.

foxpy (approved):

x/interchaintxs/client/cli/tx.go
NeverHappened: Changed to flag, did I do it ok?)

swelf19 (approved):

zavgorodnii (approved):
i have no time to test it like i did the last time, but i really hope that this works.

x/interchaintxs/client/cli/tx.go
foxpy: looks alright, though @pr0n00gler told us that:
> It's a common practice in Cosmos-SDK to wrap arguments in square brackets
so I am not 100% sure.

@pr0n00gler what do you think about it?

mmikev: I think we should return square brackets here for consistency

x/interchaintxs/client/cli/tx.go
```diff
@@ -67,7 +69,7 @@ func RegisterInterchainAccountCmd() *cobra.Command {
 
 func SubmitTxCmd() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:     "submit-tx [connection-id] [owner] [path/to/sdk_msgs.json]",
+		Use:     "submit-tx [connection-id] [owner] [path/to/sdk_msgs.json] --memo memo",
```
sotnikov-s: this seems to be redundant cause you add the flag at the line 143

x/interchaintxs/client/cli/tx.go
```diff
@@ -16,6 +16,8 @@ import (
 	"github.com/lidofinance/gaia-wasm-zone/x/interchaintxs/types"
 )
 
+const Memo = "memo"
```
sotnikov-s: please add a Flag suffix to the constant name cuz otherwise it's misleading what is the `Memo` in the code

sotnikov-s (changes requested):

sotnikov-s (approved):

PR #14, opened 2022-07-04
Feat/timeout call

NeverHappened:
https://p2pvalidator.atlassian.net/browse/LSC-33

Implement dispatch of timeout (and not only timeout) to the handler with needed arguments (zone_id, operation)

Implementation of timeout dispatch in contracts pr: https://github.com/lidofinance/lido-interchain-staking-contracts/pull/30/files


NeverHappened:
Not needed cause we're gonna pass through connectionId and operation another way

PR #18, opened 2022-07-14
feat: IBC transfer acknowledgement implemented

swelf19:
implements https://p2pvalidator.atlassian.net/jira/software/projects/LSC/boards/121?selectedIssue=LSC-87

Implemented a wrapper for original ibc transfer module witch handles an ack packet and call a sudo handler with appropriate payload.


go.mod
```diff
@@ -128,7 +128,6 @@ require (
 )
 
 replace (
-	// github.com/CosmWasm/wasmd v0.27.0 => ../wasmd
 	github.com/CosmWasm/wasmd v0.27.0 => github.com/swelf19/wasmd v0.27.1-0.20220705054332-4f6cf43be829
```
ratik: maybe move this one to neutron org?

x/sudo/sudo.go
```diff
@@ -0,0 +1,174 @@
...
+}
+
+func NewSudoHandler(wasmKeeper *wasm.Keeper, moduleName string) SudoHandler {
+	return SudoHandler{
+		moduleName: moduleName,
+		wasmKeeper: wasmKeeper,
+	}
+}
+
+// Sudo allows privileged access to a contract. This can never be called by an external tx, but only by
```
zavgorodnii: don't think this comment should be here

x/sudo/sudo.go
```diff
@@ -0,0 +1,174 @@
...
+//
+func (s *SudoHandler) SudoResponse(
+	ctx sdk.Context,
+	contractAddress sdk.AccAddress,
+	request channeltypes.Packet,
+	msg []byte,
+) ([]byte, error) {
+	s.Logger(ctx).Info("SudoResponse", "contractAddress", contractAddress, "request", request, "msg", msg)
+
+	// basically just for unit tests right now. But i think we will have the same logic in the production
```
zavgorodnii: please make this a todo so that we can find this place easily during final clean up

x/sudo/sudo.go
```diff
@@ -0,0 +1,174 @@
...
+}
+
+func (s *SudoHandler) SudoTimeout(
+	ctx sdk.Context,
+	contractAddress sdk.AccAddress,
+	request channeltypes.Packet,
+) ([]byte, error) {
+	s.Logger(ctx).Info("SudoTimeout", "contractAddress", contractAddress, "request", request)
+
+	// basically just for unit tests right now. But i think we will have the same logic in the production
```
zavgorodnii: same here

x/transfer/module.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+// OnAcknowledgementPacket implements the IBCModule interface.
+func (im IBCModule) OnAcknowledgementPacket(
+	ctx sdk.Context,
+	packet channeltypes.Packet,
+	acknowledgement []byte,
+	relayer sdk.AccAddress,
+) error {
+	err := im.IBCModule.OnAcknowledgementPacket(ctx, packet, acknowledgement, relayer)
```
zavgorodnii, line 42: maybe we should make it clear which handler returned the error here? (e.g., wrap the error)

ratik:
can we have a bit of description here and link to the task?

zavgorodnii (approved):

pr0n00gler (commented):
I don't really like the idea of putting `sudo` and `transfer` in `/x` folder. That is a place for full Cosmos-SDK modules (with keepers, msg-handlers, etc). In our case `sudo` and `transfer` are just helpers for our `interchaintxs` module.

pr0n00gler (commented):
I may be wrong about `transfer` because it's a wrapper around the IBC transfer module, but `sudo` definitely should be moved to another place

pr0n00gler (commented):
Could you add more comments for the `transfer` module?
Because it just sits in `/x` folder, but it's not clear for strangers at all what is it, why we need this, what it does, etc.

Would be good if `sudo` has comment coverage too. Basic at least

pr0n00gler (approved):

PR #20, opened 2022-07-20
feat: updated querySubmittedTransactions request arguments

swelf19:
implements https://p2pvalidator.atlassian.net/jira/software/projects/LSC/boards/121?selectedIssue=LSC-95

Changed query parameters for QuerySubmittedTransactionsRequest.
start-end request params replaced with start-limit request params


proto/interchainqueries/query.proto
```diff
@@ -77,7 +77,7 @@ message QueryRegisteredQueryResultResponse {
 message QuerySubmittedTransactionsRequest {
   uint64 query_id = 1;
   uint64 start = 2;
-  uint64 end = 3;
+  uint64 limit = 3;
```
foxpy, line 80: I don't think it is a good idea to reuse protobuf sequential number for a different field.

x/interchainqueries/client/cli/query.go
```diff
@@ -140,15 +140,15 @@ func CmdQueryTransactionsSearchResult() *cobra.Command {
 				return fmt.Errorf("failed to parse start: %w", err)
 			}
 
-			end, err := strconv.ParseUint(args[2], 10, 64)
+			limit, err := strconv.ParseUint(args[2], 10, 64)
```
foxpy, line 143: That doesn't look like an optional argument. Or do you imply `limit=0` being an unset limit? That's odd.

x/interchainqueries/keeper/keeper.go
```diff
@@ -242,9 +246,17 @@ func (k Keeper) GetQueryResultByID(ctx sdk.Context, id uint64) (*types.QueryResu
 }
 
 // GetSubmittedTransactions returns a list of transactions from start ID to end ID
-func (k Keeper) GetSubmittedTransactions(ctx sdk.Context, queryID uint64, start uint64, end uint64) ([]*types.Transaction, error) {
+func (k Keeper) GetSubmittedTransactions(ctx sdk.Context, queryID uint64, start uint64, limit uint64) ([]*types.Transaction, error) {
+	if limit == 0 {
+		limit = 100
```
foxpy, line 251: But there is a constant `x/interchainqueries/keeper.DEFAULT_SUBMITTED_TRANSACTIONS_LIMIT`

x/interchainqueries/keeper/keeper_test.go
```diff
@@ -566,12 +567,22 @@ func (suite *KeeperTestSuite) TestQueryTransactions() {
 	}
 	queriesKeeper.SetLastSubmittedTransactionIDForQuery(ctx, queryID, lastID)
 
-	start, end = 3, 8
+	start = 3
```
foxpy, line 570: Please set a limit here, too, just for clarity
```suggestion
	start, limit = 3, 5
```

foxpy (changes requested):
`make build` and `make test` fail with `missing go.sum entry` error messages. Please fix this (just run corresponding `go get`'s and commit `go.sum` changes)

proto/interchainqueries/query.proto
mmikev: Why? We are in active development, it's totally ok I think.

But we definitely can't do that so freely after release because it will break backward compatibility

foxpy: :thinking: okay, you are right

zavgorodnii:
i think what we call `start, limit` is usually called `offset, limit`?

x/interchainqueries/keeper/keeper.go
```diff
@@ -18,6 +17,11 @@ import (
 	"github.com/neutron-org/neutron/x/interchainqueries/types"
 )
 
+const (
+	DEFAULT_SUBMITTED_TRANSACTIONS_LIMIT   = 100
```
zavgorodnii, line 21: I don't think that Go uses all caps snake case for constants : ) looks like you've been coding in Rust lately hehe

x/interchainqueries/keeper/keeper.go
```diff
@@ -242,9 +246,17 @@ func (k Keeper) GetQueryResultByID(ctx sdk.Context, id uint64) (*types.QueryResu
 }
 
 // GetSubmittedTransactions returns a list of transactions from start ID to end ID
-func (k Keeper) GetSubmittedTransactions(ctx sdk.Context, queryID uint64, start uint64, end uint64) ([]*types.Transaction, error) {
+func (k Keeper) GetSubmittedTransactions(ctx sdk.Context, queryID uint64, start uint64, limit uint64) ([]*types.Transaction, error) {
+	if limit == 0 {
```
sotnikov-s, line 250: the whole limit validation and normalisation can be simplified to:
```go
switch {
case limit == 0:
	limit = DEFAULT_SUBMITTED_TRANSACTIONS_LIMIT
case limit > MAX_SUBMITTED_TRANSACTIONS_QUERY_LIMIT:
	return nil, sdkerrors.Wrapf(sdkerrors.ErrInvalidRequest, "failed to query submitted transactions, max query limit %d, got %d", MAX_SUBMITTED_TRANSACTIONS_QUERY_LIMIT, limit)
}
```
(with consts camelCased ofc)

x/interchainqueries/keeper/keeper.go
```diff
@@ -242,9 +246,17 @@ func (k Keeper) GetQueryResultByID(ctx sdk.Context, id uint64) (*types.QueryResu
 }
 
 // GetSubmittedTransactions returns a list of transactions from start ID to end ID
```
sotnikov-s, line 248: it's end ID in the comment still. please update

x/interchainqueries/client/cli/query.go
sotnikov-s: maybe a --limit flag would be better? looks like a way to make optional parameters for cli

sotnikov-s (changes requested):

PR #21, opened 2022-07-20
feat: tx query result callback

zavgorodnii:
This PR introduces a lot of changes:

1. The `SudoTxQueryResult()` callback is added. This callback is executed when a transaction result is submitted (unconditionally, every time);
2. Results for Tx queries **are not saved to state anymore**. Only the fact that a certain result (query_id + tx_hash) was submitted is saved, and duplicate results **are discarded without an error**;
3. Everything related to `QueryTransactions()` was removed.

**HOW TO TEST THIS CODE?**

Please have a look at this PR: https://github.com/neutron-org/neutron-contracts/pull/6
You can find the instructions in the readme: https://github.com/neutron-org/neutron-contracts/blob/feat/check-tx-query-result-callback/contracts/neutron_interchain_queries/README.md

**UPD:**

After a long discussion with @pr0n00gler we decided to limit the max number of transactions per submission to exactly 1.

Please re-review this PR, and please note that you now have to test it with this version of the relayer: https://github.com/neutron-org/cosmos-query-relayer/pull/16


x/interchainqueries/keeper/keeper.go
```diff
@@ -149,64 +147,21 @@ func clearQueryResult(result *types.QueryResult) types.QueryResult {
...
-	if err := k.UpdateLastRemoteHeight(ctx, queryID, uint64(maxHeight)); err != nil {
-		return sdkerrors.Wrapf(err, "failed to update last remote height for a query with id %d: %v", queryID, err)
-	}
-
-	if err := k.UpdateLastLocalHeight(ctx, queryID, uint64(ctx.BlockHeight())); err != nil {
-		return sdkerrors.Wrapf(err, "failed to update last local height for a query with id %d: %v", queryID, err)
-	}
-
-	k.SetLastSubmittedTransactionIDForQuery(ctx, queryID, lastSubmittedTxID)
+// SaveTransactionAsSubmitted simply stores a key (SubmittedTxKey + bigEndianBytes(queryID) + bigEndianBytes(txID)) with
```
mmikev: We don't have txID anymore

x/interchainqueries/keeper/keeper.go
```diff
@@ -149,64 +147,21 @@ func clearQueryResult(result *types.QueryResult) types.QueryResult {
...
-
+func (k Keeper) CheckTransactionIsAlreadySubmitted(ctx sdk.Context, queryID uint64, txHash []byte) bool {
 	store := ctx.KVStore(k.storeKey)
-	key := types.GetSubmittedTransactionIDForQueryKey(queryID, txID)
-
-	store.Set(key, txBz)
+	key := types.GetSubmittedTransactionIDForQueryKey(queryID, txHash)
 
-	return nil
+	return len(store.Get(key)) > 0
```
mmikev: Why not `return store.Has(key)`?

pr0n00gler (changes requested):
Keeper's methods `GetLastSubmittedTransactionIDForQuery` and `SetLastSubmittedTransactionIDForQuery` are unused now because of removed tx logic. Let's remove them too

foxpy:
FYI: there is a typo in branch name :ghost:

x/interchainqueries/keeper/keeper.go
zavgorodnii: lol true.

zavgorodnii:
> FYI: there is a typo in branch name 👻

chachlo popyachtsa

x/interchainqueries/keeper/process_block_results.go
```diff
@@ -101,17 +102,38 @@ func (k Keeper) VerifyBlock(ctx sdk.Context, clientID string, block *types.Block
...
+			if err = k.verifyTransaction(tmHeader, tmNextHeader, tx); err != nil {
+				ctx.Logger().Debug("ProcessBlock: failed to verifyTransaction",
+					"error", err, "query_id", queryID, "tx_hash", hex.EncodeToString(txHash))
+				return sdkerrors.Wrapf(types.ErrInternal, "failed to verifyTransaction %s: %v", hex.EncodeToString(txHash), err)
+			}
+
+			k.SaveTransactionAsSubmitted(ctx, queryID, txHash)
+
+			// Let the query owner contract process the query result.
+			if _, err := k.sudoHandler.SudoTxQueryResult(ctx, queryOwner, queryID, tmHeader.Header.Height, tx.Data); err != nil {
```
mmikev: What do you think about calling sudo once for the whole block (batch of txs), not for each transaction every time?

zavgorodnii: what would be the benefit?

mmikev: Gas consumption would be less.
Every wasm handler (Execute, [Sudo](https://github.com/CosmWasm/wasmd/blob/d63bea442bedf5b3055f3821472c7e6cafc3d813/x/wasm/keeper/keeper.go#L460), Migrate) instantiates a contract code on every call and consumes gas for such setup (unless contact's code is not pinned into wasmvm cache). As well as other micro reads from the storage

foxpy: Though that would require a sane limit for amount of transactions per batch.

zavgorodnii: After a long discussion with @pr0n00gler we decided to limit the max number of transactions per submission to exactly 1.

Please re-review this PR, and please note that you now have to test it with this version of the relayer: https://github.com/neutron-org/cosmos-query-relayer/pull/16

x/interchainqueries/keeper/msg_server.go
```diff
@@ -73,36 +77,54 @@ func (k msgServer) SubmitQueryResult(goCtx context.Context, msg *types.MsgSubmit
...
 		})
 		if err != nil {
+			ctx.Logger().Debug("SubmitQueryResult: failed to get ConnectionConsensusState",
+				"query", query, "message", msg)
 			return nil, sdkerrors.Wrapf(ibcclienttypes.ErrConsensusStateNotFound, "failed to get consensus state: %v", err)
 		}
 
 		consensusState, err := ibcclienttypes.UnpackConsensusState(resp.ConsensusState)
 		if err != nil {
+			ctx.Logger().Error("SubmitQueryResult: failed to UnpackConsensusState",
```
quasisamurai, line 89: should we log err here? (and in the rest ctx.Logger().Debug calls)

zavgorodnii: yes,  because resp.ConsensusState was retrieved from ibc keeper (internal module, 100% should return a valid response if the error is nil) without an error

internal/sudo/sudo.go
```diff
@@ -9,22 +9,32 @@ The package contains the code to postprocess incoming from a relayer acknowledge
...
 	"fmt"
+
 	"github.com/CosmWasm/wasmd/x/wasm"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
 	channeltypes "github.com/cosmos/ibc-go/v3/modules/core/04-channel/types"
 	"github.com/tendermint/tendermint/libs/log"
 )
 
+type SudoMessageTxQueryResult struct {
```
sotnikov-s, line 22: let's start making comments for everything that we write/touch in some way? thus we will smoothly decrease the size of uncommented code. anyway we will have to do it. this is a great way to accomplish good comments coverage without hating it

internal/sudo/sudo.go
```diff
@@ -173,3 +186,39 @@ func (s *SudoHandler) SudoOpenAck(
 
 	return r, err
 }
+
+// SudoTxQueryResult is used to pass a tx query result to the contract that registered the query
+// to check whether the transaction actually satisfies the initial query arguments.
```
sotnikov-s: > to check whether the transaction actually satisfies the initial query arguments

is it outdated or something? this is not the only reason of using the SudoTxQueryResult, is it? e.g. later at the place you execute this method you say `Let the query owner contract process the query result`

x/interchainqueries/keeper/keeper.go
```diff
@@ -142,87 +140,28 @@ func clearQueryResult(result *types.QueryResult) types.QueryResult {
...
-
+// SaveTransactionAsProcessed simply stores a key (SubmittedTxKey + bigEndianBytes(queryID) + tx_hash) with
+// mock data. This key can be used to check whether a certain transaction was already submitted for a specific
+// transaction query.
+func (k Keeper) SaveTransactionAsProcessed(ctx sdk.Context, queryID uint64, txHash []byte) {
 	store := ctx.KVStore(k.storeKey)
-	key := types.GetSubmittedTransactionIDForQueryKey(queryID, txID)
-
-	store.Set(key, txBz)
+	key := types.GetSubmittedTransactionIDForQueryKey(queryID, txHash)
```
sotnikov-s, line 155: what's the point of having query_id as a part of the key? hash is unique and it guarantees no collisions. also, different interchain queries may be subscribed on the same address deposits, so having the query_id will result in having duplications of transactions saved as processed

x/interchainqueries/keeper/keeper.go
```diff
@@ -142,87 +140,28 @@ func clearQueryResult(result *types.QueryResult) types.QueryResult {
...
 
-	return nil
+	store.Set(key, []byte{1})
```
sotnikov-s: []byte{} is also applicable and it can save a couple of bytes :)

x/interchainqueries/keeper/process_block_results.go
```diff
@@ -83,8 +84,8 @@ func (k Keeper) VerifyHeaders(ctx sdk.Context, clientID string, header exported.
 	return nil
 }
 
-// VerifyBlock verifies headers and transaction in the block
-func (k Keeper) VerifyBlock(ctx sdk.Context, clientID string, block *types.Block) error {
+// ProcessBlock verifies headers and transaction in the block
```
sotnikov-s: please update the description with the new functionality you've introduced

swelf19 (commented):
x/interchainqueries/types/keys.go L44 and L60 dead code.

x/interchainqueries/keeper/process_block_results.go
```diff
@@ -101,18 +102,41 @@ func (k Keeper) VerifyBlock(ctx sdk.Context, clientID string, block *types.Block
 
 	tmHeader, ok := header.(*tendermintLightClientTypes.Header)
 	if !ok {
-		return sdkerrors.Wrapf(types.ErrInvalidType, "failed to cast header to tendermint Header: %v", err)
+		ctx.Logger().Debug("ProcessBlock: failed to cast current header to tendermint Header", "query_id", queryID)
```
sotnikov-s, line 106: shouldn't fails in case of this conversion and the one below be logged on the error level?

x/interchainqueries/keeper/keeper.go
swelf19: Multiple queries/contracts cat get the tx as a result to own query. Composite key (queryID, txHash) guarantees that every recipient get tx in sudo callback

x/interchainqueries/keeper/keeper.go
zavgorodnii: i was wondering whether anyone will notice : )

x/interchainqueries/keeper/process_block_results.go
zavgorodnii: don't think so. the block here comes directly from msg.Result.Block, which can contain all sorts of erroneous data

sotnikov-s (approved):

swelf19 (approved):

pr0n00gler (approved):

PR #23, opened 2022-07-22
feat/add-registered-query-owner

sotnikov-s:
https://p2pvalidator.atlassian.net/browse/LSC-75

With: https://github.com/neutron-org/cosmos-query-relayer/pull/11

**Changes:**
- added `owner` field to the stored registered queries and to the list of attributes for emitted events;
- fixed events emission condition;
- replaced MustUnmarshal with Unmarshal + error handling in IterateRegisteredQueries to prevent panics;
- added/improved comments.


foxpy (approved):

foxpy:
Little info for further mergers: tests on this PR's branch don't pass, but they will pass after merging with master, because there is still pesky commented code which triggers failed tests on this branch while it is already not present on master.

x/interchainqueries/keeper/keeper.go
```diff
@@ -306,7 +307,10 @@ func (k Keeper) IterateRegisteredQueries(ctx sdk.Context, fn func(index int64, q
 	i := int64(0)
 	for ; iterator.Valid(); iterator.Next() {
 		query := types.RegisteredQuery{}
-		k.cdc.MustUnmarshal(iterator.Value(), &query)
+		if err := k.cdc.Unmarshal(iterator.Value(), &query); err != nil {
+			k.Logger(ctx).Error("failed to unmarshal registered query %s when iterating: %w", iterator.Key(), err)
+			continue
```
mmikev, line 312: Maybe return an error? Cause logger will not show to a user that error happened.

sotnikov-s: good question. here's my answer: no, there is no reason of doing so and here's why:
- this error happens as the result of an Unmarshal call. this unmarshal call gets a previously registered query data as the argument and tries to wrap the RegisteredQuery struct over it. it's assumed that we've validated the query data before and here we simply reuse it;
- this is an internal loop called by EndBlocker, so there is no way to let the user know about the error because user doesn't call it. instead, the user calls [RegisterInterchainQuery](https://github.com/neutron-org/neutron/blob/feat/add-registered-query-owner/x/interchainqueries/keeper/msg_server.go#L24) which validates the query data passed by the user, and this is the very place where we should check if the user data is correct and return an error any.

pr0n00gler (approved):

zavgorodnii (approved):

PR #30, opened 2022-08-09
feat: no build on run

ratik:
There was an issue with build phase on every run of container. Now `neutrond` is built on image build. 

PS: I have no success on separating build process on 2 phases bc when I copy binary of neutron from build phase to regular it can't be executed (this issue I believe only for M1) so I left this stuff in one phase.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,15 +12,19 @@ EXPOSE 1317
 EXPOSE 9090 
 EXPOSE 9091
 EXPOSE 8081
-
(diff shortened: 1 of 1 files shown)


foxpy (approved):
image now takes 20 seconds to start instead of 100'ish, and seems like everything continues to work intact

pr0n00gler (approved):

zavgorodnii (approved):

PR #33, opened 2022-08-12
feat: add envs to docker-compose

ratik:
Changes:
- added missing env variables to relayer config, making it not crash on `make start-cosmopark`;
- configured relayer to use `demowallet3` instead of `demowallet1` as its signing key, this will allow us to freely use demowallet1 in tests, without it being occupied by relayer, which prevented us to issue transactions.


foxpy (dismissed):

foxpy (dismissed):

docker-compose.yml
```diff
@@ -53,6 +53,8 @@ services:
       - RELAYER_TARGET_CHAIN_OUTPUT_FORMAT=json
       - RELAYER_TARGET_CHAIN_SIGN_MODE_STR=direct
       - RELAYER_REGISTRY_ADDRESSES=
+      - RELAYER_NEUTRON_CHAIN_ALLOW_KV_CALLBACKS=1
+      - RELAYER_ALLOW_TX_QUERIES=1
```
quasisamurai: why `1`, not `true`?

ratik: dunno, bc it works 😅
shall I change it to `true`?

foxpy: Let's hope next reviewer wouldn't ask us to rename it back to 1 :v

foxpy (approved):

docker-compose.yml
quasisamurai: I believe `true` is kinda more explicit, due the type is bool

quasisamurai (approved):

docker-compose.yml
sotnikov-s: hey guys we can save a couple of bytes by dropping the superfluous `true` and using simple `1`

docker-compose.yml
```diff
@@ -53,6 +53,8 @@ services:
       - RELAYER_TARGET_CHAIN_OUTPUT_FORMAT=json
       - RELAYER_TARGET_CHAIN_SIGN_MODE_STR=direct
       - RELAYER_REGISTRY_ADDRESSES=
+      - RELAYER_NEUTRON_CHAIN_ALLOW_KV_CALLBACKS=true
+      - RELAYER_ALLOW_TX_QUERIES=true
```
sotnikov-s, line 57: out of scope, but I don't like the way we've designed these variables. why one of them is inside the neutron chain's config, and the other one is inside the relayer's config? I think they both should be on the relayer's level bc they're not related to the neutron chain, they're about the relayer configuration. do you agree? if yes, would it hurt if we fix it here?

docker-compose.yml
ratik: nope

docker-compose.yml
ratik: I think it's really out of scope of this PR. So let's have separate task. Would you mind creating this one

pr0n00gler (approved):

docker-compose.yml
sotnikov-s: just a joke

docker-compose.yml
sotnikov-s: sure

sotnikov-s (approved):

PR #36, opened 2022-08-16
feat: get remote height #LSC-111

ratik:
as the [task](https://p2pvalidator.atlassian.net/jira/software/projects/LSC/boards/121?selectedIssue=LSC-111) describes we need a way to query remote chain for the current height

to test it properly make sure you added some sleep btw starts 1st and 2nd chains, then you can query remote height like `./build/neutrond query interchainqueries query-height connection-0 --node tcp://127.0.0.1:16657 --home data/test-1`


x/interchainqueries/client/cli/query.go
```diff
@@ -112,3 +113,27 @@ func CmdQueryRegisteredQueryResult() *cobra.Command {
 
 	return cmd
 }
+
+func CmdQueryRemoteHeight() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "query-height [connection-id]",
```
mmikev: `query-height` is a bit misleading actually. Let's call it like `query-latest-known-remote-height` or smth? I don't like this long name but you got the idea I hope :)

mmikev: And change it everywhere definitely

x/interchainqueries/keeper/grpc_query.go
```diff
@@ -6,6 +6,9 @@ import (
 	"github.com/cosmos/cosmos-sdk/store/prefix"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+	contypes "github.com/cosmos/ibc-go/v3/modules/core/03-connection/types"
+	ttypes "github.com/cosmos/ibc-go/v3/modules/light-clients/07-tendermint/types"
```
swelf19: `ttypes` double `t` looks like a typo for me, but its a matter of taste

x/interchainqueries/keeper/grpc_query.go
```diff
@@ -56,3 +59,20 @@ func (k Keeper) QueryResult(goCtx context.Context, request *types.QueryRegistere
...
+	}
+	clientState := r.GetIdentifiedClientState().GetClientState()
+
+	m := new(ttypes.ClientState)
+	err = proto.Unmarshal(clientState.Value, m)
+	if err != nil {
+		return nil, sdkerrors.Wrapf(types.ErrProtoUnmarshal, "Can't unmarshal client state")
+	}
+
+	return &types.QueryLastRemoteHeightResponse{Height: m.LatestHeight.RevisionHeight}, err
```
swelf19: The only 'nil' value is acceptable value for the error at this place, could we return nil explicitly?

x/interchainqueries/keeper/grpc_query.go
```diff
@@ -56,3 +59,20 @@ func (k Keeper) QueryResult(goCtx context.Context, request *types.QueryRegistere
 	}
 	return &types.QueryRegisteredQueryResultResponse{Result: result}, nil
 }
+
+func (k Keeper) LastRemoteHeight(goCtx context.Context, request *types.QueryLastRemoteHeight) (*types.QueryLastRemoteHeightResponse, error) {
+	req := contypes.QueryConnectionClientStateRequest{ConnectionId: request.ConnectionId}
+	r, err := k.ibcKeeper.ConnectionClientState(goCtx, &req)
+	if err != nil {
+		return nil, sdkerrors.Wrapf(types.ErrInvalidConnectionID, "Connection not found")
```
swelf19: afaik it's not recommended to start error messages with a capital letter https://github.com/golang/go/wiki/CodeReviewComments#error-strings

swelf19 (dismissed):

foxpy (dismissed):

x/interchainqueries/client/cli/query.go
```diff
@@ -112,3 +113,27 @@ func CmdQueryRegisteredQueryResult() *cobra.Command {
...
+func CmdQueryLastRemoteHeight() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "query-last-remote-height [connection-id]",
+		Short: "queries last remote height by connection id",
+		Args:  cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			clientCtx := client.GetClientContextFromCmd(cmd)
+			connectionId := args[0]
+			queryClient := types.NewQueryClient(clientCtx)
+			println("connection id", connectionId)
```
mmikev: Seems unnecessary

pr0n00gler (commented):
Would be nice to have a simple test for this query in `keeper_test.go` (or even `grpc_query_test.go` is better)

x/interchainqueries/client/cli/query.go
foxpy: woops, that looks like some post-debugging remnant

sotnikov-s: println?

x/interchainqueries/client/cli/query.go
```diff
@@ -112,3 +113,27 @@ func CmdQueryRegisteredQueryResult() *cobra.Command {
 
 	return cmd
 }
+
+func CmdQueryLastRemoteHeight() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "query-last-remote-height [connection-id]",
+		Short: "queries last remote height by connection id",
```
sotnikov-s, line 120: as far as I can see, there is no way to get the remote height but via using connection id, is there?

sotnikov-s (changes requested):

x/interchainqueries/client/cli/query.go
sotnikov-s: lol y don't I see the comment in the Changes section? otherwise I wouldn't have left the same one

x/interchainqueries/client/cli/query.go
ratik: you may have many networks connected so you have to identify one by connection id

foxpy (dismissed):

x/interchainqueries/client/cli/query.go
foxpy: imo GitHub review UI is very underdeveloped

sotnikov-s (dismissed):

pr0n00gler (approved):

swelf19 (approved):

PR #48, opened 2022-09-07
feat: add deposit collection for interchain queries

albertandrejev:
We need to introduce TTLs for KV-queries: if a KV-query was not updated for N local blocks (added as a parameter, controlled by governance), delete the KV-query.

Implements https://p2pvalidator.atlassian.net/browse/NTRN-184

Related PRs: 
- https://github.com/neutron-org/neutron-contracts/pull/29
- https://github.com/neutron-org/neutron-docs/pull/9

DoD
- [x] Parameter introduced;
- [x] Deposit collection implemented;
- [x] Deposit payout added;
- [x] Unit tests


x/interchainqueries/types/registered_query.go
```diff
@@ -0,0 +1,15 @@
+package types
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+)
+
+func (queryInfo *RegisteredQuery) GetOwnerAddress() (creator sdk.AccAddress, err error) {
+	creator, errCreator := sdk.AccAddressFromBech32(queryInfo.Owner)
+	return creator, sdkerrors.Wrapf(errCreator, ErrInvalidOwner.Error(), queryInfo.Owner)
```
mmikev: I don't think we need another error like `ErrInvalidOwner` cause we use `sdkerrors.ErrInvalidAddress` to handle errors with addresses, so I suggest:

```suggestion
	return creator, sdkerrors.Wrapf(sdkerrors.ErrInvalidAddress, "failed to decode owner address: %s", msg.Sender)
```

x/interchainqueries/keeper/keeper.go
```diff
@@ -260,3 +263,26 @@ func (k Keeper) checkRegisteredQueryExists(ctx sdk.Context, id uint64) bool {
 
 	return store.Has(types.GetRegisteredQueryByIDKey(id))
 }
+
+func (k Keeper) CollectDeposit(ctx sdk.Context, queryInfo types.RegisteredQuery) error {
+	owner, err := queryInfo.GetOwnerAddress()
+	if err != nil {
+		panic(err.Error())
```
mmikev, line 287: I don't think `panic` is good here. Please, return an error like everywhere else

x/interchainqueries/types/registered_query.go
```diff
@@ -0,0 +1,15 @@
...
+}
+
+func (queryInfo *RegisteredQuery) GetDepositCoin() (wager sdk.Coin) {
+	return sdk.NewCoin(sdk.DefaultBondDenom, sdk.NewInt(int64(queryInfo.Deposit)))
```
mmikev: `sdk.DefaultBondDenom` is not a proper constant here:
1. Because it's hardcoded as `stake`. We should use our own coin denom (`untrn`) and we must have such constant somewhere in app.go;
2. We should not have **bond** denom at all, since Neutron doesn't have staking features

x/interchainqueries/keeper/keeper.go
```diff
@@ -260,3 +263,26 @@ func (k Keeper) checkRegisteredQueryExists(ctx sdk.Context, id uint64) bool {
...
+		return err
+	}
+
+	return nil
+}
+
+func (k Keeper) MustPayOutDeposit(ctx sdk.Context, deposit sdk.Coin, sender sdk.AccAddress) error {
+	err := k.bank.SendCoinsFromModuleToAccount(ctx, types.ModuleName, sender, sdk.NewCoins(deposit))
+	if err != nil {
+		panic(types.ErrCannotPayDeposit.Error())
```
mmikev: The same about `panic()`

x/interchainqueries/keeper/msg_server.go
```diff
@@ -82,12 +89,15 @@ func (k msgServer) RemoveInterchainQuery(goCtx context.Context, msg *types.MsgRe
...
-	if query.GetOwner() != msg.GetSender() {
+	timoutBlock := query.LastSubmittedResultLocalHeight + k.GetParams(ctx).QuerySubmitTimeout
+	if uint64(ctx.BlockHeight()) <= timoutBlock && query.GetOwner() != msg.GetSender() {
 		ctx.Logger().Debug("RemoveInterchainQuery: authorization failed",
 			"msg", msg)
 		return nil, sdkerrors.Wrap(sdkerrors.ErrUnauthorized, "authorization failed")
 	}
+
 	k.RemoveQueryByID(ctx, query.Id)
+	k.MustPayOutDeposit(ctx, query.GetDepositCoin(), msg.GetSigners()[0])
```
mmikev: unhandled error

pr0n00gler (changes requested):

proto/interchainqueries/genesis.proto
```diff
@@ -36,6 +36,9 @@ message RegisteredQuery {
 
   // The remote chain last block height when the query result was updated.
   uint64 last_submitted_result_remote_height = 10;
+
+  // Amount of coins deposited for the query.
+  uint64 deposit = 11;
```
mmikev: You added a new field into `RegisteredQuery` struct. Please, do not forget to update wasm bindings for relevant structure (in [Neutron itself](https://github.com/neutron-org/neutron/blob/caf377b38db1723f2d3941e61f450af5830de212/wasmbinding/bindings/query.go#L53) and in the [Neutron CosmWasm SDK](https://github.com/neutron-org/neutron-contracts/blob/a47bfac69667da57f8bf6ea81c9f16240e145c6d/packages/neutron-sdk/src/bindings/types.rs#L25))

x/interchainqueries/keeper/msg_server.go
```diff
@@ -82,12 +89,15 @@ func (k msgServer) RemoveInterchainQuery(goCtx context.Context, msg *types.MsgRe
 		return nil, sdkerrors.Wrapf(err, "failed to get query by query id: %v", err)
 	}
 
-	if query.GetOwner() != msg.GetSender() {
+	timoutBlock := query.LastSubmittedResultLocalHeight + k.GetParams(ctx).QuerySubmitTimeout
```
mmikev: tim**e**out

x/interchainqueries/types/registered_query.go
zavgorodnii: We are going to have staking enabled for the testnet, let's keep it for now; changing `stake` to `ntrn` is a good idea

x/interchainqueries/keeper/keeper.go
albertandrejev: But it means that something wrong with code. If the module cannot pay, it means the escrow account has failed. This error is much more serious: an invariant has been violated and the whole application must be terminated.

x/interchainqueries/keeper/keeper.go
albertandrejev: But if we can not get query owner address it means that something wrong with address, not sure that this is expected situation and should be covered with exception. What do you think?

x/interchainqueries/keeper/msg_server.go
albertandrejev: yes, because it should `panic`. Anyway let's think about it, how it is better to cover this situation.

mmikev: I think we can just do not return anything from `MustPayOutDeposit`, since there is only one function call which is covered by `panic` anyway

x/interchainqueries/keeper/keeper.go
mmikev: Agree. But please include `err` into the panic call to make error trace easier to understand what went wrong exactly

x/interchainqueries/types/registered_query.go
albertandrejev: I decided to solve this problem from another side. We will store denom in the module param like other modules do.

x/interchainqueries/types/errors.go
```diff
@@ -11,16 +11,19 @@ var (
...
+	ErrProtoUnmarshal            = sdkerrors.Register(ModuleName, 1110, "failed to unmarshal protobuf bytes")
+	ErrInvalidType               = sdkerrors.Register(ModuleName, 1111, "invalid type")
+	ErrInternal                  = sdkerrors.Register(ModuleName, 1112, "internal error")
+	ErrInvalidProof              = sdkerrors.Register(ModuleName, 1113, "merkle proof is invalid")
+	ErrInvalidHeader             = sdkerrors.Register(ModuleName, 1114, "header is invalid")
+	ErrInvalidHeight             = sdkerrors.Register(ModuleName, 1115, "height is invalid")
+	ErrNoQueryResult             = sdkerrors.Register(ModuleName, 1116, "no query result")
+	ErrNotContract               = sdkerrors.Register(ModuleName, 1117, "not a contract")
+	ErrCannotPayDeposit          = sdkerrors.Register(ModuleName, 1118, "can not pay deposit to signer")
+	ErrInvalidOwner              = sdkerrors.Register(ModuleName, 1119, "owner address is invalid: %s")
```
mmikev: `ErrCannotPayDeposit ` and `ErrInvalidOwner` are unused now, please remove them.

Also remove `ErrInvalidZoneID` too, cause it's unused too.

x/interchainqueries/keeper/msg_server.go
```diff
@@ -82,12 +90,19 @@ func (k msgServer) RemoveInterchainQuery(goCtx context.Context, msg *types.MsgRe
 		return nil, sdkerrors.Wrapf(err, "failed to get query by query id: %v", err)
 	}
 
-	if query.GetOwner() != msg.GetSender() {
+	fmt.Println("RemoveInterchainQuery")
+	fmt.Println(query)
+	fmt.Println(k.GetParams(ctx))
```
mmikev: Seems like printlns from dev process, please remove it

pr0n00gler (changes requested):

pr0n00gler (dismissed):
Looks good to me, but we need to to wait for integration tests before merging

zavgorodnii (dismissed):

sotnikov-s (commented):
are relative integration tests coming soon?

testutil/interchainqueries/keeper/interchainadapter.go
```diff
@@ -44,6 +45,7 @@ func InterchainQueriesKeeper(t testing.TB) (*keeper.Keeper, sdk.Context) {
 		paramsSubspace,
 		nil, // TODO: do a real ibc keeper
 		nil, // TODO: do a real wasm keeper
+		*new(bankkeeper.Keeper),
```
sotnikov-s: why not just nil?

albertandrejev:
> are relative integration tests coming soon?

yes

x/interchainqueries/keeper/keeper_test.go
```diff
@@ -82,9 +101,18 @@ func (suite *KeeperTestSuite) TestRegisterInterchainQuery() {
 
 		tt.malleate(contractAddress.String())
 
+		if tt.topupBalance {
+			senderAddress := suite.ChainA.SenderAccounts[0].SenderAccount.GetAddress()
+
+			coinsAmnt := sdktypes.NewCoins(sdktypes.NewCoin(sdktypes.DefaultBondDenom, sdktypes.NewInt(int64(10_000_000))))
+			bankKeeper := suite.GetNeutronZoneApp(suite.ChainA).BankKeeper
+
+			bankKeeper.SendCoins(ctx, senderAddress, contractAddress, coinsAmnt)
```
sotnikov-s: could you please move this top up logic to a separate func? I can see four very same code blocks in this file making top ups

x/interchainqueries/keeper/msg_server.go
```diff
@@ -82,12 +89,15 @@ func (k msgServer) RemoveInterchainQuery(goCtx context.Context, msg *types.MsgRe
 		return nil, sdkerrors.Wrapf(err, "failed to get query by query id: %v", err)
 	}
 
-	if query.GetOwner() != msg.GetSender() {
+	timeoutBlock := query.LastSubmittedResultLocalHeight + k.GetParams(ctx).QuerySubmitTimeout
```
sotnikov-s: what you guys think about saving the QuerySubmitTimeout as well as we save QueryDeposit individually for each query? e.g. I registered a query when the module's QuerySubmitTimeout was of 50 blocks hardly relying on this value, and later if this value changes to 30, I will 1) lose money; 2) lose query.

x/interchainqueries/types/params.go
```diff
@@ -1,30 +1,49 @@
...
-	return NewParams()
+	return NewParams(DefaultQuerySubmitTimeout, DefaultQueryDeposit)
 }
 
 // ParamSetPairs get the params.ParamSet
 func (p *Params) ParamSetPairs() paramtypes.ParamSetPairs {
-	return paramtypes.ParamSetPairs{}
+	return paramtypes.ParamSetPairs{
+		paramtypes.NewParamSetPair(KeyQuerySubmitTimeout, &p.QuerySubmitTimeout, func(value interface{}) error { return nil }),
+		paramtypes.NewParamSetPair(KeyQueryDeposit, &p.QueryDeposit, validateCoins),
```
sotnikov-s, line 45: is there any usage of these validate functions at all, if below in the `(p Params) Validate()` we simply return nil?

testutil/interchainqueries/keeper/interchainadapter.go
albertandrejev: because we are using bank keeper in our tests. It should send coins to another address (to top up smart contract).

x/interchainqueries/keeper/msg_server.go
albertandrejev: Good catch! Yes, I guess we should store `QuerySubmitTimeout` into `RegisteredQuery` as well. Otherwise it may lead to different confustions.

testutil/interchainqueries/keeper/interchainadapter.go
sotnikov-s: I mean, bankkeeper.Keeper is an interface, and the `*new(bankkeeper.Keeper)` statement results in a nil value. what's the point? if it really is used in tests, then something must be really bad with them for we should see nil pointer dereferences somewhere, but I can see tests passing

albertandrejev: I removed `nil`.

x/interchainqueries/types/params.go
albertandrejev: It is required by `NewParamSetPair`

PR #49, opened 2022-09-08
feat: remove endblocker events

zavgorodnii:
This PR:

1. Removes queries processing from the EndBlocker completely;
2. Adds events emitting to the Update, Create, Remove actions (queries);
3. Adds pagination and `connection_id`, `owner` filters to `GetRegisteredQueries()`;
4. Changes events type from "message" to "neutron";
5. Updates the `openapi.yml` swagger specs.

It is related to the following PRs:

1. https://github.com/neutron-org/neutron-contracts/pull/27
2. https://github.com/neutron-org/neutron-query-relayer/pull/35
3. https://github.com/neutron-org/neutron-integration-tests/pull/10

**HOW TO TEST THIS CODE?**

See the corresponding section in https://github.com/neutron-org/neutron-query-relayer/pull/35.

> NOTE: most of the huge diff is the generated swagger specs `openapi.yml`


x/interchainqueries/keeper/grpc_query.go
```diff
@@ -31,19 +34,34 @@ func (k Keeper) RegisteredQueries(goCtx context.Context, req *types.QueryRegiste
...
-	store := prefix.NewStore(ctx.KVStore(k.storeKey), types.RegisteredQueryKey)
-	iterator := sdk.KVStorePrefixIterator(store, nil)
-	defer iterator.Close()
+func (k Keeper) GetRegisteredQueries(ctx sdk.Context, req *types.QueryRegisteredQueriesRequest) (*types.QueryRegisteredQueriesResponse, error) {
+	if req == nil {
+		return nil, status.Error(codes.InvalidArgument, "empty request")
+	}
 
+	store := prefix.NewStore(ctx.KVStore(k.storeKey), types.RegisteredQueryKey)
 	queries := make([]types.RegisteredQuery, 0)
```
mmikev: Maybe we can preallocate memory here?
```go
queries := make([]types.RegisteredQuery, 0, req.Pagination.Limit)
```

x/interchainqueries/keeper/grpc_query.go
```diff
@@ -31,19 +34,34 @@ func (k Keeper) RegisteredQueries(goCtx context.Context, req *types.QueryRegiste
...
+	}
+
 	store := prefix.NewStore(ctx.KVStore(k.storeKey), types.RegisteredQueryKey)
-	iterator := sdk.KVStorePrefixIterator(store, nil)
-	defer iterator.Close()
+	queries := make([]types.RegisteredQuery, 0, req.Pagination.Limit)
 
-	queries := make([]types.RegisteredQuery, 0)
-	for ; iterator.Valid(); iterator.Next() {
+	pageRes, err := querytypes.Paginate(store, req.Pagination, func(key, value []byte) error {
```
swelf19, line 48: i dont think the pagination works the way you expect. `onResult` callback for unmarshaling only. On the client side you will get responses with random amount of queries(which meets filter params) until the end of the iterator

x/interchainqueries/keeper/msg_server.go
```diff
@@ -234,14 +247,52 @@ func (k msgServer) SubmitQueryResult(goCtx context.Context, msg *types.MsgSubmit
...
+}
+
+func getEventsQueryRemoved(query *types.RegisteredQuery) sdk.Events {
+	return sdk.Events{
+		sdk.NewEvent(
+			types.EventTypeNeutronMessage,
+			sdk.NewAttribute(sdk.AttributeKeyModule, types.AttributeValueCategory),
+			sdk.NewAttribute(sdk.AttributeKeyAction, types.AttributeValueQueryRemoved),
+			sdk.NewAttribute(types.AttributeKeyQueryID, strconv.FormatUint(query.Id, 10)),
+			sdk.NewAttribute(types.AttributeKeyConnectionID, query.ConnectionId),
```
swelf19, line 289: I dont know why do we need all these attributes for the removeQuery event, but if so, can we create a more generic function for both `getEventsQueryUpdated` and `getEventsQueryUpdated` almost all attributes the same

swelf19 (commented):
do not see a pagination added to the cli command.
`QueryRegisteredQueriesRequest` wasmbindings was not fixed

x/interchainqueries/keeper/msg_server.go
zavgorodnii: removed most of the attributes from the remove action

x/interchainqueries/keeper/grpc_query.go
zavgorodnii: You are right. Please add a task to backlog with a full description of the issue and a possible solution (the one that we discussed on a call). We'll fix this next week.

zavgorodnii:
> do not see a pagination added to the cli command. `QueryRegisteredQueriesRequest` wasmbindings was not fixed

added the bindings

pr0n00gler:
> But not to the cli?

god damn it you are right

wasmbinding/test/custom_query_test.go
```diff
@@ -43,10 +43,10 @@ func (suite *CustomQuerierTestSuite) TestInterchainQueryResult() {
 		Keys: []*icqtypes.KVKey{
 			{Path: host.StoreKey, Key: clientKey},
 		},
-		QueryType:         string(icqtypes.InterchainQueryTypeKV),
-		UpdatePeriod:      1,
-		ConnectionId:      suite.Path.EndpointA.ConnectionID,
-		LastEmittedHeight: uint64(ctx.BlockHeight()),
+		QueryType: string(icqtypes.InterchainQueryTypeKV),
+
```
sotnikov-s: redundant empty line

x/interchainqueries/keeper/grpc_query.go
sotnikov-s: you mean there will be a `[]types.RegisteredQuery` behind the `value []byte`? as far as I can see, the onResult func gets a single value from the iterator: https://github.com/cosmos/cosmos-sdk/blob/main/types/query/pagination.go#L95. please elaborate

wasmbinding/bindings/query.go
```diff
@@ -34,6 +34,32 @@ type QueryInterchainAccountAddressRequest struct {
 }
 
 type QueryRegisteredQueriesRequest struct {
+	Owners       []string     `json:"owners,omitempty"`
+	ConnectionId string       `json:"connection_id,omitempty"`
+	Pagination   *PageRequest `json:"pagination,omitempty"`
+}
+
+type PageRequest struct {
```
sotnikov-s: cosmos sdk already has defined this struct: https://github.com/cosmos/cosmos-sdk/blob/c752ddda4f878f7d12d3cf2d42f2fb42c6edfc1a/types/query/pagination.pb.go#L32

x/interchainqueries/keeper/grpc_query.go
```diff
@@ -31,19 +34,36 @@ func (k Keeper) RegisteredQueries(goCtx context.Context, req *types.QueryRegiste
...
 		query := types.RegisteredQuery{}
-		k.cdc.MustUnmarshal(iterator.Value(), &query)
-		queries = append(queries, query)
+		k.cdc.MustUnmarshal(value, &query)
+
+		var (
+			passedOwnerFilter        = len(req.GetOwners()) == 0 || containsString(req.GetOwners(), query.GetOwner())
+			passesConnectionIDFilter = req.GetConnectionId() == "" || query.ConnectionId == req.GetConnectionId()
```
sotnikov-s: passedOwnerFilter with passe**d**, but passesConnectionIDFilter with passe**s**. please unify

x/interchainqueries/keeper/grpc_query.go
```diff
@@ -31,19 +34,36 @@ func (k Keeper) RegisteredQueries(goCtx context.Context, req *types.QueryRegiste
...
-	queries := make([]types.RegisteredQuery, 0)
-	for ; iterator.Valid(); iterator.Next() {
+	pageRes, err := querytypes.Paginate(store, req.Pagination, func(key, value []byte) error {
 		query := types.RegisteredQuery{}
-		k.cdc.MustUnmarshal(iterator.Value(), &query)
-		queries = append(queries, query)
+		k.cdc.MustUnmarshal(value, &query)
+
+		var (
+			passedOwnerFilter        = len(req.GetOwners()) == 0 || containsString(req.GetOwners(), query.GetOwner())
```
sotnikov-s: you better create a map out of the `req.GetOwners()` to pass the owner check in O(1) instead of running over the whole slice. you can also use the result map in a func similar to the `containsString` which will handle both `empty map` and `exists in map` cases to make it more concise

x/interchainqueries/keeper/msg_server.go
sotnikov-s: there are still three attributes that are added to both event types. please consider making a func that returns common for both event types attributes and passing it to the end of the `sdk.NewEvent(` func via `attrs...)`

x/interchainqueries/module.go
```diff
@@ -76,6 +77,9 @@ func (AppModuleBasic) RegisterRESTRoutes(clientCtx client.Context, rtr *mux.Rout
 
 // RegisterGRPCGatewayRoutes registers the gRPC Gateway routes for the module.
 func (AppModuleBasic) RegisterGRPCGatewayRoutes(clientCtx client.Context, mux *runtime.ServeMux) {
+	if err := types.RegisterQueryHandlerClient(context.Background(), mux, types.NewQueryClient(clientCtx)); err != nil {
```
sotnikov-s, line 80: could you please explain this piece if added code? why we didn't have it before? kinda complicated to understand from the func comment and innards

sotnikov-s (changes requested):

x/interchainqueries/module.go
zavgorodnii: this is standard code that enables REST query handlers for the module

zavgorodnii: > there are still three attributes that are added to both event types. please consider making a func that returns common for both event types attributes and passing it to the end of the sdk.NewEvent( func via attrs...)

i think this is too much, no need to introduce an intermediate function just for a couple of attributes

x/interchainqueries/keeper/grpc_query.go
zavgorodnii: not sure this is a necessary optimisation, but okay

x/interchainqueries/keeper/grpc_query.go
zavgorodnii: > you mean there will be a []types.RegisteredQuery behind the value []byte? as far as I can see, the onResult func gets a single value from the iterator: https://github.com/cosmos/cosmos-sdk/blob/main/types/query/pagination.go#L95. please elaborate

not sure what your point is. we unmarshal the `value []byte` to a single query struct

wasmbinding/bindings/query.go
swelf19: I actually suggest to return custom binding, and that's why, standart cosmos-sdk PageRequest has two fields with the same meaning - "offset", these fields Offset and Key, but offset has a way worst perfomance, using offset field Paginate function iterates over keys and drop items while count < offset, when with the Key field it starts to iterate right from the needed key. I suggest to remove Offset from wasm binding, i believe it eventually will be removed from cosmos-sdk PageRequest as well.

sotnikov-s: we can just let the relayer users know that Offset is way more expensive than keys. also, it's said in the Offset description that it is used when key is unavailable, which means to be a possible scenario
```go
// offset is a numeric offset that can be used when key is unavailable.
// It is less efficient than using key. Only one of offset or key should
// be set.
Offset uint64 `protobuf:"varint,2,opt,name=offset,proto3" json:"offset,omitempty"`
```
ah, or they mean the cases when there is no Key provided within the request?

sotnikov-s (dismissed):

sotnikov-s (approved):

pr0n00gler (approved):

swelf19 (approved):

PR #58, opened 2022-09-30
feat: don't use ioutil, it is deprecrated

faddat:
Don't use ioutil because it is deprecated.


foxpy (approved):
[Indeed](https://pkg.go.dev/io/ioutil)

pr0n00gler (approved):

PR #59, opened 2022-10-01
chore: gofumpt

faddat:
Gofumpt code formatting provides a clear standard to 
how your code should be formatted.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -316,7 +316,6 @@ func New(
 	wasmOpts []wasm.Option,
 	baseAppOptions ...func(*baseapp.BaseApp),
 ) *App {
-
 	appCodec := encodingConfig.Marshaler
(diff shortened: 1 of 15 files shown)


foxpy (commented):
go fmt has an advantage of being preinstalled with go toolchain.

In my opinion, this PR is not about fixing code style _for now_, instead, it suggests us to start using gofumpt. @oopcode your opinion on that?

faddat:
So, more or less-- the sctricter formatting of gofumpt standardizes the code better.  We've begun to use it across osmosis, the cosmos-sdk, and gaia.

foxpy (approved):
> We've begun to use it across osmosis, the cosmos-sdk, and gaia.

If that's the case, I am more than happy to approve this PR.

pr0n00gler (approved):

PR #65, opened 2022-10-01
use latest known good versions

faddat:


foxpy (changes requested):
welp, it doesn't compile now

faddat:
Hey sorry I meant to mark this as draft there's a small change that I need to make I'll try and make that now

In the tests the parameters for GenTX changed

faddat:
It isn't the txCtx.R thing -- it is that upstream isn't bumped.  To use 0.45.8, see #66

PR #66, opened 2022-10-01
Update to changes from mainline wasm

faddat:
this pr uses this branch of wasmd:

* https://github.com/neutron-org/wasmd/pull/2

That will bring neutron up to date with the `main` branch of wasmd.

Generally speaking, until there's clarity for a neutron release, I'd recommend creating "stable mixes" of the features that you desire, and work without hesitation even from untagged versions.  This will allow you to ensure that the released code is of the highest quality possible, and as compatible as possible with the upstreams.


zavgorodnii:
Hi! We are using a patched version of wasmd, so this PR is not useful for us just yet : )

PR #67, opened 2022-10-02
sdk 46

faddat:
If you don't keep up with upstream changes, you will end up refactoring from 45 to 47 and from ibc v3.0.0 to ibc v6 or v7 and that'll take a very long time. 

The notable changes here have to do with the composition of an ibc "stack" around ibc modules and the introduction of the ibc middleware.  This PR also introduces the ibc fee module.  

Note: currently, it is stuck at the cursed initialization function.  

- sdk 46
- use the canonical golangci-lint github action
- format code nice-like
- fumpt
- fix naming
- neutron now uses pretty strict linting
- tidy with 1.19


zavgorodnii:
Unfortunately, IS does [not](https://github.com/cosmos/interchain-security/blob/main/go.mod#L6) support 0.46. As far as I understand, the informal team is going to add support for 0.47, skipping 0.46 entirely.

P.S. I feel bad about closing this PR; it's a lot of work : ( Thank you so much!

PR #68, opened 2022-10-04
Fix: pagination with filtration in GetRegisteredQueries

pr0n00gler:
Task: https://p2pvalidator.atlassian.net/jira/software/projects/NTRN/boards/121?selectedIssue=NTRN-191

Additional tests for the query are implemented


swelf19 (approved):

x/interchainqueries/keeper/grpc_query.go
```diff
@@ -54,11 +54,18 @@ func (k Keeper) GetRegisteredQueries(ctx sdk.Context, req *types.QueryRegistered
...
 
-		if passedOwnerFilter && passedConnectionIDFilter {
+		// if result does not satisfy the filter, return (false, nil) to tell FilteredPaginate method to skip this value
+		if !(passedOwnerFilter && passedConnectionIDFilter) {
+			return false, nil
+		}
+
+		// when accumulate equals true, it means we are in the right offset/limit position
+		// so we check value satisfies the filter and add it to the final result slice
+		if accumulate && passedOwnerFilter && passedConnectionIDFilter {
```
NeverHappened, line 64: seems that its second and unnecessary check for `passedOwnerFilter && passedConnectionIDFilter`

NeverHappened: also maybe just do early return here? (just suggestion)
```
if !accumulate {
  return true, nil
}

queries = append(queries, query)
...
```

ratik (approved):

x/interchainqueries/keeper/grpc_query.go
mmikev: > seems that its second and unnecessary check for passedOwnerFilter && passedConnectionIDFilter

Yes, but I think this improves readability, because it explicitly describes the condition of the final result accumulation.

> also maybe just do early return here?

The same about readability. I think an early return will decrease readability here, because current solution explicitly says when to accumulate the result.

NeverHappened (approved):

foxpy (approved):

zavgorodnii (approved):

PR #73, opened 2022-10-12
Chore: create channel in hermes network scripts

foxpy:
Without this PR, `make init` didn't create a channel, leaving Docker and "native" setups to remain slightly different.


zavgorodnii:
tested with `make init`, seems to work fine

zavgorodnii (approved):

swelf19 (approved):
we need to update the readme https://github.com/neutron-org/neutron-contracts/tree/main/contracts/ibc_transfer to not create transfer chanel twice

PR #74, opened 2022-10-12
Audit fix 5: unbounded iteration in MsgRegisterInterchainQuery::ValidateBasic()

foxpy:
Tracking task: [NTRN-216](https://p2pvalidator.atlassian.net/browse/NTRN-216).


x/interchainqueries/types/types.go
```diff
@@ -45,6 +43,9 @@ const (
 
 	// AttributeValueQueryRemoved represents the value for the 'action' event attribute.
 	AttributeValueQueryRemoved = "query_removed"
+
+	// maxTransactionsFilters defines maximum allowed amount of tx filters in msgRegisterInterchainQuery
+	maxTransactionsFilters = 16
```
mmikev: Why 16?
Isn't it a pretty small number?

foxpy: Bold of you to assume I have any kind of idea which number is good.
Bring your proposals!

mmikev: Need to think about it :)
I'm just curious why you chose exactly 16, maybe some empirical calculations

Also check please, that there is no limit for tx search conditions in Tendermint itself. If there is one then the answer is obvious - we need to rely on the limit from Tendermint

foxpy: alright, I will look into it

zavgorodnii: given the fact that tx filters must have their correlates in the message fields, it's sort of safe to assume that we should set the `maxTransactionsFilters` to the maximum number of fields in the sdk * some coefficient?

foxpy: Well:
- I haven't encountered any limit on amount of tx filters, it is possible to apply thousands of them;
- @pr0n00gler tells the biggest Cosmos message they have seen is  as big as 10 fields, they also suggest to set limit to 50.

I personally don't have any ideas what to set this constant to, so I vote for 50.

zavgorodnii (approved):
but please add a pr to the docs repo with a note about the filters limit

foxpy:
> but please add a pr to the docs repo with a note about the filters limit

Well, it will make sense for me to add it when we agree on a limit.

x/interchainqueries/types/types.go
foxpy: Outside of this thread there has been made a decision to use a limit of 32.

pr0n00gler (approved):
What do you think of also adding such limit on the Neutron CosmWasm SDK level?
And adding some doc comments to the rust code about limit, so users can see it directly in the SDK.

PR #77, opened 2022-10-16
Audit fix 6: unbounded iteration in PerformSubmitTx()

foxpy:
Tracking task: [NTRN-217](https://p2pvalidator.atlassian.net/browse/NTRN-217).


zavgorodnii (approved):

wasmbinding/message_plugin.go
```diff
@@ -201,6 +201,20 @@ func (m *CustomMessenger) PerformSubmitTx(ctx sdk.Context, contractAddr sdk.AccA
 		InterchainAccountId: submitTx.InterchainAccountId,
 		Timeout:             submitTx.Timeout,
 	}
+	params := m.Keeper.GetParams(ctx)
+	if uint64(len(submitTx.Msgs)) > params.GetMsgSubmitTxMaxMessages() {
```
swelf19: you validate only the wasm bindings. But user(contract) still can craft the proto(stargate) msg and ignore the limitation. am i right?

foxpy: I have no idea.

CC @pr0n00gler @oopcode

foxpy:
Actually, this loop might be removed entirely. I will open this PR again when I finish investigating.

wasmbinding/message_plugin.go
foxpy: Moved validation from `PerformSubmitTx()` to keeper's `SubmitTx()`

foxpy:
> Actually, this loop might be removed entirely.

We cannot get rid of this loop since protobuf's `Any` uses custom JSON (un)marshalling logic and we heavily rely on (un)marshalling it.

pr0n00gler (approved):

swelf19 (approved):

PR #78, opened 2022-10-17
feat: kv query keys validation

ratik:
issue No. 13 in the audit report

https://p2pvalidator.atlassian.net/browse/NTRN-222

Add validation for keys in submission kv query


x/interchainqueries/types/tx.go
```diff
@@ -91,6 +91,22 @@ func (msg MsgRegisterInterchainQuery) ValidateBasic() error {
 		return sdkerrors.Wrap(ErrInvalidQueryType, "invalid query type")
 	}
 
+	if InterchainQueryType(msg.QueryType).IsKV() {
+		if len(msg.Keys) == 0 {
+			return sdkerrors.Wrap(ErrEmptyKeys, "keys cannot be empty")
+		}
+
+		//// check msg.Keys has path
```
zavgorodnii: `//`

zavgorodnii (approved):

x/interchainqueries/types/tx.go
```diff
@@ -91,6 +91,22 @@ func (msg MsgRegisterInterchainQuery) ValidateBasic() error {
 		return sdkerrors.Wrap(ErrInvalidQueryType, "invalid query type")
 	}
 
+	if InterchainQueryType(msg.QueryType).IsKV() {
+		if len(msg.Keys) == 0 {
+			return sdkerrors.Wrap(ErrEmptyKeys, "keys cannot be empty")
+		}
+
+		// check msg.Keys has path
+		for _, key := range msg.Keys {
```
mmikev: There is a couple of paragraphs in the audit report about unbounded loops. What do you think about this one?

zavgorodnii: true; did we introduce a cap for kv keys? if not, we should, and we should also check that cap during query creation

ratik: so should I add one more env var to store this param or lets say set it like constant = 128 or something?

mmikev: I personally think that constant is fine. But maybe there are some other thoughts

ratik: just a thought that we have pile of shit from constants already, lol

mmikev: If we are going to stick with governance params then we can't check `.Keys` length in `ValidateBasic()`. That's why I like constant solution more

foxpy: please just make a constant, it is the easiest one to implement and change in future, if desired.

x/interchainqueries/keeper/msg_server.go
```diff
@@ -132,6 +141,11 @@ func (k msgServer) UpdateInterchainQuery(goCtx context.Context, msg *types.MsgUp
 		return nil, sdkerrors.Wrap(sdkerrors.ErrUnauthorized, "authorization failed")
 	}
 
+	if types.InterchainQueryType(query.QueryType).IsKV() && uint64(len(msg.NewKeys)) > MaxKVQueryKeysCount {
```
mmikev: This check should be in `.ValidateBasic` method of the message

x/interchainqueries/keeper/msg_server.go
```diff
@@ -49,6 +53,11 @@ func (k msgServer) RegisterInterchainQuery(goCtx context.Context, msg *types.Msg
 		return nil, sdkerrors.Wrapf(types.ErrInvalidConnectionID, "failed to get connection with ID '%s': %v", msg.ConnectionId, err)
 	}
 
+	if types.InterchainQueryType(msg.QueryType).IsKV() && uint64(len(msg.Keys)) > MaxKVQueryKeysCount {
```
mmikev: This check should be in `.ValidateBasic` method of the message

pr0n00gler (changes requested):

foxpy (changes requested):
Please run `make format`

x/interchainqueries/types/tx.go
```diff
@@ -159,6 +180,10 @@ func (msg MsgUpdateInterchainQueryRequest) ValidateBasic() error {
 		return sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, "one of new_keys or new_update_period should be set")
 	}
 
+	if uint64(len(msg.NewKeys)) > MaxKVQueryKeysCount {
+		return sdkerrors.Wrapf(ErrTooManyKVQueryKeys, "keys count cannot be more than %d", MaxKVQueryKeysCount)
+	}
```
sotnikov-s, line 177: we have such a thorough validation on query registration, but on query update we barely validate it, only the length. I believe we should have the same validation on update as we have on registration, e.g. move the algorithm to a standalone func and reuse in both places

sotnikov-s (changes requested):

x/interchainqueries/types/tx.go
ratik: moved

sotnikov-s: No no, I mean, we have the following checks on query registration: keys number is not too much, not zero, each key payload. But on update we only check whether there are not too many keys, which means I can possibly pass keys with invalid payload.

x/interchainqueries/types/tx_test.go
```diff
@@ -345,6 +403,18 @@ func TestMsgUpdateQueryRequestValidate(t *testing.T) {
...
+			func() sdktypes.Msg {
+				return &iqtypes.MsgUpdateInterchainQueryRequest{
+					QueryId:         1,
+					NewKeys:         craftKVKeys(200),
+					NewUpdatePeriod: 0,
+					Sender:          TestAddress,
+				}
+			},
+			iqtypes.ErrTooManyKVQueryKeys,
+		},
```
sotnikov-s, line 441: Please expand the update test according to the latest changes in terms of update payload validation

pr0n00gler (approved):

sotnikov-s (approved):

swelf19 (approved):

foxpy (approved):

PR #82, opened 2022-10-19
feat: add ContractManager

albertandrejev:
Refactor usage of SudoHandler and wasmKeeper in Interchain modules (tx and icq)

### Summary

There are two things that are a bit messy in structure of neutron: 
- cyclic dependency on wasmKeeper and interchainTxKeper/interchainQueriesKeeper (see app.go instantiation)
- passing wasmKeeper into interchainTxKeeper and also creating sudoHandler inside. 

I’m not sure that first problem is solvable easily. Needs a bit of research.
For the second problem I offer to rename sudoHandler → ContractManager with HasContractInfo and Sudo methods in it. And use it everywhere in interchain tx and query modules.

### DoD

Interchain tx and queries module don’t use wasmKeeper directly. SudoHandler renamed to the ContractManager (not sure about name?) and used in both interchain tx and query modules 

https://p2pvalidator.atlassian.net/browse/NTRN-188


internal/sudo/contract_manager.go
```diff
@@ -75,37 +74,72 @@ type OpenAckDetails struct {
 	CounterpartyVersion   string `json:"counterparty_version"`
 }
 
+type ContractMethods interface {
```
sotnikov-s, line 77: I personally don't like the name. it looks like an enum, but is an interface

internal/sudo/contract_manager.go
```diff
@@ -75,37 +74,72 @@ type OpenAckDetails struct {
...
+	HasContractInfo(sdk.Context, sdk.AccAddress) bool
+	Sudo(sdk.Context, sdk.AccAddress, []byte) ([]byte, error)
+}
+
+var _ ContractMethods = (*ContractManager)(nil)
```
sotnikov-s, line 82: what's the point of this check? the ContractManager is never used as ContractMethods

internal/sudo/contract_manager.go
```diff
@@ -75,37 +74,72 @@ type OpenAckDetails struct {
...
+
+type ContractManager struct {
+	wasmKeeper ContractMethods
```
sotnikov-s, line 85: what do you think about renaming the field so it looks more like the interface name? wasmKeeper imo doesn't fit here

internal/sudo/contract_manager.go
```diff
@@ -75,37 +74,72 @@ type OpenAckDetails struct {
...
+}
+
 type Handler struct {
```
sotnikov-s, line 88: wdyt about splitting the file for a better readability into three files:
1. all models e.g. results, responses etc.;
2. ContractManager + ContractMethods;
3. Handler.

or at least move the Handler definition and its methods after the ContractManager's ones?

internal/sudo/contract_manager.go
```diff
@@ -75,37 +74,72 @@ type OpenAckDetails struct {
...
 	return Handler{
-		moduleName: moduleName,
-		wasmKeeper: wasmKeeper,
+		moduleName:      moduleName,
+		contractManager: cm,
 	}
 }
 
-func (s *Handler) SudoResponse(
+func (cm *ContractManager) SetWasmKeeper(wasmKeeper ContractMethods) {
```
sotnikov-s, line 104: please cover as much structs and methods with comments as you can

internal/sudo/contract_manager.go
sotnikov-s: do we really need this interface? seems like it can be removed and wasmKeeper used as it is inside the ContractManager

internal/sudo/contract_manager.go
sotnikov-s: do we even need both Handler and ContractManager?
1. they are very much coupled to each other (ContractManager provides a constructor for the Handler, and the constructor requires a ContractManager);
2. ContractManager is very thin and it mostly proxies requests to a wasm.Keeper.

what you think about merging them into a single struct that will:
1. have access to chain contracts through a wasm.Keeper;
2. call sudo to these contracts.

that could simplify structure a lot imo

x/interchaintxs/keeper/keeper.go
```diff
@@ -69,8 +65,8 @@ func NewKeeper(
 
 		icaControllerKeeper: icaControllerKeeper,
 		scopedKeeper:        scopedKeeper,
-		wasmKeeper:          wasmKeeper,
-		sudoHandler:         sudo.NewSudoHandler(wasmKeeper, types.ModuleName),
+		contractManager:     contractManager,
+		sudoHandler:         contractManager.NewSudoHandler(types.ModuleName),
```
sotnikov-s, line 69: such pair instantiations are the reason why I've suggested to merge these two structures above, I find them not pretty :)

sotnikov-s: and I don't think this solves the second issue of the task this PR is related to:

> passing wasmKeeper into interchainTxKeeper and also creating sudoHandler inside.

sotnikov-s (changes requested):

x/interchaintxs/keeper/keeper.go
albertandrejev: The problem here is that I need to pass `ModuleName` inside it. So I need to instantiate new instance of `SudoHandler` inside keeper. I will try to find another, more suitable approach, but at the moment this is the most convenient way.

internal/sudo/contract_manager.go
```diff
@@ -75,37 +74,72 @@ type OpenAckDetails struct {
...
+	cm.wasmKeeper = wasmKeeper
+}
+
+func (cm *ContractManager) HasContractInfo(ctx sdk.Context, contractAddress sdk.AccAddress) bool {
+	if cm.wasmKeeper == nil {
+		panic("wasmKeeper pointer is nil")
+	}
```
foxpy, line 111: What's the point of this check and panic? Without this check, you will get a panic on nil pointer dereference anyway.

foxpy (changes requested):

internal/sudo/contract_manager.go
albertandrejev: Simply to have more meaningful error.

internal/sudo/contract_manager.go
albertandrejev: This approach will simplify testing and to add additional limits to wasmKeeper methods , otherwise it will be required to use something like mockery to prepare objects for testing inside unit test. But maybe there is more convenient way to pass keeper object and to mock/test this keeper's methods.

internal/sudo/contract_manager.go
albertandrejev: Yes you are right, it was left from previous tries, I need to alter ContractMethods method and add `NewSudoHandler` to it, this helps me to use this interface in the keepers. to pass ContractManager.

internal/sudo/contract_manager.go
albertandrejev: > I personally don't like the name. it looks like an enum, but is an interface

Let's discuss it

internal/sudo/contract_manager.go
albertandrejev: agree

internal/sudo/contract_manager.go
albertandrejev: > wdyt about splitting the file for a better readability into three files:
> ...

Good point.

albertandrejev: > 1. they are very much coupled to each other (ContractManager provides a constructor for the Handler, and the constructor requires a ContractManager);

Yes, I will decrease coupling by adding `NewSudoHandler` to the `ContractMethods` interface.

> 2. ContractManager is very thin and it mostly proxies requests to a wasm.Keeper.

That's was the idea of Contract Manager to reduce dependency from wasm.Keeper. It inverts control from keeper to interface and interface realization.

internal/sudo/contract_manager.go
foxpy: eh, I guess it will be meaningful enough anyway without panicing manually. You are free to close this thread without deleting this code though. I will not open this thread again.

PR #84, opened 2022-10-25
feat: ignore sudo error on ACK

albertandrejev:
In this PR fix for issue #2 from the audit report will be added.

In case of Sudo handler error state of the contract will be reverted and ACK as well as acknowledgement identifiers will be stored in the store. New module to get list of failed acks ids with contracts addresses will be added. Module name will be `contractmanager`

1. [Docs](https://github.com/neutron-org/neutron-docs/pull/30)
2. [Tests](https://github.com/neutron-org/neutron-integration-tests/pull/33)
3. [Relayer fix](https://github.com/neutron-org/neutron-query-relayer/pull/56)

More info: https://p2pvalidator.atlassian.net/browse/NTRN-253


x/contractmanager/client/cli/query.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	"fmt"
+	// "strings"
+
+	"github.com/spf13/cobra"
+
+	"github.com/cosmos/cosmos-sdk/client"
+	// "github.com/cosmos/cosmos-sdk/client/flags"
+	// sdk "github.com/cosmos/cosmos-sdk/types"
```
foxpy: I would have deleted these commented imports

x/contractmanager/client/cli/query_failure.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/cosmos/cosmos-sdk/client/flags"
+	"github.com/neutron-org/neutron/x/contractmanager/types"
+	"github.com/spf13/cobra"
+)
+
+func CmdAllFailures() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "all-failures",
+		Short: "shows all failures",
+		Args:  cobra.RangeArgs(0, 1),
```
foxpy, line 16: Why do you define from 0 to 1 arguments? I don't see any argument handling code further.

x/contractmanager/client/cli/query_failure_test.go
```diff
@@ -0,0 +1,169 @@
...
+
+	sdktypes "github.com/cosmos/cosmos-sdk/types"
+	"github.com/neutron-org/neutron/testutil/contractmanager/network"
+	"github.com/neutron-org/neutron/testutil/contractmanager/nullify"
+	"github.com/neutron-org/neutron/x/contractmanager/client/cli"
+	"github.com/neutron-org/neutron/x/contractmanager/types"
+)
+
+// Prevent strconv unused error
+var _ = strconv.IntSize
```
foxpy: Seems like you could just delete it now.

foxpy (changes requested):

foxpy (commented):
Tested: this PR does NOT break integration tests.

x/contractmanager/client/cli/query_failure.go
albertandrejev: Forgot to remove it. Thanks for pointing

foxpy (approved):
utACK

albertandrejev:
> utACK

?

albertandrejev:
oops

proto/contractmanager/genesis.proto
```diff
@@ -0,0 +1,28 @@
...
+
+message Failure {
+  // Address of the failed contract
+  string address = 1; 
+  // id of the failure under specific address
+  uint64 id = 2;
+  // ACK id to restore
+  uint64 ackID = 3; 
+  // Ackonowledgement type
+  string ackType = 4;
```
sotnikov-s: re `ackID`, `ackType`, and `failureList` below: it's recommended to write field names in snake case:
https://developers.google.com/protocol-buffers/docs/style#message_and_field_names

proto/contractmanager/query.proto
```diff
@@ -0,0 +1,58 @@
...
+  // params holds all the parameters of this module.
+  Params params = 1 [(gogoproto.nullable) = false];
+}
+
+message QueryGetFailureRequest {
+	string address = 1;
+}
+
+message QueryGetFailureResponse {
+	repeated Failure failures = 1 [(gogoproto.nullable) = false];
```
sotnikov-s: I believe this query also requires pagination for a single address can have a great number of failures as well. also, since it returns a slice of failures, the name better be GetFailure**s**... or maybe GetFailuresByAddress?

proto/contractmanager/genesis.proto
```diff
@@ -0,0 +1,28 @@
+syntax = "proto3";
+package neutronorg.neutron.contractmanager;
+
+import "gogoproto/gogo.proto";
+import "contractmanager/params.proto";
+// this line is used by starport scaffolding # genesis/proto/import
+
+option go_package = "github.com/neutron-org/neutron/x/contractmanager/types";
+
+message Failure {
```
sotnikov-s, line 11: please describe the `Failure` message itself

testutil/contractmanager/keeper/contractmanager.go
```diff
@@ -0,0 +1,53 @@
...
+	typesparams "github.com/cosmos/cosmos-sdk/x/params/types"
+	"github.com/neutron-org/neutron/x/contractmanager/keeper"
+	"github.com/neutron-org/neutron/x/contractmanager/types"
+	"github.com/stretchr/testify/require"
+	"github.com/tendermint/tendermint/libs/log"
+	tmproto "github.com/tendermint/tendermint/proto/tendermint/types"
+	tmdb "github.com/tendermint/tm-db"
+)
+
+func ContractmanagerKeeper(t testing.TB) (*keeper.Keeper, sdk.Context) {
```
sotnikov-s: ContractmanagerKeeper -> Contract**M**anagerKeeper

testutil/contractmanager/keeper/contractmanager.go
```diff
@@ -0,0 +1,53 @@
...
+	require.NoError(t, stateStore.LoadLatestVersion())
+
+	registry := codectypes.NewInterfaceRegistry()
+	cdc := codec.NewProtoCodec(registry)
+
+	paramsSubspace := typesparams.NewSubspace(cdc,
+		types.Amino,
+		storeKey,
+		memStoreKey,
+		"ContractmanagerParams",
```
sotnikov-s: ContractmanagerParams -> Contract**M**anagerParams

testutil/contractmanager/network/network.go
```diff
@@ -0,0 +1,80 @@
...
+)
+
+type (
+	Network = network.Network
+	Config  = network.Config
+)
+
+// New creates instance with fully configured cosmos network.
+// Accepts optional config, that will be used in place of the DefaultConfig() if provided.
+func New(t *testing.T, configs ...network.Config) *network.Network {
```
sotnikov-s, line 30: `configs ...network.Config` — is it a kind of general cosmos pattern? kinda strange taking into account the way we use this New and the returned value. I ask because I can see that we use the same network constructor in ICQ and ICTx modules whereas there are more clean (imo) ways to define optional parameters (e.g. `With...` methods, or several constructors, or even a `nil` value suppport)

this comment is just out of curiosity, an informal one

x/contractmanager/client/cli/query.go
```diff
@@ -0,0 +1,31 @@
...
+		Short:                      fmt.Sprintf("Querying commands for the %s module", types.ModuleName),
+		DisableFlagParsing:         true,
+		SuggestionsMinimumDistance: 2,
+		RunE:                       client.ValidateCmd,
+	}
+
+	cmd.AddCommand(CmdQueryParams())
+	cmd.AddCommand(CmdAllFailures())
+	cmd.AddCommand(CmdContractFailures())
+	// this line is used by starport scaffolding # 1
```
zavgorodnii, line 27: please remove all scaffolding mentions!

x/contractmanager/keeper/failure.go
```diff
@@ -0,0 +1,70 @@
...
+	defer iterator.Close()
+
+	if iterator.Valid() {
+		var val types.Failure
+		k.cdc.MustUnmarshal(iterator.Value(), &val)
+
+		return val.Id + 1
+	}
+
+	return 0
```
zavgorodnii, line 37: are we sure we want to be so silent about this?

x/contractmanager/client/cli/query.go
albertandrejev: It is useful to have them because `ignite` is using it to alter files (to add new messages, queries etc).

x/contractmanager/client/cli/query_failure.go
```diff
@@ -0,0 +1,73 @@
...
+	flags.AddPaginationFlagsToCmd(cmd, cmd.Use)
+	flags.AddQueryFlagsToCmd(cmd)
+
+	return cmd
+}
+
+func CmdContractFailures() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "contract-failures [address]",
+		Short: "shows a failures for specific contract address",
```
sotnikov-s: shows a failures -> shows failures

x/contractmanager/keeper/failure.go
albertandrejev: Yes, because our first ID is 0. We will return 0 in case if there is no any elements with specific contract address in the store.

x/transfer/keeper/keeper.go
```diff
@@ -46,10 +47,12 @@ func NewKeeper(
 	cdc codec.BinaryCodec, key sdk.StoreKey, paramSpace paramtypes.Subspace,
 	ics4Wrapper types.ICS4Wrapper, channelKeeper types.ChannelKeeper, portKeeper types.PortKeeper,
 	authKeeper types.AccountKeeper, bankKeeper types.BankKeeper, scopedKeeper capabilitykeeper.ScopedKeeper,
+	contractmanagerKeeper wrappedtypes.ContractManagerKeeper,
 ) KeeperTransferWrapper {
 	return KeeperTransferWrapper{
 		Keeper: keeper.NewKeeper(cdc, key, paramSpace, ics4Wrapper, channelKeeper, portKeeper,
 			authKeeper, bankKeeper, scopedKeeper),
-		channelKeeper: channelKeeper,
+		channelKeeper:         channelKeeper,
```
zavgorodnii: private after public?

sotnikov-s (changes requested):

x/contractmanager/keeper/sudo.go
```diff
@@ -1,4 +1,4 @@
-package sudo
+package keeper
 
 /*
 Wasm contracts have the special entrypoint called sudo. The main purpose of the entrypoint is to be called from a trusted cosmos module, e.g. via a governance process.
```
sotnikov-s, line 4: the comment needs to be either removed or updated: it still says about the `sudo` package

x/contractmanager/types/keys.go
```diff
@@ -0,0 +1,41 @@
...
+	ModuleName = "contractmanager"
+
+	// StoreKey defines the primary module store key
+	StoreKey = ModuleName
+
+	// RouterKey defines the module's message routing key
+	RouterKey = ModuleName
+
+	// MemStoreKey defines the in-memory store key
+	MemStoreKey = "mem_contractmanager"
```
sotnikov-s: maybe `"mem" + ModuleName`?

oopcode:
[still looking at the recovery code]

don't we need the recovery logic also for ibc transfers?

x/interchainqueries/types/expected_keepers.go
```diff
@@ -18,3 +19,12 @@ type BankKeeper interface {
 	SendCoinsFromAccountToModule(ctx sdk.Context, senderAddr sdk.AccAddress, recipientModule string, amt sdk.Coins) error
 	// Methods imported from bank should be defined here
 }
+
+type ContractManagerKeeper interface {
```
sotnikov-s, line 23: I like this design of contract manager more than in the nearby PR. an idea: what do you think about naming the `contractmanager` module as simple `sudo`?

x/interchaintxs/keeper/keeper.go
```diff
@@ -38,9 +34,8 @@ type (
 		scopedKeeper  capabilitykeeper.ScopedKeeper
 		channelKeeper icatypes.ChannelKeeper
 
-		icaControllerKeeper icacontrollerkeeper.Keeper
-		wasmKeeper          *wasm.Keeper
-		sudoHandler         sudo.Handler
+		icaControllerKeeper   icacontrollerkeeper.Keeper
+		contractmanagerKeeper types.ContractManagerKeeper
```
sotnikov-s: contractmanagerKeeper -> contract**M**anagerKeeper

x/interchainqueries/types/expected_keepers.go
albertandrejev: Nearby PR is not actual anymore. 
Not sure that I understand your idea about renaming, because current module provides not only sudo functionality.

x/interchaintxs/keeper/ibc_handlers.go
```diff
@@ -1,17 +1,59 @@
...
+		k.contractmanagerKeeper.AddContractFailure(ctx, senderAddress.String(), packet.GetSequence(), "ack")
+
+	}
+}
+
+func (k *Keeper) createCachedContext(ctx sdk.Context) (cacheCtx sdk.Context, writeFn func(), newGasMeter sdk.GasMeter) {
+	gasLeft := ctx.GasMeter().Limit() - ctx.GasMeter().GasConsumed()
+	newLimit := gasLeft - GasReserve
+
+	if newLimit > gasLeft {
```
zavgorodnii: i think that this uint overflow should be handled more explicitly, e.g.

```
var newLimit uint64
if GasReserve > gasLeft {
    newLimit = 0
} else {
    newLimit = gasLeft - GasReserve
}
```

x/interchainqueries/types/expected_keepers.go
sotnikov-s: the idea is simple: since the main idea of the ContractManager module is to store contracts sudo-related info, execute sudo calls to contracts and provide queries for getting this data, it's imo meaningful to rename it to Sudo: SudoKeeper, available via cli as `neutrond query sudo ...` etc. Maybe it even makes sense to split this module to two: `ContractManager` module with `HasContractInfo` and `AddContractFailure` methods, and a `Sudo` one with all the `Sudo...` methods. Out of this task scope. Wyt?

testutil/contractmanager/network/network.go
albertandrejev: This code is autogenerated by ignite so I guess it was created in this way to be more generic.

x/interchainqueries/types/expected_keepers.go
albertandrejev: Good idea, but I think that Sudo is too generic name and may be confusing. Also if we will split module into 2 different modules we will:
1. Move back to the previous state of the sudo helpers
2. We will have one module with ability to write to state and to query information from state, and another module without any interfaces to process messages or to query state. I mean this module will not have any ability to directly interact with cli on tx. Not sure that in this case we need to have another module for sudo, because previously we had this solution.

And current approach may help us to cover other contract related interactions in future.

x/contractmanager/client/cli/query_failure.go
```diff
@@ -23,47 +24,17 @@ func CmdAllFailures() *cobra.Command {
 
 			queryClient := types.NewQueryClient(clientCtx)
 
-			params := &types.QueryAllFailureRequest{
-				Pagination: pageReq,
+			address := ""
+			if len(args) > 0 {
```
sotnikov-s, line 28: btw is there a use of command execution with an empty address, i.e. of getting the whole failures happened in the chain? maybe the simple by-address-one is enough?

sotnikov-s (commented):
LGTM, but I'll wait for integration tests before giving an approval

x/contractmanager/client/cli/query_failure.go
albertandrejev: At least you can. Do not see any issue with it.

proto/contractmanager/query.proto
```diff
@@ -0,0 +1,51 @@
...
+	rpc Params(QueryParamsRequest) returns (QueryParamsResponse) {
+		option (google.api.http).get = "/neutron-org/neutron/contractmanager/params";
+	}
+	// Queries a Failure by address.
+	rpc Failures(QueryFailuresRequest) returns (QueryFailuresResponse) {
+		option (google.api.http).get = "/neutron-org/neutron/contractmanager/failure/{address}";
+	}
+
+	// Queries a list of Failure items.
+	rpc AllFailures(QueryFailuresRequest) returns (QueryFailuresResponse) {
```
swelf19: i do not see any difference between `AllFailures` and `Failures`. And code is the same for both methods. Method named `AllFailures` should at least omit the filtering with address params

x/interchaintxs/keeper/ibc_handlers.go
```diff
@@ -1,17 +1,61 @@
...
 	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
 	channeltypes "github.com/cosmos/ibc-go/v3/modules/core/04-channel/types"
 
-	"github.com/neutron-org/neutron/internal/sudo"
+	contractmanagertypes "github.com/neutron-org/neutron/x/contractmanager/types"
 	"github.com/neutron-org/neutron/x/interchaintxs/types"
 )
 
+const (
+	// GasReserve is the amount of gas on the context gas meter we need to reserve in order to add contract failure to keeper
```
swelf19, line 17: lines 16-57 the same in the transfer wrapper and interchain txs, is it possible to reuse the common code?

albertandrejev: This is the problem, I want to have modules to be independent as possible. In order to reuse common code we need to introduce new entity, that will be used between different modules, but I do not want to break this modularity.

proto/contractmanager/query.proto
albertandrejev: the difference in the `address` parameter. But we need to have two grpc paths, one with address parameter another without. I am not sure that if I remove it (I mean second `rpc` endpoint) it will work properly with and without `address` parameter.

x/interchaintxs/keeper/ibc_handlers.go
zavgorodnii: yup, we discussed it with Albert; let's leave it as is

proto/contractmanager/query.proto
swelf19: I got the idea, i mean there is no difference which method to call via (g)rpc, both takes the same arguments(paging, address), both executes the same code. What do you think to swap names `func (k Keeper) Failures` and `func (k Keeper) AllFailures`, its strange, when the more generic method `Failures` calls the more specific `AllFailures` and passes the filter arguments.

I've removed `AllFailures` from proto file, regenerated the code, and `Failures` works properly with empty `address`, but trailing slash is required.

albertandrejev: Tested. If I will remove one of the endpoints then other will stop working. It means if I will remove `AllFailures` rpc, then it will be longer not possible to get list of all failures. Same with other, in this case it will not be possible to get `Failures` for specific address.

albertandrejev: I tried it without slash :)

zavgorodnii (approved):

sotnikov-s (approved):

pr0n00gler (approved):

PR #89, opened 2022-11-13
Fix: transfer from a non-contract account works

pr0n00gler:
This PR fixes the problem when a a non-contract account cannot execute an IBC transfer msg.
IBC fees are not applied for non-contract transfers since there are no sudo happening during ACK/Timeout processing.

**MERGE WITH:**:
* https://github.com/neutron-org/neutron-integration-tests/pull/35

### DIFF
--- a/x/transfer/keeper/keeper.go
+++ b/x/transfer/keeper/keeper.go
@@ -43,7 +43,7 @@ func (k KeeperTransferWrapper) Transfer(goCtx context.Context, msg *wrappedtypes
 	}
 
 	// if the sender is a contract, lock fees.
-	// Because contracts are required to pay fees for the acknoledgements
+	// Because contracts are required to pay fees for the acknowledgements
 	if k.ContractManagerKeeper.HasContractInfo(ctx, senderAddr) {
 		if err := k.FeeKeeper.LockFees(ctx, senderAddr, feetypes.NewPacketID(msg.SourcePort, msg.SourceChannel, sequence), msg.Fee); err != nil {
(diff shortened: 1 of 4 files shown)


x/transfer/keeper/keeper.go
```diff
@@ -42,8 +42,12 @@ func (k KeeperTransferWrapper) Transfer(goCtx context.Context, msg *wrappedtypes
 		)
 	}
 
-	if err := k.FeeKeeper.LockFees(ctx, senderAddr, feetypes.NewPacketID(msg.SourcePort, msg.SourceChannel, sequence), msg.Fee); err != nil {
-		return nil, sdkerrors.Wrapf(err, "failed to lock fees to pay for transfer msg: %v", msg)
+	// if the sender is a contract, lock fees.
+	// Because contracts are required to pay fees for the acknoledgements
```
zavgorodnii: acknowledgements

zavgorodnii:
we should probably fix the cli also (support the custom type)?

pr0n00gler:
> we should probably fix the cli also (support the custom type)?

No, on line https://github.com/neutron-org/neutron/pull/89/files#diff-4799fb6161e4b3b5b9659cb9658afb89208ed27e1157e758aa87562aa083380bR128 we convert original `MsgTransfer` into our `MsgTransfer` with fees (they are set to zero), so the original `Transfer` message is already supported

zavgorodnii:
utack, waiting for integration tests

zavgorodnii (approved):

ratik (approved):

swelf19 (approved):

PR #92, opened 2022-11-16
fix: added missing parts to update transaction filter

ratik:
So these ones were missed

### DIFF
--- a/wasmbinding/bindings/msg.go
+++ b/wasmbinding/bindings/msg.go
@@ -73,9 +73,10 @@ type RemoveInterchainQuery struct {
 type RemoveInterchainQueryResponse struct{}
 
 type UpdateInterchainQuery struct {
-	QueryId         uint64         `json:"query_id,omitempty"`
-	NewKeys         []*types.KVKey `json:"new_keys,omitempty"`
-	NewUpdatePeriod uint64         `json:"new_update_period,omitempty"`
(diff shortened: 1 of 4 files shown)


pr0n00gler (dismissed):

swelf19 (commented):
1) I think we need validate new transaction filter (ValidateTransactionsFilter) 
2) It was hard in the contracts to correspond query type and new query data, but it's possible now on neutron

ratik:
> 1. I think we need validate new transaction filter (ValidateTransactionsFilter)
> 2. It was hard in the contracts to correspond query type and new query data, but it's possible now on neutron

okay. added

swelf19:
> > 1. I think we need validate new transaction filter (ValidateTransactionsFilter)
> > 2. It was hard in the contracts to correspond query type and new query data, but it's possible now on neutron
> 
> okay. added

2) my bad, actually there was a "validation" of query type a few lines below the code you added.  https://github.com/neutron-org/neutron/blob/eb350c5447bbe9db380332efe925aedb14bf8330/x/interchainqueries/keeper/msg_server.go#L148 
i think we need to revert changes back in p.2

swelf19 (approved):

sotnikov-s (approved):

zavgorodnii (approved):

PR #94, opened 2022-11-23
fix: feerefunder genesis validation

oldremez:
According to the audit report:

> Lack of validation for the GenesisState of the feerefunder module
> Severity: Minor
> In `neutron:x/feerefunder/types/genesis.go:20`, the Validate function is not validating FeeInfos and Params parameters. 
> It should ensure that Payer is a correct address and that both PacketId and Fee structs are well constructed.
> Also, the Validate function for Params in `neutron:x/feerefunder/types/params.go:43-45` is not implemented and it is
always returning nil.
> Recommendation
> We recommend implementing validation for the GenesisState of the feerefunder module

Changes:
1. Added validation of the `GenesisState` structure. Everything is basic-validated but Sequence (since it is just a number)
2. Added unit tests to check if validation works properly
3. Fixed an issue with providing the non-zero minimum `RecvFee`
4. `RecvFee` is provided as `nil` to be compatible with the `TestGenesis` test since even if we provide an empty slice as `RecvFee`, `ExportGenesis` gets `RecvFee` as nil and it breaks the tests (while being semantically equivalent). This issue needs to be addressed further.

How to test it:
Use unit tests (`make test`) and regular integration tests pipeline to test the code. No changes in external components were made.

### DIFF
--- a/x/feerefunder/types/genesis.go
+++ b/x/feerefunder/types/genesis.go
@@ -1,11 +1,12 @@
 package types
 
 import (
-// this line is used by starport scaffolding # genesis/types/import
+	// this line is used by starport scaffolding # genesis/types/import
+	wasmtypes "github.com/CosmWasm/wasmd/x/wasm/types"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+	host "github.com/cosmos/ibc-go/v3/modules/core/24-host"
 )
-
-// DefaultIndex is the default global index
-const DefaultIndex uint64 = 1
 
 // DefaultGenesis returns the default genesis state
 func DefaultGenesis() *GenesisState {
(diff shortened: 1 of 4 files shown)


x/feerefunder/types/genesis.go
```diff
@@ -19,6 +20,29 @@ func DefaultGenesis() *GenesisState {
 // failure.
 func (gs GenesisState) Validate() error {
 	// this line is used by starport scaffolding # genesis/types/validate
-
+	for _, info := range gs.FeeInfos {
+		addr, err := sdk.AccAddressFromBech32(info.Payer)
+		if err != nil {
+			return err
```
zavgorodnii: please wrap all errors with a meaningful message, e.g.:

```
if err != nil {
 	return fmt.Errorf("failed to parse the payer address: %w", err)
}
```

p.s. also please add newlines after if err != nil blocks hehe

oldremez: done

zavgorodnii (dismissed):
didn't test, but looks good

x/feerefunder/types/genesis.go
```diff
@@ -19,6 +21,34 @@ func DefaultGenesis() *GenesisState {
 // failure.
 func (gs GenesisState) Validate() error {
 	// this line is used by starport scaffolding # genesis/types/validate
+	for _, info := range gs.FeeInfos {
+		addr, err := sdk.AccAddressFromBech32(info.Payer)
+		if err != nil {
+			return fmt.Errorf("failed to parse the payer address %v: %v", info.Payer, err)
```
swelf19: Pls use the `%w` to wrap the errors instead of `%v`

sotnikov-s: a good practice is wrapping returning errors (notice the `%w` instead of `%v`) to be capable of unwrapping them outside (never used it tbh, but it's better to have such a capability than not):
```go
return fmt.Errorf("failed to parse the payer address %v: %w", info.Payer, err)
```

x/feerefunder/types/genesis.go
```diff
@@ -19,6 +21,34 @@ func DefaultGenesis() *GenesisState {
...
+		if err := host.PortIdentifierValidator(info.PacketId.PortId); err != nil {
+			return fmt.Errorf("port id %v is invalid: %v", info.PacketId.PortId, err)
+		}
 
+		if err := host.ChannelIdentifierValidator(info.PacketId.ChannelId); err != nil {
+			return fmt.Errorf("channel id %v is invalid: %v", info.PacketId.PortId, err)
+		}
+
+		if err := info.Fee.Validate(); err != nil {
+			return fmt.Errorf("invalid fees %v: %v", info.Fee, err)
```
sotnikov-s: a printed structure will be much more descriptive if you print it with `%+v`: https://goplay.space/#kNKZc9hcm3N

or, since `Fee` implements the `Stringer` interface, you can print it using `%s`

x/feerefunder/types/genesis_test.go
```diff
@@ -19,10 +41,162 @@ func TestGenesisState_Validate(t *testing.T) {
...
+					Payer:    TestContractAddressNeutron,
+					PacketId: validPacketId,
+					Fee: types.Fee{
+						RecvFee:    validRecvFee,
+						AckFee:     validAckFee,
+						TimeoutFee: validTimeoutFee,
+					},
+				}},
+			},
+			valid: true,
```
sotnikov-s: I'd be horny to see matching actual errors to expected ones (not simply valid/invalid) in the test cases to make sure the cases fail exactly because of the reasons we design

oldremez: Made just to make you horny (implemented it via substring search, haven't figured out how to make it better)

x/feerefunder/types/genesis.go
oldremez: Done

x/feerefunder/types/genesis.go
oldremez: Done

oldremez: Done

sotnikov-s (approved):

x/feerefunder/types/genesis_test.go
sotnikov-s: I appreciate it

pr0n00gler (approved):

swelf19 (approved):

PR #98, opened 2022-11-24
Fix: proper fees checking

oldremez:
According to the audit report, fees are not checked properly and a fee with single coin that is not in the `MinFees` can bypass the check. This PR is to solve this issue. [(NTRN-275)](https://p2pvalidator.atlassian.net/browse/NTRN-275)

Integration tests are [here](https://github.com/neutron-org/neutron-integration-tests/pull/43).

P.S. I'm still a bit in doubt about the `export_test.go` hack to export private `checkFees` method for unit tests, any recommendations in this regard are very welcome.

### DIFF
--- a/x/feerefunder/keeper/keeper_test.go
+++ b/x/feerefunder/keeper/keeper_test.go
@@ -1,50 +1,18 @@
-package keeper
+package keeper_test
 
 import (
-	"github.com/cosmos/cosmos-sdk/codec"
-	codectypes "github.com/cosmos/cosmos-sdk/codec/types"
-	"github.com/cosmos/cosmos-sdk/store"
-	storetypes "github.com/cosmos/cosmos-sdk/store/types"
-	typesparams "github.com/cosmos/cosmos-sdk/x/params/types"
-	tmproto "github.com/tendermint/tendermint/proto/tendermint/types"
-	tmdb "github.com/tendermint/tm-db"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+	testutil_keeper "github.com/neutron-org/neutron/testutil/keeper"
+	"github.com/pkg/errors"
 	"testing"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/neutron-org/neutron/x/feerefunder/types"
 	"github.com/stretchr/testify/require"
-	"github.com/tendermint/tendermint/libs/log"
 )
 
 func TestKeeperCheckFees(t *testing.T) {
-	storeKey := sdk.NewKVStoreKey(types.StoreKey)
-	memStoreKey := storetypes.NewMemoryStoreKey(types.MemStoreKey)
-
-	db := tmdb.NewMemDB()
-	stateStore := store.NewCommitMultiStore(db)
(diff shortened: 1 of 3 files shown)


zavgorodnii (dismissed):

zavgorodnii (dismissed):

albertandrejev (dismissed):

x/feerefunder/keeper/keeper_test.go
```diff
@@ -0,0 +1,86 @@
+package keeper_test
```
sotnikov-s, line 1: **I KNOW IT'S TREMENDOUS AND TERRIFYING BUT I'D REALLY APPRECIATE IF YOU READ IT** 😄 

re your hack question: haven't seen such approach before — using the `xxx_test` package name for test files. I've even googled it: https://stackoverflow.com/a/31443271 and long story short is: such an approach is applicable when you're making a black-box testing which assumes you use only the exported fields of the entity you're testing. but here you need a white-box testing since you use a private field, but it's not possible because of the import loop between this package and the testutil one.

but I don't really get why we do it this way. I can see that it must have been imported from gaia or something — it's like a pattern how they did it at the point we imported it. we have a standalone folder `testutil` which looks big but actually each file there contains less than 100 lines of code. moreover, it's only the `Keepers` that are module-specific there, and the rest of the files (nullify, network, and sample) are 100% same.

so, what I suggest to do in scope of a separate task:
1. since `nullify`, `network`, and `sample` files from `testutil` are 100% same and don't import any neutron module-specific things, we can keep them in the root of the `testutil` folder in the eponymous package and be safe from import cycles;
2. the only module-specific things from `testutil` — `Keepers` — we can move to the `keeper_test.go` files of the modules and keep the original package names in the `keeper_test.go` files (without the `_test` suffix)

------
**upd:** okay, now I see that this whole trick is about creating a test `Keeper` in the `genesis_test.go` files, and it makes it impossible to (have a `Keeper` builder func inside a module's keeper package) && (import the `Keeper` builder func from the module root, i.e. from the `genesis_test.go`). I really don't like the way it's designed, but I couldn't come up with a perfect solution for it. it's either like this **or** code duplications **or** `Keeper` builder funcs public for modules. frankly, I'd like more to be it as described in the latter: we could have, say, a `DefaultKeeper` func public for keeper packages which creates such a `Keeper` visible from both inside and outside, so by this we will be possible to get rid of the `xxx_test` package names and `testutil` keepers. wyt?

x/feerefunder/keeper/keeper_test.go
```diff
@@ -0,0 +1,86 @@
...
+		},
+	})
+
+	for _, tc := range []struct {
+		desc  string
+		fees  *types.Fee
+		valid bool
+	}{
+		{
+			desc: "single proper denom but insufficient",
```
sotnikov-s: also: please stick with the naming pattern: SingleProperDenomButInsufficient. here and in the rest of the tests

oldremez: fixed

zavgorodnii (approved):

ratik (approved):

sotnikov-s (approved):
integration and unit tests pass

PR #95, opened 2022-11-24
feat: NTRN-267. Add RemoveInterchanQuery cli support, fix proto package names and query wasmbinding

albertandrejev:
This PR will add ability to remove interchain query via CLI. Also it fixes proto package name and query wasmbinding (https://p2pvalidator.atlassian.net/browse/NTRN-263).

PR #106, opened 2022-11-29
fix: missing deposit array into registered query #NTRN-263

ratik:
So seems some strange shit happened and we messed with the field names and so on
To test it 
1. run cosmopark
2. get bash script from the task https://p2pvalidator.atlassian.net/browse/NTRN-263
3. copy interchain_queries contract and the bash script into it
4. apt-get update && apt-get install jq -y
5. run bash script

### DIFF
--- a/wasmbinding/bindings/query.go
+++ b/wasmbinding/bindings/query.go
@@ -75,7 +75,7 @@ type RegisteredQuery struct {
 	// The remote chain last block height when the query result was updated.
 	LastSubmittedResultRemoteHeight uint64 `json:"last_submitted_result_remote_height"`
 	// Amount of coins deposited for the query.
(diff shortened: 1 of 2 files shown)


albertandrejev:
Already fixed in the https://github.com/neutron-org/neutron/pull/95/files#diff-84abdccc6bcefcdb026d549c9ccf24a612648edffaabefbe8df24c4bfbcf6ece

albertandrejev:
Do we still need this PR? Because I added this missed code to the https://github.com/neutron-org/neutron/pull/95/files#diff-84abdccc6bcefcdb026d549c9ccf24a612648edffaabefbe8df24c4bfbcf6ece already.

zavgorodnii (approved):

PR #107, opened 2022-11-30
fix: NTRN-278. Fix chains startup to work with interchain security

albertandrejev:
Make chains startup to work with latest version of neutron with CCV modules.

Related PRs:
- https://github.com/neutron-org/neutron-integration-tests/pull/46
- https://github.com/neutron-org/neutron-contracts/pull/67


network/init.sh
```diff
@@ -2,6 +2,7 @@
 
 ADMIN_ADDRESS=neutron14hj2tavq8fpesdwxxcu44rty3hh90vhujrvcmstl4zr3txmfvw9s5c2epq
 BINARY=neutrond
+GAIAD_BINARY=gaiad
```
foxpy: Let's rename BINARY → NEUTRON_BINARY as well?

foxpy: or maybe even NEUTRON**D**_BINARY :thinking:

zavgorodnii (approved):
tested

network/dockerbuilds/Dockerfile.gaia
```diff
@@ -0,0 +1,22 @@
+# syntax=docker/dockerfile:1
```
swelf19, line 1: don't exactly understand, where do we use this file?

network/init.sh
```diff
@@ -96,8 +96,6 @@ sed -i -e 's/prometheus-retention-time = 0/prometheus-retention-time = 1000/g' $
 
 sed -i -e 's#"tcp://0.0.0.0:26656"#"tcp://0.0.0.0:'"$P2PPORT_2"'"#g' $CHAIN_DIR/$CHAINID_2/config/config.toml
 sed -i -e 's#"tcp://127.0.0.1:26657"#"tcp://0.0.0.0:'"$RPCPORT_2"'"#g' $CHAIN_DIR/$CHAINID_2/config/config.toml
-sed -i -e 's/timeout_commit = "5s"/timeout_commit = "1s"/g' $CHAIN_DIR/$CHAINID_2/config/config.toml
```
swelf19, line 99: what is the purpose of increasing timeouts? just make chains block time differs?

Dockerfile
```diff
@@ -1,3 +1,12 @@
+# syntax=docker/dockerfile:1
```
swelf19, line 1: seems like we do not use `start-cosmopark` from this repo, our cosmopark configuration moved to integration tests. May be we can get rid of the dead code?

swelf19 (approved):

Dockerfile
zavgorodnii: this actually makes sense

network/init.sh
foxpy: IMO, having different block times might expose some races in contracts, so maybe we should have increased timeout? (I cannot prove what I am saying, I am just speculating).

Dockerfile
zavgorodnii: we'll do that is a separate pr

network/init.sh
albertandrejev: The reason why I used different block time for different parts of the infrastructure is to have environment similar to real world.

PR #117, opened 2022-12-26
feat: burn fees or send to treasury #NTRN-300

foxpy:
Merge with https://github.com/neutron-org/neutron-sdk/pull/74 and https://github.com/neutron-org/neutron-integration-tests/pull/57


pr0n00gler (changes requested):
Why did you decide to split fees manually in the ante handler, and not use already split amount of consumer fees [here](https://github.com/cosmos/interchain-security/blob/v0.2.0/x/ccv/consumer/keeper/distribution.go#L35) and do all the work in EndBlocker?

Sorry, it should be a comment, not `request changes`

foxpy:
> Why did you decide to split fees manually in the ante handler, and not use already split amount of consumer fees [here](https://github.com/cosmos/interchain-security/blob/v0.2.0/x/ccv/consumer/keeper/distribution.go#L35) and do all the work in EndBlocker?

Simplified that part :+1:

proto/feeburner/query.proto
```diff
@@ -0,0 +1,46 @@
...
+// Query defines the gRPC querier service.
+service Query {
+  // Parameters queries the parameters of the module.
+  rpc Params(QueryParamsRequest) returns (QueryParamsResponse) {
+    option (google.api.http).get = "/neutron/feeburner/params";
+  }
+
+  // TotalBurnedNeutronsAmount queries total amount of burned neutron fees.
+  rpc TotalBurnedNeutronsAmount(QueryTotalBurnedNeutronsAmountRequest) returns (QueryTotalBurnedNeutronsAmountResponse) {
+    option (google.api.http).get = "/neutron/feeburner/totalburnedneutronsamount";
```
albertandrejev: maybe we can use more user friendly path here, with delimiters?

x/feeburner/client/cli/query_total_burned_neutrons.go
```diff
@@ -0,0 +1,34 @@
...
+
+	"github.com/cosmos/cosmos-sdk/client"
+	"github.com/cosmos/cosmos-sdk/client/flags"
+	"github.com/neutron-org/neutron/x/feeburner/types"
+	"github.com/spf13/cobra"
+)
+
+func CmdQueryTotalBurnedNeutrons() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "total-burned-neutrons",
```
albertandrejev: I think you can use same endpoint name for the REST call

x/feeburner/types/params.go
```diff
@@ -0,0 +1,149 @@
...
+var _ paramtypes.ParamSet = (*Params)(nil)
+
+var (
+	KeyNeutronDenom     = []byte("NeutronDenom")
+	DefaultNeutronDenom = "stake"
+	KeyGaiaDenom        = []byte("GaiaDenom")
+	DefaultGaiaDenom    = "uatom"
+	KeyTreasuryAddress  = []byte("TreasuryAddress")
+	// XXX: that's just a random address, I think there should be something better for a default
+	DefaultTreasuryAddress = "neutron12m0f7r5wzuskcuzrreppjgsu8fvzkqced7uevn"
```
albertandrejev: Treasury should be added in the genesis, so we need to put here genesis address of the treasury contract. We need to have instructions for genesis generation, or update `init.sh` to get contract address from there if it is possible.

proto/feeburner/params.proto
```diff
@@ -0,0 +1,17 @@
...
+
+option go_package = "github.com/neutron-org/neutron/x/feeburner/types";
+
+// Params defines the parameters for the module.
+message Params {
+  option (gogoproto.goproto_stringer) = false;
+  // Defines Neutron denom
+  string neutron_denom = 1;
+  // Defines Gaia denom
+  string gaia_denom = 2;
```
albertandrejev: why do you need gaia denom here?

proto/feeburner/tx.proto
```diff
@@ -0,0 +1,13 @@
+syntax = "proto3";
+package neutron.feeburner;
+
+// this line is used by starport scaffolding # proto/tx/import
+
+option go_package = "github.com/neutron-org/neutron/x/feeburner/types";
+
+// Msg defines the Msg service.
+service Msg {
+  // this line is used by starport scaffolding # proto/tx/rpc
```
albertandrejev: I am pretty sure that you can get rid of tx implementation in this module.

x/feeburner/client/cli/tx.go
```diff
@@ -0,0 +1,34 @@
...
+	"time"
+
+	"github.com/spf13/cobra"
+
+	"github.com/cosmos/cosmos-sdk/client"
+	// "github.com/cosmos/cosmos-sdk/client/flags"
+	"github.com/neutron-org/neutron/x/feeburner/types"
+)
+
+var DefaultRelativePacketTimeoutTimestamp = uint64((time.Duration(10) * time.Minute).Nanoseconds())
```
albertandrejev: Please remove this file, we do not have txs for this module so it maybe confusing to have it in the tx section of the cli.

proto/feeburner/params.proto
```diff
@@ -0,0 +1,17 @@
...
+  // Defines Treasury address
+  string treasury_address = 3;
```
albertandrejev: do we need treasury address here? What for? for informational purpose only?

foxpy: Well, if you say that treasury address should be in genesis, I will remove it from params then.

proto/feeburner/params.proto
foxpy: Because it is an IBC denom and it should be configured? or can we just calculate it in advance and set it in genesis, too?

proto/feeburner/params.proto
albertandrejev: Sorry for confusing. No need to remove it from params, because it should be set via params in the genesis, and can be updated later via PramsChange proposal. My question was about do we really need this information at all? I saw that you are sending coins from "Gaia" demon to the contract, but is this required? Maybe we can choose other mechanics?

proto/feeburner/params.proto
albertandrejev: No, not sure that we can predefine it in the genesis. But what is the reason of having it in the params?  Because it is possible to have any coin as fee coin (osmo, juno, xprt etc) and not only atom. So we need to have other approach or to have list of 3rd party denoms here.

proto/feeburner/query.proto
```diff
@@ -0,0 +1,46 @@
...
+}
+
+// QueryTotalBurnedNeutronsAmountRequest is request type for the
+// Query/QueryTotalBurnedNeutronsAmount method.
+message QueryTotalBurnedNeutronsAmountRequest {}
+
+// QueryTotalBurnedNeutronsAmountResponse is response type for the
+// Query/QueryTotalBurnedNeutronsAmount method.
+message QueryTotalBurnedNeutronsAmountResponse {
+  TotalBurnedNeutronsAmount total_burned_neutrons_amount = 2 [ (gogoproto.nullable) = false ];
```
albertandrejev: Can we change index of the variable to 1?

proto/feeburner/params.proto
zavgorodnii: I think we can simplify by saying that all NTRN tokens are burned, and all non-NTRN tokens are sent to treasury. @foxpy can you implement this?

proto/feeburner/params.proto
zavgorodnii: This implementation complies with the specification: https://www.notion.so/Design-Minimal-Design-v0-6-1-59dac6d17c5e4505bf2c9786425d86bb#f9af00c053bb4ab9b704bab100844cb8

(the treasury is a smart-contract)

testutil/feeburner/keeper/feeburner.go
```diff
@@ -0,0 +1,53 @@
...
+	typesparams "github.com/cosmos/cosmos-sdk/x/params/types"
+	"github.com/neutron-org/neutron/x/feeburner/keeper"
+	"github.com/neutron-org/neutron/x/feeburner/types"
+	"github.com/stretchr/testify/require"
+	"github.com/tendermint/tendermint/libs/log"
+	tmproto "github.com/tendermint/tendermint/proto/tendermint/types"
+	tmdb "github.com/tendermint/tm-db"
+)
+
+func FeeburnerKeeper(t testing.TB) (*keeper.Keeper, sdk.Context) {
```
zavgorodnii, line 20: Is this a constructor? Maybe `NewFeeburnerKeeper()` then?

x/feeburner/module.go
```diff
@@ -0,0 +1,197 @@
...
+	balance = am.bankKeeper.GetBalance(ctx, moduleAddr, params.GaiaDenom)
+	if !balance.IsZero() {
+		err := am.bankKeeper.SendCoins(
+			ctx,
+			moduleAddr,
+			sdk.MustAccAddressFromBech32(params.TreasuryAddress),
+			sdk.Coins{balance},
+		)
+		if err != nil {
+			// TODO: maybe panic or print warning?
```
zavgorodnii: i think that we should panic here. @albertandrejev @pr0n00gler  wdyt?

proto/feeburner/params.proto
foxpy: > all NTRN tokens are burned

that means `NeutronDenom` param should be left intact.

> all non-NTRN tokens are sent to treasury

I think I can implement that using `ViewKeeper`'s `GetAllBalances` function, of course if it does what I expect it to.

testutil/feeburner/keeper/feeburner.go
foxpy: That's actually automatically generated by ignite, and naming convention for this function is the same across testutils for all other modules.

x/feeburner/module.go
```diff
@@ -0,0 +1,197 @@
...
+
+// BeginBlock contains the logic that is automatically triggered at the beginning of each block
+func (am AppModule) BeginBlock(_ sdk.Context, _ abci.RequestBeginBlock) {}
+
+// EndBlock contains the logic that is automatically triggered at the end of each block
+func (am AppModule) EndBlock(ctx sdk.Context, _ abci.RequestEndBlock) []abci.ValidatorUpdate {
+	moduleAddr := am.accountKeeper.GetModuleAddress(consumertypes.ConsumerRedistributeName)
+	if moduleAddr == nil {
+		// TODO: maybe panic or print warning?
+		return []abci.ValidatorUpdate{}
```
albertandrejev: I think it should panic here.

x/feeburner/module.go
albertandrejev: it will not break our tokenomics if it fails. Right now we do not have any scenario what to do with non-NTRN denoms, don't we? So I think we can print warning here.

x/feeburner/module.go
```diff
@@ -0,0 +1,197 @@
...
+
+	params := am.keeper.GetParams(ctx)
+
+	balance := am.bankKeeper.GetBalance(ctx, moduleAddr, params.NeutronDenom)
+	if !balance.IsZero() {
+		am.keeper.RecordBurnedFees(ctx, balance)
+		err := am.bankKeeper.BurnCoins(ctx, consumertypes.ConsumerRedistributeName, sdk.Coins{balance})
+		if err != nil {
+			// TODO: maybe panic or print warning?
+			return []abci.ValidatorUpdate{}
```
albertandrejev: panic here as well.

x/feeburner/module.go
albertandrejev: it will not break our tokenomics if it fails. Right now we do not have any scenario what to do with non-NTRN denoms, don't we? So I think we can print warning here.

testutil/feeburner/keeper/feeburner.go
zavgorodnii: aaah okay

x/feeburner/module.go
oopcode: okay, let it be so

proto/feeburner/total_burned_neutrons_amount.proto
```diff
@@ -0,0 +1,16 @@
...
+import "gogoproto/gogo.proto";
+
+option go_package = "github.com/neutron-org/neutron/x/feeburner/types";
+
+// TotalBurnedNeutronsAmount defines total amount of burned neutron fees
+message TotalBurnedNeutronsAmount {
+  cosmos.base.v1beta1.Coin coin = 1 [
+    (gogoproto.moretags) = "yaml:\"coin\"",
+    (gogoproto.nullable) = false,
+    (gogoproto.castrepeated) = "github.com/cosmos/cosmos-sdk/types.Coin"
```
albertandrejev: Do we really need `castrepeated` here?

x/feeburner/keeper/keeper.go
```diff
@@ -0,0 +1,76 @@
...
+func (k Keeper) GetTotalBurnedNeutronsAmount(ctx sdk.Context) types.TotalBurnedNeutronsAmount {
+	store := ctx.KVStore(k.storeKey)
+
+	var totalBurnedNeutronsAmount types.TotalBurnedNeutronsAmount
+	bzTotalBurnedNeutronsAmount := store.Get(KeyBurnedFees)
+	if bzTotalBurnedNeutronsAmount != nil {
+		k.cdc.MustUnmarshal(bzTotalBurnedNeutronsAmount, &totalBurnedNeutronsAmount)
+	}
+
+	if totalBurnedNeutronsAmount.Coin.Denom == "" {
```
albertandrejev, line 67: When it is possible to have this situation? I have doubts about this line.

proto/feeburner/total_burned_neutrons_amount.proto
foxpy: We absolutely don't need it here, I will remove it. Thank you!

x/feeburner/keeper/keeper.go
foxpy: During processing of the very first block. I couldn't figure out how to code this part in a cleaner way.

albertandrejev (dismissed):

zavgorodnii (dismissed):

pr0n00gler (dismissed):

pr0n00gler (approved):

albertandrejev (approved):

PR #125, opened 2023-01-13
Fix: rename denom stake -> untrn NTRN-316

quasisamurai:
1.  [neutron](https://github.com/neutron-org/neutron/pull/125)
2. [ICQ](https://github.com/neutron-org/neutron-query-relayer/pull/71)
3. [integration tests](https://github.com/neutron-org/neutron-integration-tests/pull/60)

[ACTIONS TEST RUN](https://github.com/neutron-org/neutron-tests/actions/runs/3940223175/jobs/6839531213)


foxpy (changes requested):
Some of `stake` mentions haven't been replaced, you can find them with `grep --exclude-dir=.git stake -R .`

network/init.sh
```diff
@@ -103,7 +103,7 @@ DAO_INIT='{
...
               },
               {
                 "code_id": 6,
```
NeverHappened: Don't forget that this new contract needs to change msg as well

NeverHappened (changes requested):

NeverHappened (changes requested):
Also need changes to DefaultBondDenom:
See how its implemented in osmosis: https://github.com/osmosis-labs/osmosis/blob/main/app/params/config.go#L15

We have this as well:
x/feerefunder/types/params.go

Also you can search in project for `DefaultBondDenom` and change it to our new DefaultBondDenom.

Also see `DefaultNeutronDenom` constant in feeburner module - should be renamed.

app/params/denom.go
```diff
@@ -0,0 +1,4 @@
+package params
+
+// DefaultBondDenom defines neutron's denom
+const DefaultBondDenom = "untrn"
```
mmikev: Let's avoid `bond` word, cause we don't have staking features anymore

testutil/chain.go
```diff
@@ -0,0 +1,106 @@
...
+	channeltypes "github.com/cosmos/ibc-go/v3/modules/core/04-channel/types"
+	ibctesting "github.com/cosmos/ibc-go/v3/testing"
+	"github.com/cosmos/ibc-go/v3/testing/mock"
+	"github.com/neutron-org/neutron/app/params"
+	"github.com/stretchr/testify/require"
+	tmproto "github.com/tendermint/tendermint/proto/tendermint/types"
+	tmtypes "github.com/tendermint/tendermint/types"
+)
+
+func NewTestChainWithValSet(t *testing.T, coord *ibctesting.Coordinator, appIniter ibctesting.AppIniter, chainID string, valSet *tmtypes.ValidatorSet, signers map[string]tmtypes.PrivValidator) *ibctesting.TestChain {
```
sotnikov-s, line 20: please add a comment describing that this is the same func as [here](https://github.com/cosmos/ibc-go/blob/af9b461c63274b9ce5917beb89a2c92e865419df/testing/chain.go#L94) just with our native denom. it will imply the other mates that this piece of code is not to be reviewed thoroughly

testutil/chain.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	coord.CommitBlock(chain)
+
+	return chain
+}
+
+// NewTestChain initializes a new test chain with a default of 4 validators
+// Use this function if the tests do not need custom control over the validator set
+func NewTestChain(t *testing.T, coord *ibctesting.Coordinator, appIniter ibctesting.AppIniter, chainID string) *ibctesting.TestChain {
```
sotnikov-s, line 84: this func is exactly the same as [here](https://github.com/cosmos/ibc-go/blob/af9b461c63274b9ce5917beb89a2c92e865419df/testing/chain.go#L159). why copy+paste, not import and reuse?

sotnikov-s (approved):

foxpy (approved):

pr0n00gler (approved):

PR #128, opened 2023-01-18
init.sh refactoring

oldremez:
This PR is to make init.sh in a proper way


network/init.sh
```diff
@@ -80,59 +80,152 @@ $NEUTROND_BINARY add-genesis-account $($NEUTROND_BINARY --home $CHAIN_DIR/$CHAIN
...
-$NEUTROND_BINARY add-wasm-message store ${VOTING_REGISTRY_CONTRACT} --output json  --run-as ${ADMIN_ADDRESS} --home $CHAIN_DIR/$CHAINID_1
-$NEUTROND_BINARY add-wasm-message store ${PRE_PROPOSAL_CONTRACT} --output json  --run-as ${ADMIN_ADDRESS} --home $CHAIN_DIR/$CHAINID_1
-$NEUTROND_BINARY add-wasm-message store ${PROPOSAL_MULTIPLE_CONTRACT} --output json  --run-as ${ADMIN_ADDRESS} --home $CHAIN_DIR/$CHAINID_1
-$NEUTROND_BINARY add-wasm-message store ${PRE_PROPOSAL_MULTIPLE_CONTRACT} --output json  --run-as ${ADMIN_ADDRESS} --home $CHAIN_DIR/$CHAINID_1
-$NEUTROND_BINARY add-wasm-message store ${TREASURY_CONTRACT} --output json --run-as ${ADMIN_ADDRESS} --home $CHAIN_DIR/$CHAINID_1
-
-# Instantiate the contract
+
+function store_binary() {
+  $NEUTROND_BINARY add-wasm-message store $1 --output json --run-as ${ADMIN_ADDRESS} --keyring-backend=test --home $CHAIN_DIR/$CHAINID_1
```
albertandrejev: imho it is better to isolate function from global variables and use `local` variables only. Also please give name to the `$1` parameter.

network/init.sh
```diff
@@ -80,59 +80,152 @@ $NEUTROND_BINARY add-genesis-account $($NEUTROND_BINARY --home $CHAIN_DIR/$CHAIN
...
+  "name": "Neutron",
+  "initial_items": null,
+  "proposal_modules_instantiate_info": [
+    {
+      "code_id": '${PROPOSAL_CONTRACT_BINARY_ID}',
+      "label": "DAO_Neutron_cw-proposal-single",
+      "msg": "'"${PROPOSAL_SINGLE_INIT_MSG_BASE64}"'"
+    },
+    {
+      "code_id": '"${PROPOSAL_MULTIPLE_CONTRACT_BINARY_ID}"',
```
albertandrejev, line 204: Please remove double quotes here or add double quotes on line 196.

network/init.sh
oldremez: Does bash even [support named variables for functions](https://unix.stackexchange.com/a/427602) huh?
By using `local` variables do you mean passing everything as parameters?
Basically, I use to think about this function more as a macro rather than a function that's why I made it this way

network/init.sh
oldremez: Agreed, made it consistent

network/init.sh
oldremez: Oh I got what you're saying, you propose to create a local var and give it the value of the parameter. Will do

zavgorodnii (approved):

PR #129, opened 2023-01-19
feat: feePayer for Sudo fees

ratik:
related stuff: 

- [neutron-dev-contracts](https://github.com/neutron-org/neutron-dev-contracts/pull/6) 
- [neutron-sdk](https://github.com/neutron-org/neutron-sdk/pull/77) 
- [docs](https://github.com/neutron-org/neutron-docs/pull/45) 
- [integration-test](https://github.com/neutron-org/neutron-integration-tests/pull/70) 
- [test-run](https://github.com/neutron-org/neutron-tests/actions/runs/4014773994/)
- [Jira Task](https://p2pvalidator.atlassian.net/browse/NTRN-331)

1. A fee payer is an address that holds tokens that can be used to pay for the interchain transaction fees.
2. The fee payer can grant an allowance to a contract address, which allows the contract to use tokens from this address for the fees. Optionally, a limit, expire date, and period can be set, please refer to the [feegrant module's documentation in the Cosmos SDK](https://docs.cosmos.network/v0.46/modules/feegrant/) for more information.
3. When an interchain transaction is requested by a contract, the feerefunder module checks the allowance in general by using the feegrant module's GetAllowance function.
4. The feerefunder module then calls the Accept method on the returned interface with the total fees as an argument to check if the contract has permission to use the required amount of tokens and to deduct them from the allowance.
5. If the allowance is enough for the interchain transaction, the contract can execute the transaction.
6. The feegrant module is responsible for ensuring that the contract has enough tokens to pay for the fees, if it doesn't have enough tokens, the transaction should return an error message.
7. The payer field from the Fee struct is used to specify which address should be used to pay the fees.


proto/interchaintxs/v1/tx.proto
```diff
@@ -55,7 +55,7 @@ message MsgSubmitTx {
 
 // MsgSubmitTxResponse defines the response for Msg/SubmitTx
 message MsgSubmitTxResponse {
-  // channel's sequence_id for outgoing ibc packet. Unique per a channel.
+  // channel's sequence_id for outgoing ibc packet. Unique per a channel.y
```
mmikev: typo

NeverHappened: accident `y`?

x/feerefunder/keeper/keeper.go
```diff
@@ -34,37 +35,58 @@ func NewKeeper(
...
+			return err
+		}
+		if allowance != nil { // otherwise there is no allowance
+			coins := sdk.NewCoins()
+			coins = append(coins, fee.TimeoutFee...)
+			coins = append(coins, fee.AckFee...)
+			coins = append(coins, fee.RecvFee...)
+			_, err = allowance.Accept(ctx, coins, []sdk.Msg{})
+			if err != nil {
+				return err
```
NeverHappened: wrap an error

x/feerefunder/keeper/keeper.go
```diff
@@ -34,37 +35,58 @@ func NewKeeper(
...
+	payer := payerInfo.Sender
 
 	if err := k.checkFees(ctx, fee); err != nil {
 		return sdkerrors.Wrapf(err, "failed to lock fees")
 	}
 
+	if payerInfo.FeePayer != nil && !payerInfo.FeePayer.Empty() {
+		allowance, err := k.feegrantKeeper.GetAllowance(ctx, payerInfo.FeePayer, payerInfo.Sender)
+		if err != nil {
+			return err
```
NeverHappened: wrap an error

x/feerefunder/keeper/keeper.go
```diff
@@ -34,37 +35,58 @@ func NewKeeper(
...
+		}
```
mmikev, line 76: new line after error handling, please

x/feerefunder/keeper/keeper.go
mmikev: wrap error to make it more meaningful?

x/feerefunder/keeper/keeper_test.go
```diff
@@ -156,18 +162,80 @@ func TestKeeperLockFees(t *testing.T) {
...
+	validFee := types.Fee{
+		RecvFee:    nil,
+		AckFee:     ackFee,
+		TimeoutFee: timeoutFee,
+	}
+
+	channelKeeper.EXPECT().GetChannel(ctx, packet.PortId, packet.ChannelId).Return(channeltypes.Channel{}, true)
+	feegrantKeeper.EXPECT().GetAllowance(ctx, payerInfo.FeePayer, payerInfo.Sender).Return(nil, fmt.Errorf("feegrant error"))
+	err := k.LockFees(ctx, payerInfo, packet, validFee)
+	require.ErrorContains(t, err, "feegrant error")
```
NeverHappened, line 254: how about adding one more case for when allowance.Accept() returns error? It actually is pretty real, cause it's when there are no more coins left to grant error

x/interchaintxs/keeper/msg_server.go
```diff
@@ -83,6 +83,12 @@ func (k Keeper) SubmitTx(goCtx context.Context, msg *ictxtypes.MsgSubmitTx) (*ic
 		return nil, sdkerrors.Wrapf(ictxtypes.ErrNotContract, "%s is not a contract address", msg.FromAddress)
 	}
 
+	feePayerAddr, err := sdk.AccAddressFromBech32(msg.Fee.Payer)
+	if msg.Fee.Payer != "" && err != nil {
+		k.Logger(ctx).Debug("SubmitTx: failed to parse fee payer address", "fee payer", msg.Fee.Payer)
+		return nil, sdkerrors.Wrapf(sdkerrors.ErrInvalidAddress, "failed to parse address: %s", msg.Fee.Payer)
+	}
+
```
mmikev: Just a suggestion: maybe we can define a method which accepts `sender` and `payer` and returns `(feetypes.PayerInfo, error)`?

pr0n00gler (changes requested):

go.mod
```diff
@@ -7,6 +7,7 @@ require (
 	github.com/CosmWasm/wasmvm v1.1.1
 	github.com/confio/ics23/go v0.7.0
 	github.com/cosmos/admin-module v0.0.0-00010101000000-000000000000
+	github.com/cosmos/cosmos-proto v1.0.0-beta.1
```
NeverHappened, line 10: it's planned, right?)

NeverHappened (approved):
LGTM, mentioned small things

x/interchaintxs/keeper/msg_server.go
NeverHappened: +1, cause it looks a bit hard to read

go.mod
ratik: that's from ignite

ratik: also it came from target branch lol

x/feerefunder/keeper/keeper_test.go
```diff
@@ -156,18 +162,113 @@ func TestKeeperLockFees(t *testing.T) {
...
+	feeAllowance.EXPECT().Accept(ctx, fees, []sdk.Msg{}).Return(false, nil)
+	feegrantKeeper.EXPECT().GetAllowance(ctx, payerInfo.FeePayer, payerInfo.Sender).Return(feeAllowance, nil)
+	bankKeeper.EXPECT().SendCoinsFromAccountToModule(ctx, payerInfo.FeePayer, types.ModuleName, validFee.Total()).Return(nil)
+	err = k.LockFees(ctx, payerInfo, packet, validFee)
 	require.NoError(t, err)
 	require.Equal(t, sdk.Events{
 		sdk.NewEvent(
 			types.EventTypeLockFees,
-			sdk.NewAttribute(types.AttributeKeyPayer, payer.String()),
+			sdk.NewAttribute(types.AttributeKeyPayer, payerInfo.Sender.String()),
```
sotnikov-s: I don't understand why is it `payerInfo.Sender`, not `payerInfo.FeePayer`?

x/feerefunder/keeper/keeper.go
```diff
@@ -204,6 +227,23 @@ func (k Keeper) StoreFeeInfo(ctx sdk.Context, feeInfo types.FeeInfo) {
...
 }
 
+func (k Keeper) GetPayerInfo(ctx sdk.Context, sender string, payer string) (types.PayerInfo, error) {
+	senderAddr, err := sdk.AccAddressFromBech32(sender)
+	if err != nil {
+		k.Logger(ctx).Debug("Transfer: failed to parse sender address", "sender", sender)
+		return types.PayerInfo{}, sdkerrors.Wrapf(sdkerrors.ErrInvalidAddress, "failed to parse address: %s", sender)
+	}
+	feePayerAddr, err := sdk.AccAddressFromBech32(payer)
+	if payer != "" && err != nil {
```
sotnikov-s: why try to create an address from the payer if the payer is empty? this leads to confusing possible empty but not nil `FeePayer` in the `types.PayerInfo` below which is not good: in go, it's usually a prepared ready-to-use value given or nil.
please consider having the `FeePayer` equal to nil in case payer is not provided here. this will also improve [this part of code](https://github.com/neutron-org/neutron/blob/62a86f1b1943bce8140080fe512cf9a6454b0a02/x/feerefunder/keeper/keeper.go#L72): will be no need to check if its empty, only if its nil.

x/feerefunder/keeper/keeper_test.go
ratik: bc one is actually tx sender and the other is actually fee payer and both of them are kind of payment information. So mb you have any good idea for a struct name which consist of sender and fee payer?

x/feerefunder/keeper/keeper.go
ratik: good point!

x/feerefunder/keeper/keeper_test.go
sotnikov-s: no no naming is good. I meant why the `types.AttributeKeyPayer` attribute value here is `payerInfo.Sender` if we've passed a `payerInfo` containing the `FeePayer` field into the `LockFees`? we use the `payer` as the `types.AttributeKeyPayer` value which is set to `payerInfo.FeePayer` [here](https://github.com/neutron-org/neutron/blob/efff248f4d848ad1329c9e98aa0419de521c66a0/x/feerefunder/keeper/keeper.go#L87). how then it can be equal to `payerInfo.Sender` here, not `payerInfo.FeePayer`?

ratik: oh, got it. so the payer is Sender when we don't have FeePayer. 
image

ratik: thanks, nice spot 👍

x/feerefunder/keeper/keeper_test.go
```diff
@@ -156,18 +163,113 @@ func TestKeeperLockFees(t *testing.T) {
...
+	require.ErrorContains(t, err, "failed to parse address")
+
+	p, err := k.GetPayerInfo(ctx, TestAddress, "")
+	require.NoError(t, err)
+	require.Equal(t, p, types.PayerInfo{
+		Sender:   sdk.MustAccAddressFromBech32(TestAddress),
+		FeePayer: nil,
+	})
+
+	p, err = k.GetPayerInfo(ctx, TestAddress, TestAddress)
```
NeverHappened: To test logic to verify that it's different addresses it would probably be better to send different addresses as args

sotnikov-s (approved):

pr0n00gler (approved):

PR #130, opened 2023-01-20
chore: improve feeburner test cov #ntrn-328

NeverHappened:
improved coverage of feeburner.
to do that, needed to extract function from EndBlock
also changed to just panic if any error occurs during the process in EndBlock 

https://github.com/neutron-org/neutron-tests/actions/runs/4084447916


x/feeburner/keeper/keeper.go
```diff
@@ -71,6 +78,51 @@ func (k Keeper) GetTotalBurnedNeutronsAmount(ctx sdk.Context) types.TotalBurnedN
...
+	}
+
+	if len(fundsForTreasury) > 0 {
+		err := k.bankKeeper.SendCoins(
+			ctx,
+			moduleAddr, sdk.MustAccAddressFromBech32(params.TreasuryAddress),
+			fundsForTreasury,
+		)
+		if err != nil {
+			//TODO: check that works
```
mmikev: what works?

NeverHappened: not finished, doing that thing!) meant formatting of error

pr0n00gler (commented):
Target branch must be `neutron_audit_informal_17_01_2023`

testutil/mocks/interchaintxs/types/expected_keepers.go
```diff
@@ -116,15 +116,15 @@ func (m *MockContractManagerKeeper) EXPECT() *MockContractManagerKeeperMockRecor
 }
 
 // AddContractFailure mocks base method.
-func (m *MockContractManagerKeeper) AddContractFailure(ctx types.Context, channelId, address string, ackID uint64, ackType string) {
+func (m *MockContractManagerKeeper) AddContractFailure(ctx types.Context, channelID, address string, ackID uint64, ackType string) {
```
NeverHappened: that's just gomock output

x/feeburner/keeper/keeper.go
```diff
@@ -71,6 +79,50 @@ func (k Keeper) GetTotalBurnedNeutronsAmount(ctx sdk.Context) types.TotalBurnedN
...
+	}
+
+	if len(fundsForTreasury) > 0 {
+		err := k.bankKeeper.SendCoins(
+			ctx,
+			moduleAddr, sdk.MustAccAddressFromBech32(params.TreasuryAddress),
+			fundsForTreasury,
+		)
+		if err != nil {
+			return fmt.Errorf("error sending funds to treasury for address=%s, tokens=%+v: %v", params.TreasuryAddress, fundsForTreasury, err)
```
mmikev: What is the specific reason of returning an error here? We panic everywhere in this method except this particular line.

NeverHappened: That's a bit out of scope). But:
I thought the reason was that it's okay to skip sending to treasury in this block, because we'll do again in later blocks anyway.
But now I think your point is valid actually, and this is not an okay behaviour.

zavgorodnii (approved):

oldremez (approved):

PR #138, opened 2023-02-09
Chore: generalize init scripts #NTRN-364

oldremez:
The aim of this PR is to unite the current version of the init scripts from the Neutron repository and from the integration tests repository to avoid duplication. Basically, it's just a refactoring of those scripts with moving a lot of stuff to environment variables rather than hard coding. Also, the scripts were refactored a lot for the sake of escaping and stuff. The main significant breaking change is port reordering.

Linked PR: https://github.com/neutron-org/neutron-integration-tests/pull/80

To review this PR please:
1. Take a look at the overall approach
2. Check the tests in docker environment are passed here: https://github.com/neutron-org/neutron-tests/actions/runs/4363286672
3. Try to run the local tests in a non-docker environment. They're kinda partially broken but the following set works pretty well, doesn't require to re-load the environment, and is enough to check if everything is ok at this point. (!!! don't forget to fix ports in your ICQ-relayer config !!!)
```
NO_DOCKER=1 yarn test:interchain_tx_query_resubmit
NO_DOCKER=1 yarn test:lockdrop
NO_DOCKER=1 yarn test:simple
NO_DOCKER=1 yarn test:tokenomics
```

Since this PR is complex in terms of resolving conflicts, most style suggestions and so on will be rejected as out-of-scope, sorry. Still, while it's not gonna block the PR merging I'll be happy to implement all the suggestions later.


foxpy (commented):
Looks good, I didn't test it though

PR #139, opened 2023-02-13
feat: remove processed tx hashes on remove TX InterchainQuery

sotnikov-s:
**task:** https://p2pvalidator.atlassian.net/browse/NTRN-349

**this PR:**
- adds relative processed TX hashes removal on removal of a TX interchain query;
- makes the above-mentioned calls partially refunded: only interaction with the storage is refunded, more details about refunding below;
- prettifies ProcessBlock method.

**refund details:**
before refund is added:
- small tx query removal cost: 13892;
- large tx query removal cost (10,000 tx hashes): 10908412;
- kv query removal cost: 12661.

after refund is added:
- small tx query removal cost: 10649;
- large tx query removal cost (10,000 tx hashes): 10649;
- kv query removal cost: 10661.

**tests run:**
https://github.com/neutron-org/neutron-tests/actions/runs/4164639065

### DIFF
--- a/x/interchainqueries/keeper/keeper.go
+++ b/x/interchainqueries/keeper/keeper.go
@@ -132,9 +132,11 @@ func (k Keeper) GetAllRegisteredQueries(ctx sdk.Context) []*types.RegisteredQuer
 	return queries
 }
 
-func (k Keeper) RemoveQueryByID(ctx sdk.Context, id uint64) {
+// RemoveQuery removes the given query and all relative result data from the store.
+func (k Keeper) RemoveQuery(ctx sdk.Context, query *types.RegisteredQuery) {
 	store := ctx.KVStore(k.storeKey)
-	store.Delete(types.GetRegisteredQueryByIDKey(id))
+	store.Delete(types.GetRegisteredQueryByIDKey(query.Id))
+	k.removeQueryResult(ctx, query)
 }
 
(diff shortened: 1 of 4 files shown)


x/interchainqueries/keeper/keeper.go
```diff
@@ -195,9 +197,23 @@ func (k Keeper) GetQueryResultByID(ctx sdk.Context, id uint64) (*types.QueryResu
...
+	switch {
+	case queryType.IsKV():
+		store.Delete(types.GetRegisteredQueryResultByIDKey(query.Id))
+	case queryType.IsTX():
+		prefixStore := prefix.NewStore(ctx.KVStore(k.storeKey), types.GetSubmittedTransactionIDForQueryKeyPrefix(query.Id))
+		iterator := prefixStore.Iterator(nil, nil)
+		defer iterator.Close()
+		for ; iterator.Valid(); iterator.Next() {
+			txHashKey := iterator.Key()
+			prefixStore.Delete(txHashKey)
```
swelf19: Every delete operation costs 1000 gas, are you sure we do not meet out of gas error during iteration process?

sotnikov-s: good catch. I also suspect that some queries with lots of recorded transactions are going to be unprofitable for the remover even given the deposit, so this will result in garbage queries that nobody wants to remove

PR #140, opened 2023-02-17
feat: optimize SubmitQueryResult NTRN-353

sotnikov-s:
**task:** https://p2pvalidator.atlassian.net/browse/NTRN-353

- reduce number of interactions with Store on `SubmitQueryResult`;
- add early check of KV query result's remote height for compliance.

**tests:** https://github.com/neutron-org/neutron-tests/actions/runs/4204510406

here's the result of SubmitQueryResult calls for neutron interchainqueries keeper tests without and with the changes of this PR:

| test name | gas consumption original | gas consumption with the improvement |
------------|--------------------------|----------------------------------------|
| invalid query id | 1027 | 1027 |
| valid KV storage proof | 31866 | **21860** |
| invalid number of KvResults | 1495 | 1495 |
| invalid query type | 1363 | 1363 |
| nil proof | 5980 | 5980 |
| non-registered key in KV result | 5980 | 5980 |
| non-registered path in KV result | 5980 | 5980 |
| non existence KV proof | 24726 | **15440** |
| header with invalid height | 3987 | 3987 |
| invalid KV storage value | 5980 | 5980 |
| query result height is too old | 16303 | **1504** |
| non existence KV proof with special bytes in key | 27201 | **17015** |

### DIFF
--- a/wasmbinding/test/custom_query_test.go
+++ b/wasmbinding/test/custom_query_test.go
@@ -39,7 +39,7 @@ func (suite *CustomQuerierTestSuite) TestInterchainQueryResult() {
 	clientKey := host.FullClientStateKey(suite.Path.EndpointB.ClientID)
 	lastID := neutron.InterchainQueriesKeeper.GetLastRegisteredQueryKey(ctx) + 1
 	neutron.InterchainQueriesKeeper.SetLastRegisteredQueryKey(ctx, lastID)
-	registeredQuery := icqtypes.RegisteredQuery{
+	registeredQuery := &icqtypes.RegisteredQuery{
 		Id: lastID,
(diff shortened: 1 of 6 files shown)


foxpy (commented):
The change set looks good and I would like to approve it, though, at first I think it would be great to measure gas consumption for query result submission before and after the patch.

sotnikov-s:
> The change set looks good and I would like to approve it, though, at first I think it would be great to measure gas consumption for query result submission before and after the patch.

good idea. I've attached the results table to the PR description

foxpy (approved):

pr0n00gler (approved):

PR #149, opened 2023-03-01
NTRN-389 audit modules fixes

quasisamurai:
[Test run](https://github.com/neutron-org/neutron-tests/actions/runs/4341210957/jobs/7580624365)

[TASK](https://p2pvalidator.atlassian.net/browse/NTRN-389)

### DIFF
--- a/x/interchainqueries/keeper/keeper.go
+++ b/x/interchainqueries/keeper/keeper.go
@@ -256,6 +256,7 @@ func clearQueryResult(result *types.QueryResult) types.QueryResult {
 		KvResults: storageValues,
 		Block:     nil,
 		Height:    result.Height,
+		Revision:  result.Revision,
 	}
 
 	return cleanResult
--- a/x/interchainqueries/keeper/keeper.go
+++ b/x/interchainqueries/keeper/keeper.go
@@ -285,8 +285,7 @@ func (k Keeper) GetClientState(ctx sdk.Context, clientID string) (*tendermintLig
 func (k *Keeper) CollectDeposit(ctx sdk.Context, queryInfo types.RegisteredQuery) error {
 	owner, err := queryInfo.GetOwnerAddress()
 	if err != nil {
-		ctx.Logger().Debug("failed to get owner address from given query", "query_info", queryInfo)
-		return err
+		panic(err.Error())
 	}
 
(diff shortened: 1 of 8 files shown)


x/interchainqueries/types/message_remove_interchain_query.go
```diff
@@ -7,24 +7,22 @@ import (
 	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
 )
 
-var _ sdk.Msg = &MsgRemoveInterchainQueryRequest{}
```
mmikev, line 10: Why is this removed?
It's pretty necessary compile check that `MsgRemoveInterchainQueryRequest` really implements sdk.Msg

x/interchainqueries/keeper/keeper.go
```diff
@@ -284,7 +285,8 @@ func (k Keeper) GetClientState(ctx sdk.Context, clientID string) (*tendermintLig
 func (k *Keeper) CollectDeposit(ctx sdk.Context, queryInfo types.RegisteredQuery) error {
 	owner, err := queryInfo.GetOwnerAddress()
 	if err != nil {
-		panic(err.Error())
+		ctx.Logger().Debug("failed to get owner address from given query", "query_info", queryInfo)
```
albertandrejev: Do you know scenario where Owner will not be valid Address? I think that if we will simply write message to log it may lead to data inconsistency, problem hiding, and this will start growing. Network halting is not the best solution of course, but at least in this case we will see this weird situation if we will catch unexpected error like this.

x/interchaintxs/types/tx.go
```diff
@@ -10,8 +10,6 @@ import (
 	"github.com/gogo/protobuf/proto"
 )
 
-var _ codectypes.UnpackInterfacesMessage = MsgSubmitTx{}
```
albertandrejev: Same question as @pr0n00gler

x/interchainqueries/types/message_remove_interchain_query.go
quasisamurai: that happened accidentally, tnx

zavgorodnii (dismissed):

zavgorodnii (approved):

albertandrejev (approved):

PR #152, opened 2023-03-07
NTRN-387 add more wasmbindings

quasisamurai:
[TASK](https://p2pvalidator.atlassian.net/browse/NTRN-387)

[TEST RUN](https://github.com/neutron-org/neutron-tests/actions/runs/4448950096)

Here is implementation of the bindings for the rest types of proposals we're gonna support


app/proposals_whitelisting.go
```diff
@@ -19,21 +19,14 @@ func IsConsumerProposalWhitelisted(content govtypes.Content) bool {
 	switch c := content.(type) {
 	case *proposal.ParameterChangeProposal:
 		return isConsumerParamChangeWhitelisted(c.Changes)
-	case *govtypes.TextProposal,
-		*upgradetypes.SoftwareUpgradeProposal,
+	case *upgradetypes.SoftwareUpgradeProposal,
```
zavgorodnii, line 22: i have a question regarding the remaining proposal types. here we register some of the proposal types while setting up the `wasmd` module: https://github.com/neutron-org/neutron/blob/56875ea0bd6765ad4504778a1f0808164504b11e/app/app.go#L465 do we need to register the remaining proposal types in the same place?

quasisamurai: This one was updated in PR. I just moved `wasmd`-related `Route` initialization right after creating `wasmKeeper` init

quasisamurai:
screenshot

to be fair, I actually don't know why[this](https://github.com/neutron-org/neutron/actions/runs/4456768754) appears. it seems not reproducible locally

oldremez:
pls update the test run

wasmbinding/message_plugin.go
```diff
@@ -318,7 +320,94 @@ func (m *CustomMessenger) performSubmitAdminProposal(ctx sdk.Context, contractAd
 		}
 	}
 
-	if proposal.ParamChangeProposal == nil && proposal.SoftwareUpgradeProposal == nil && proposal.CancelSoftwareUpgradeProposal == nil {
+	if proposal.UpgradeProposal != nil {
+		p := proposal.UpgradeProposal
+		err := msg.SetContent(&ibcclienttypes.UpgradeProposal{
```
oldremez, line 329: What if there will be several admin proposals in one message? Only a single one will be executed? Regardless of the answer, let's maybe write the test to cover this case?

wasmbinding/message_plugin.go
```diff
@@ -318,7 +320,94 @@ func (m *CustomMessenger) performSubmitAdminProposal(ctx sdk.Context, contractAd
...
+			Title:       p.Title,
+			Description: p.Description,
+			Contract:    p.Contract,
+		})
+		if err != nil {
+			return nil, sdkerrors.Wrap(err, "failed to set content on ClearAdminProposal")
+		}
+	}
+
+	if proposal.ParamChangeProposal == nil &&
```
oldremez: Wouldn't it be better to check `msg` content here? Like, if at least one of these isn't `nil` then content should be filled, isn't it?

foxpy (dismissed):

wasmbinding/message_plugin.go
oldremez: The fix lgtm but I would add test anyway

quasisamurai:
[FRESH RUN](https://github.com/neutron-org/neutron-tests/actions/runs/4508959637/jobs/7939905038)

oldremez (approved):

foxpy (approved):

PR #156, opened 2023-03-09
feat: remove tx hashes in EndBlock NTRN-378

sotnikov-s:
**task:** https://p2pvalidator.atlassian.net/browse/NTRN-378

Benchmarked `TxQueriesCleanup` duration for different values of `TxQueryRemovalLimit` for case of single query to be removed having `TxQueryRemovalLimit` tx hashes (avg of 10 runs for each case):
| TxQueryRemovalLimit | avg TxQueriesCleanup duration |
|-|-|
| 1_000 | 0.001s |
| 10_000 | 0.013s |
| 100_000 | 0.166s |
| 1_000_000 | 2.186s |

Benchmarked `TxQueriesCleanup` duration for different values of `TxQueryRemovalLimit` for case of `TxQueryRemovalLimit` count of queries to be removed each having a single tx hash (avg of 10 runs for each case):
| TxQueryRemovalLimit | avg TxQueriesCleanup duration |
|-|-|
| 1_000 | 0.027s |
| 10_000 | 0.093s |
| 100_000 | 3.891s |
| 1_000_000 | more than I could wait |

Assuming validators' computational resources are at least as good as mine (m1 macbook pro 2021) and based on the results, I suggest assigning TxQueryRemovalLimit 5k or 10k to maintant ICQ's EndBlock succinct and efficient.

**relative PRs:**
- https://github.com/neutron-org/neutron-integration-tests/pull/96

**tests:** https://github.com/neutron-org/neutron-tests/actions/runs/4392779384


proto/interchainqueries/params.proto
```diff
@@ -18,4 +18,8 @@ message Params {
     (gogoproto.castrepeated) = "github.com/cosmos/cosmos-sdk/types.Coins",
     (gogoproto.nullable) = false
   ];
+
+  // Amount of tx hashes to be removed during a single EndBlock. Can vary to balance
+  // between network cleaning speed and EndBlock duration. A zero value means no limit.
+  uint64 tx_query_removal_limit = 3;
```
mmikev, line 24: Why is it 3 by default?
You said 5k is good enough?

x/interchainqueries/keeper/keeper.go
```diff
@@ -132,9 +132,51 @@ func (k Keeper) GetAllRegisteredQueries(ctx sdk.Context) []*types.RegisteredQuer
...
+}
+
+// TxQueriesCleanup cleans the module store from obsolete registered TX queries and relative
+// stored transaction hashes. Cleans up to params.TxQueryRemovalLimit hashes at a time or all
+// the hashes if params.TxQueryRemovalLimit is 0.
+func (k Keeper) TxQueriesCleanup(ctx sdk.Context) {
+	rmLimit := k.GetParams(ctx).TxQueryRemovalLimit
+	limited := rmLimit != 0
+
+	queriesToRm := make([]*TxQueryToRemove, 0, rmLimit/10)
```
mmikev, line 159: Could you explain why do you divide `rmLimit` by 10 here?

x/interchainqueries/keeper/keeper.go
```diff
@@ -334,3 +370,55 @@ func (k Keeper) MustPayOutDeposit(ctx sdk.Context, deposit sdk.Coins, sender sdk
...
+	if len(ids) == 0 {
+		return nil
+	}
+	return ids
+}
+
+// getTxQueryHashes retrieves up to limit tx hashes related to a query with the given ID from the
+// Keeper's store and a bool meaning if all the query's hashes have been retrieved. If limit is 0,
+// it retrieves all the hashes for the given query.
+func (k Keeper) getTxQueryHashes(ctx sdk.Context, queryID uint64, limit uint64) (txHashes [][]byte, complete bool) {
```
mmikev: Pretty weird method for me, actually.
Why not to return `TxQueryToRemove` structure itself?

proto/interchainqueries/params.proto
sotnikov-s: 3 is the number of the field in the proto msg

x/interchainqueries/keeper/keeper.go
sotnikov-s: just to preallocate some memory. using the whole `rmLimit` would be too much imo because I believe the case when we have `rmLimit` number of queries to be removed is not plausible. on the other hand, leaving it without any preallocation is not efficient

x/interchainqueries/keeper/keeper.go
sotnikov-s: good idea! will look into it

proto/interchainqueries/params.proto
mmikev: Oh lol, i need to get some sleep

x/interchainqueries/keeper/keeper.go
sotnikov-s: done. now it's much more clear, ty

pr0n00gler (commented):
I think we need some metrics/logs about tx hashes removing rate to set `TxQueryRemovalLimit` properly:
* how long does it take to remove tx hashes;
* how many tx hashes left, how many were deleted (maybe our limit is not enough, and our queue grows faster than we can delete per block);
* etc

That's what came up in my mind first, maybe you have better ideas about this

sotnikov-s:
> how long does it take to remove tx hashes;

that's a good idea, will add

> how many tx hashes left, how many were deleted (maybe our limit is not enough, and our queue grows faster than we can delete per block);

this is not trivial because we can't easily calculate how many tx hashes there are. but we can add such a handler to be used in monitoring. I'd create a task for it because it's out of scope and there is a lot of more urgent stuff

pr0n00gler:
> this is not trivial because we can't easily calculate how many tx hashes there are. but we can add such a handler to be used in monitoring. I'd create a task for it because it's out of scope and there is a lot of more urgent stuff

Sure, but we definitely need at least some basic metrics/logs here. Otherwise it's literally a black box right now and we know nothing about what's happening here (in terms of performance).

sotnikov-s:
@pr0n00gler, I've added logging:

```json
{
   "level":"debug",
   "module":"x/interchainqueries",
   "duration_ms":0,
   "hashes_removed":0,
   "queries_removed":0,
   "time":"2023-03-13T15:50:23+03:00",
   "message":"TxQueriesCleanup performed"
}
```

pr0n00gler:
> @pr0n00gler, I've added logging:
> 
> ```json
> {
>    "level":"debug",
>    "module":"x/interchainqueries",
>    "duration_ms":0,
>    "hashes_removed":0,
>    "queries_removed":0,
>    "time":"2023-03-13T15:50:23+03:00",
>    "message":"TxQueriesCleanup performed"
> }
> ```



> @pr0n00gler, I've added logging:
> 
> ```json
> {
>    "level":"debug",
>    "module":"x/interchainqueries",
>    "duration_ms":0,
>    "hashes_removed":0,
>    "queries_removed":0,
>    "time":"2023-03-13T15:50:23+03:00",
>    "message":"TxQueriesCleanup performed"
> }
> ```

👍

sotnikov-s:
@pr0n00gler I don't need 👍, I need ✅

pr0n00gler (approved):

x/interchainqueries/keeper/keeper.go
```diff
@@ -334,3 +379,55 @@ func (k Keeper) MustPayOutDeposit(ctx sdk.Context, deposit sdk.Coins, sender sdk
...
+			return result
+		}
+	}
+	result.CompleteRemoval = true
+	return result
+}
+
+// TxQueryToRemove contains data related to a single query listed for removal and needed in the
+// removal process.
+type TxQueryToRemove struct {
```
albertandrejev, line 424: maybe you can move this struct to the top of the file, next to `Keeper struct` ?

sotnikov-s: I don't really like having all the structs at the top of the file because of two reasons:

1. different structs have different degrees of importance to the reader. say, `Keeper` is the main structure in the `keeper.go` file and having it at the top is crucial because the reader will most likely look for this struct after opening this file. on the other hand, `TxQueryToRemove` is mostly an auxiliary struct;

2. I like keeping such a structure across files (roughly):
- consts
- struct A constructor and def
- struct A methods
- struct B constructor and def
- struct B methods
- ...
- auxiliary functions
in this way structure definition integrity is kept and one can read the whole structure without any breaks and distractions to other stuff.

hope it makes sense to you

x/interchainqueries/keeper/keeper.go
```diff
@@ -132,9 +133,59 @@ func (k Keeper) GetAllRegisteredQueries(ctx sdk.Context) []*types.RegisteredQuer
...
+		queryToRm := k.calculateTxQueryRemoval(ctx, queryID, rmLimit)
+		queriesToRm = append(queriesToRm, queryToRm)
+
+		if limited {
+			rmLimit -= uint64(len(queryToRm.Hashes))
+			if rmLimit <= 0 {
+				break
+			}
+		}
+	}
```
albertandrejev, line 170: Why don't you want to use nested loop here? It will reduce memory/resource consumption because of lesser amount of iterations.
Something like
```go
for TxQueriesToRemove.Iterator {
   queryToRm := k.calculateTxQueryRemoval(ctx, queryID, rmLimit)
   queriesToRm = append(queriesToRm, queryToRm)
   ....
}
```

sotnikov-s: I tried it but it didn't work: there are iterators over the same resource inside both `calculateTxQueryRemoval` and `GetTxQueriesToRemove`, and the iterator is tricky designed so one of them got blocked even without write operations. i.e. when I iterated inside another iterator loop I got deadlocked. idk why but this is why I've chosen such an approach

albertandrejev (approved):

NeverHappened (approved):

PR #157, opened 2023-03-09
fix: limit InterchainAccountID length #NTRN-385

foxpy:
Tests run: https://github.com/neutron-org/neutron-tests/actions/runs/4383230435

Merge with https://github.com/neutron-org/neutron-docs/pull/51

### DIFF
--- a/x/interchaintxs/types/tx.go
+++ b/x/interchaintxs/types/tx.go
@@ -10,7 +10,7 @@ import (
 	"github.com/gogo/protobuf/proto"
 )
 
-const interchainAccountIdLimit = 128 -
+const interchainAccountIDLimit = 128 -
 	len("icacontroller-") -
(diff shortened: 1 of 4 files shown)


pr0n00gler (approved):

x/interchaintxs/types/tx.go
```diff
@@ -10,6 +10,12 @@ import (
 	"github.com/gogo/protobuf/proto"
 )
 
+// const interchainAccountIDLimit = 47
+const interchainAccountIDLimit = 128 -
+	len("icacontroller-") -
+	len("neutron1unyuj8qnmygvzuex3dwmg9yzt9alhvyeat0uu0jedg2wj33efl5qmysp02") - // just a random contract address
```
zavgorodnii, line 16: this is weird af, but it looks like there is no length constant for string representation of contract addresses

zavgorodnii (approved):

PR #161, opened 2023-03-11
Add overrule preproposal to genesis + address generation #NTRN-303

oldremez:
This PR does the following:
1. Adds overrule-pre-proposal module to genesys
2. Initializes the DAO so that it could handle overrule proposals
3. (last but not the least) makes all addresses in init script generated instead of hardcoded

Since it changes contracts' addresses, the following PR in integration tests is needed to test this one: https://github.com/neutron-org/neutron-integration-tests/pull/95

Tests run https://github.com/neutron-org/neutron-tests/actions/runs/4393949558


sotnikov-s (approved):
I like it, but the only thing that concerns me in the init script is that we still have a restriction for those instance_ids. e.g. if we'll ever have to add another contract instantiation here, we'll have to care about the `1, 2, 3...12` numbers in the `generate-contract-address` params so the new contract is placed in the right place here and below in the `instantiate-contract` section. could you please add a comment somewhere there mentioning it?

oldremez:
@sotnikov-s I refactored the generation section so we wouldn't need to deal with 1,2,3… numbers and added enhanced comments about the order of lines importance

zavgorodnii (approved):

PR #163, opened 2023-03-14
fix: non audited fees drain #ntrn-400

NeverHappened:
Task: https://p2pvalidator.atlassian.net/browse/NTRN-400

Fixes condition where we can lock fees with many denoms not in params, that in turn will cause gas usage problems later


x/feerefunder/keeper/keeper.go
```diff
@@ -237,3 +245,14 @@ func (k Keeper) distributeFee(ctx sdk.Context, receiver sdk.AccAddress, fee sdk.
 	}
 	return nil
 }
+
+// someFeesAreNotPresentInParams returns true if one or more coins from `fees` are not present in coins from `params`
+// assumes that `params` is sorted
+func someFeesAreNotPresentInParams(fees sdk.Coins, params sdk.Coins) bool {
```
zavgorodnii: I would probably change the param name to `allowedCoins`

zavgorodnii (approved):

pr0n00gler (approved):

PR #177, opened 2023-03-22
feat: cron #ntrn-413

NeverHappened:
adds a cron module
can add (dao only) or remove (dao or security only) schedules to execute arbitrary cosmwasm contract msgs at given periods

implements this task: https://p2pvalidator.atlassian.net/browse/NTRN-413

Test run: https://github.com/neutron-org/neutron-tests/actions/runs/4572781964


x/cron/keeper/keeper.go
```diff
@@ -0,0 +1,205 @@
...
+// AddSchedule adds new schedule to execution for every block `period`.
+// First schedule execution is supposed to be on `now + period` block.
+func (k *Keeper) AddSchedule(ctx sdk.Context, name string, period uint64, msgs []types.MsgExecuteContract) {
+	schedule := types.Schedule{
+		Name:              name,
+		Period:            period,
+		Msgs:              msgs,
+		LastExecuteHeight: uint64(ctx.BlockHeight()), // let's execute newly added schedule on `now + period` block
+	}
+	k.storeSchedule(ctx, schedule)
```
swelf19, line 93: Dont you think we need to check that the schedule with the name already exists?

x/cron/keeper/keeper.go
```diff
@@ -0,0 +1,205 @@
...
+	for ; iterator.Valid(); iterator.Next() {
+		var schedule types.Schedule
+		k.cdc.MustUnmarshal(iterator.Value(), &schedule)
+
+		if k.intervalPassed(ctx, schedule) {
+			res = append(res, schedule)
+			count++
+
+			if count >= params.Limit {
+				return res
```
swelf19, line 164: Should we warn about the limit reached. There is a possible corner case - 10 schedules with 3 block interval each. The last schedule will never be executed and its to easier to know about with some warnings.

x/cron/keeper/grpc_query_schedule.go
```diff
@@ -0,0 +1,56 @@
...
+
+	"github.com/cosmos/cosmos-sdk/store/prefix"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/cosmos-sdk/types/query"
+	"github.com/neutron-org/neutron/x/cron/types"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+func (k Keeper) ScheduleAll(c context.Context, req *types.QueryAllScheduleRequest) (*types.QueryAllScheduleResponse, error) {
```
swelf19: Is it ok to name method `*All` when you use pagination?

x/cron/keeper/grpc_query_schedule.go
```diff
@@ -0,0 +1,56 @@
...
+
+	var schedules []types.Schedule
+	ctx := sdk.UnwrapSDKContext(c)
+
+	store := ctx.KVStore(k.storeKey)
+	scheduleStore := prefix.NewStore(store, types.ScheduleKey)
+
+	pageRes, err := query.Paginate(scheduleStore, req.Pagination, func(key []byte, value []byte) error {
+		var schedule types.Schedule
+		if err := k.cdc.Unmarshal(value, &schedule); err != nil {
```
swelf19: There are a lot of places where you use MustUnmarshal/MustMarshal. I think we need to make a choose

x/cron/keeper/keeper.go
```diff
@@ -0,0 +1,205 @@
...
+	k.storeSchedule(ctx, schedule)
+}
+
+// RemoveSchedule removes schedule with a given `name`
+func (k *Keeper) RemoveSchedule(ctx sdk.Context, name string) {
+	k.removeSchedule(ctx, name)
+}
+
+// GetAllSchedules returns all schedules
+func (k *Keeper) GetAllSchedules(ctx sdk.Context) []types.Schedule {
```
swelf19, line 123: dead code?

x/cron/keeper/grpc_query_schedule.go
NeverHappened: I think this is code generated

x/cron/keeper/grpc_query_schedule.go
NeverHappened: This was code generated. But I can fix

x/cron/keeper/keeper.go
NeverHappened: Yeah, I can add a check for that

x/cron/keeper/grpc_query_schedule.go
NeverHappened: Renamed query to Schedules

proto/cron/params.proto
```diff
@@ -0,0 +1,17 @@
...
+
+import "gogoproto/gogo.proto";
+
+option go_package = "github.com/neutron-org/neutron/x/cron/types";
+
+// Params defines the parameters for the module.
+message Params {
+  option (gogoproto.goproto_stringer) = false;
+  // Admin address that can add/remove schedules
+  string admin_address = 1;
```
zavgorodnii: wait! why do you  need to keep this as part of the cron module params? can't the admin address be retrieved from the admin module state?

proto/cron/schedule.proto
```diff
@@ -0,0 +1,20 @@
+syntax = "proto3";
+package neutron.cron;
+
+option go_package = "github.com/neutron-org/neutron/x/cron/types";
+
+import "gogoproto/gogo.proto";
+
+message Schedule {
+  string name = 1;
+  uint64 period = 2;
```
zavgorodnii: please add a comment explaining what period this is (blocks? seconds?)

proto/cron/tx.proto
```diff
@@ -0,0 +1,14 @@
+syntax = "proto3";
+package neutron.cron;
+
+// this line is used by starport scaffolding # proto/tx/import
+
+option go_package = "github.com/neutron-org/neutron/x/cron/types";
+
+// Msg defines the Msg service.
+service Msg {
+// this line is used by starport scaffolding # proto/tx/rpc
```
zavgorodnii, line 10: You sure everything's right here?

x/cron/client/cli/query.go
```diff
@@ -0,0 +1,33 @@
+package cli
+
+import (
+	"fmt"
+	// "strings"
+
+	"github.com/spf13/cobra"
+
+	"github.com/cosmos/cosmos-sdk/client"
+	// "github.com/cosmos/cosmos-sdk/client/flags"
```
zavgorodnii: commented code

x/cron/client/cli/tx.go
```diff
@@ -0,0 +1,26 @@
+package cli
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+
+	"github.com/cosmos/cosmos-sdk/client"
+	// "github.com/cosmos/cosmos-sdk/client/flags"
```
zavgorodnii: commented code

x/cron/keeper/keeper.go
```diff
@@ -191,6 +180,12 @@ func (k *Keeper) removeSchedule(ctx sdk.Context, name string) {
 	store.Delete(types.GetScheduleKey(name))
 }
 
+func (k *Keeper) scheduleExists(ctx sdk.Context, name string) bool {
+	store := ctx.KVStore(k.storeKey)
+	bz := store.Get(types.GetScheduleKey(name))
```
swelf19: Why do you prefer `Get` over `Has` there

x/cron/keeper/keeper.go
```diff
@@ -0,0 +1,218 @@
...
+			}
+		}
+	}
+
+	return res
+}
+
+// executeSchedule executes given schedule and changes LastExecuteHeight
+func (k *Keeper) executeSchedule(ctx sdk.Context, schedule types.Schedule) {
+	for idx, msg := range schedule.Msgs {
```
zavgorodnii, line 182: nope, this implementation does not work, unfortunately. if we allow for execution of multiple messages, we need to guarantee transactionality. please use a cached context to execute messages from a schedule one by one, and only update the state if all of the messages were executed successfully. in case of any questions, talk to @pr0n00gler 

also please implement tests that will check this particular aspect of the implementation

proto/cron/tx.proto
NeverHappened: Well we do not allow for direct messages for Cron module, that's why it's empty. Or WDYM?

x/cron/keeper/keeper.go
```diff
@@ -0,0 +1,264 @@
...
+		paramstore:    ps,
+		accountKeeper: accountKeeper,
+	}
+}
+
+func (k *Keeper) Logger(ctx sdk.Context) log.Logger {
+	return ctx.Logger().With("module", fmt.Sprintf("x/%s", types.ModuleName))
+}
+
+// ExecuteReadySchedules gets all schedules that are due for execution (with limit that is equals to Params.Limit)
```
zavgorodnii: is equal

x/cron/keeper/keeper.go
```diff
@@ -0,0 +1,264 @@
...
+	}
+
+	return res
+}
+
+func (k *Keeper) GetScheduleCount(ctx sdk.Context) int32 {
+	return k.getScheduleCount(ctx)
+}
+
+func (k *Keeper) StoreSchedule(ctx sdk.Context, schedule types.Schedule) {
```
zavgorodnii: not sure what's happening here... you have `AddSchedule()`, `storeSchedule()` and `StoreSchedule()`, and the last one is only used in genesis. looks a bit weird?

x/cron/keeper/keeper.go
```diff
@@ -0,0 +1,264 @@
...
+	store.Delete(types.GetScheduleKey(name))
+}
+
+func (k *Keeper) scheduleExists(ctx sdk.Context, name string) bool {
+	store := prefix.NewStore(ctx.KVStore(k.storeKey), types.ScheduleKey)
+	return store.Has(types.GetScheduleKey(name))
+}
+
+func (k *Keeper) intervalPassed(ctx sdk.Context, schedule types.Schedule) bool {
+	return uint64(ctx.BlockHeight()) > (schedule.LastExecuteHeight + schedule.Period)
```
zavgorodnii, line 226: wdyt about doing something fancy like `(ctx.BlockHeight() % schedule.Period) == 0`

x/cron/keeper/keeper.go
```diff
@@ -0,0 +1,264 @@
...
+
+// executeSchedule executes all msgs in a given schedule and changes LastExecuteHeight
+// if at least one msg execution fails, rollback all messages
+func (k *Keeper) executeSchedule(ctx sdk.Context, schedule types.Schedule) error {
+	// Even if contract execution returned an error, we still increase the height
+	// and execute it after this interval
+	schedule.LastExecuteHeight = uint64(ctx.BlockHeight())
+	k.storeSchedule(ctx, schedule)
+
+	subCtx, commit := ctx.CacheContext()
```
zavgorodnii: in the sdk these variables are usually named `cacheCtx, writeFn`

x/cron/keeper/keeper.go
swelf19: looks fancy, but in case a task queue size more than the "limit". Lets say the limit is 3, but we have 4 tasks with period 3, then we constantly drop a 4th task

x/cron/keeper/keeper.go
NeverHappened: I used `StoreSchedule` to store schedules with specified `LastExecuteHeight`, otherwise they get stored with current block height which can be not easy to manage, especially in tests.

`storeSchedule` is private

x/cron/keeper/keeper.go
NeverHappened: This can happen, but I meant this limit for situation where too many schedules are ready on one block.
So situation where there are too many schedules overall can happen, but it should not be feasible, since it's government decisions that add schedules

NeverHappened: Looked better into it, and agree with @swelf19, current solution better for me since it's moves the unhandled load on the next block, your solution moves the load onto next period, so if the period is overloaded, it will happen forever

x/cron/keeper/keeper.go
NeverHappened: Ok, so I removed the StoreSchedule, and now it's much clearer!

x/cron/keeper/keeper.go
oopcode: @swelf19 what?.. what does queue size have to do with checking whether enough blocks have passed?

zavgorodnii (dismissed):

swelf19 (dismissed):

NeverHappened:
https://github.com/neutron-org/neutron-tests/actions/runs/4598701363

pr0n00gler (dismissed):

NeverHappened:
https://github.com/neutron-org/neutron-tests/actions/runs/4603665947

pr0n00gler (approved):

swelf19 (approved):

NeverHappened:
https://github.com/neutron-org/neutron-tests/actions/runs/4605338708

PR #178, opened 2023-03-23
Feat: packet forward middleware

pr0n00gler:
**Merge with:** https://github.com/neutron-org/neutron-integration-tests/pull/112
**Tests run:** https://github.com/neutron-org/neutron-tests/actions/runs/4545383271


swelf19 (commented):
What are the middleware sideeffects? does it charge some fees? What do you think to implement some unit tests?

proto/transfer/v1/tx.proto
```diff
@@ -41,7 +41,9 @@ message MsgTransfer {
   uint64 timeout_timestamp = 7
       [ (gogoproto.moretags) = "yaml:\"timeout_timestamp\"" ];
 
-  neutron.feerefunder.Fee fee = 8
+  string memo = 8;
```
zavgorodnii, line 44: usually reordering of fields is not a good practice, is there any specific reason to do it here?

zavgorodnii:
please add a documentation pr for the new module!

zavgorodnii:
in the integration tests there is only a success scenario; probably we would also like to check that error propagation works properly. please implement n error scenario, or create a task to implement one

proto/transfer/v1/tx.proto
mmikev: Yes, we want to save compatibility with the original transfer message from ibc-go

zavgorodnii (approved):

swelf19 (approved):

PR #190, opened 2023-04-03
Fix: unblock fee-pool addr to be able to send rewards to a provider chain

pr0n00gler:
**Tests:** https://github.com/neutron-org/neutron-tests/actions/runs/4595545396

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -368,7 +368,7 @@ func New(
 		ccvconsumertypes.ConsumerToSendToProviderName).String())
 
 	app.BankKeeper = bankkeeper.NewBaseKeeper(
-		appCodec, keys[banktypes.StoreKey], app.AccountKeeper, app.GetSubspace(banktypes.ModuleName), app.ModuleAccountAddrs(),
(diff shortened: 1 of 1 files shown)


NeverHappened (dismissed):

ratik (dismissed):

foxpy (approved):

albertandrejev (approved):

app/app.go
```diff
@@ -796,6 +796,19 @@ func (app *App) ModuleAccountAddrs() map[string]bool {
 	return modAccAddrs
 }
 
+// BlockedAddrs returns the set of addresses that are not allowed
+// to send and receive funds
```
swelf19, line 800: `to send` NewBaseKeeper comment does not say anything like that, are you sure `send` is forbidden?

swelf19 (approved):

app/app.go
mmikev: Yes, it's forbidden in the ibc transfer module. That's why we can't send rewards to a provider chain

PR #203, opened 2023-04-11
fix: update `InstantiateMsg` for lockdrop vault in init-neutrond.sh

foxpy:


foxpy:
lockdrop vault has been deleted from genesis

PR #205, opened 2023-04-12
fix: unconditional distribute fee

zavgorodnii:
This PR makes sure that the Timeout fee and the Acknowledgement fee get paid to the relayer even if we recover from out of gas panic during execution of the contract callback.

Please not that I couldn't figure out a way to properly write the integration tests for this scenario: if you simply try to write a lot of data to the contract store to run out of gas, we don't get a panic, we get an out of gas error from wasmvm that is propagated to `HandleAcknowledgement` / `HandleTimeout`.


zavgorodnii:
https://github.com/neutron-org/neutron-tests/actions/runs/4688913462

pr0n00gler (approved):

PR #214, opened 2023-04-18
fix: applied the CWA-2023-002 patch

zavgorodnii:
https://github.com/CosmWasm/advisories/blob/main/CWAs/CWA-2023-002.md


zavgorodnii:
https://github.com/neutron-org/neutron-tests/actions/runs/4735927118

sotnikov-s (approved):

PR #216, opened 2023-04-20
feat: NTRN-430. Add commission for Token Factory

albertandrejev:
We need to collect commission for coin creation using Token Factory, collected commission should go to the treasury.


x/feeburner/types/params.go
```diff
@@ -27,24 +27,24 @@ func ParamKeyTable() paramtypes.KeyTable {
...
 	)
 }
 
 // NewParams creates a new Params instance
 func NewParams(neutronDenom, reserveAddress string) Params {
 	return Params{
-		NeutronDenom:   neutronDenom,
-		ReserveAddress: reserveAddress,
+		NeutronDenom:    neutronDenom,
+		TreasuryAddress: reserveAddress,
```
zavgorodnii: incorrect variable name

x/tokenfactory/keeper/createdenom.go
```diff
@@ -64,3 +69,32 @@ func (k Keeper) validateCreateDenom(ctx sdk.Context, creatorAddr string, subdeno
 
 	return denom, nil
 }
+
+func (k Keeper) chargeForCreateDenom(ctx sdk.Context, creatorAddr string, subdenom string) (err error) {
+	// Send creation fee to community pool
+	creationFee := k.GetParams(ctx).DenomCreationFee
+	accAddr, err := sdk.AccAddressFromBech32(creatorAddr)
+	if err != nil {
+		return err
```
zavgorodnii: in the whole function -- please wrap errors

zavgorodnii (changes requested):

proto/feeburner/params.proto
```diff
@@ -19,5 +19,5 @@ message Params {
   // of neutron mainnet will include a proper name (reserve_address).
   // If we still want to upgrade the network from the old version to the new one,
   // we need to implement the handler to migrate treasury_address to reserve_address.
-  string reserve_address = 2;
+  string treasury_address = 2;
```
sotnikov-s: description of the field is outdated

x/tokenfactory/keeper/createdenom.go
```diff
@@ -64,3 +70,31 @@ func (k Keeper) validateCreateDenom(ctx sdk.Context, creatorAddr string, subdeno
...
+			return sdkerrors.Wrapf(types.ErrUnableToCharge, "wrong fee collector address: %v", err)
+		}
+
+		err = k.bankKeeper.SendCoins(
+			ctx,
+			accAddr, feeCollectorAddr,
+			creationFee,
+		)
+
+		return sdkerrors.Wrap(err, "unable to send coins to fee collector")
```
sotnikov-s: I believe you missed the `if err != nil` check here. it works because of the following hidden behaviour:

```go
// If err is nil, this returns nil, avoiding the need for an if statement when
// wrapping a error returned at the end of a function
func Wrap(err error, description string) error {
```

I don't think we should leave it this way cause it's so much obscure and against the most common go practice ever: checking `if err != nil` before err return. I understand that this is a bit more concise, but imo it rudely disrupts the main golang dogma. wyt?

same above at the `createDenomAfterValidation` call

x/tokenfactory/keeper/createdenom_test.go
```diff
@@ -13,6 +13,9 @@ func (suite *KeeperTestSuite) TestMsgCreateDenom() {
 	suite.Setup()
 
 	// Creating a denom should work
+	senderAddress := suite.ChainA.SenderAccounts[0].SenderAccount.GetAddress()
+	suite.TopUpWallet(suite.ChainA.GetContext(), senderAddress, suite.TestAccs[0])
```
sotnikov-s, line 21: shouldn't we also make sure that the creation fee is charged and received by the fee collector?

x/tokenfactory/keeper/keeper_test.go
```diff
@@ -38,32 +38,45 @@ func (suite *KeeperTestSuite) Setup() {
...
-	fmt.Printf("Setup: %s\n", suite.TestAccs)
 
 	suite.SetupTokenFactory()
 
 	suite.queryClient = types.NewQueryClient(suite.QueryHelper)
 
 	tokeFactoryKeeper := suite.GetNeutronZoneApp(suite.ChainA).TokenFactoryKeeper
+	tokeFactoryKeeper.SetParams(suite.ChainA.GetContext(), types.NewParams(
+		sdktypes.NewCoins(sdktypes.NewInt64Coin(types.DefaultNeutronDenom, 1_000_000)),
+		"neutron1vguuxez2h5ekltfj9gjd62fs5k4rl2zy5hfrncasykzw08rezpfsd2rhm7",
```
sotnikov-s: please move the addr to a const and describe it. I can see it's used in different places across the code

x/tokenfactory/types/errors.go
```diff
@@ -19,4 +19,7 @@ var (
 	ErrSubdenomTooLong          = sdkerrors.Register(ModuleName, 8, fmt.Sprintf("subdenom too long, max length is %d bytes", MaxSubdenomLength))
 	ErrCreatorTooLong           = sdkerrors.Register(ModuleName, 9, fmt.Sprintf("creator too long, max length is %d bytes", MaxCreatorLength))
 	ErrDenomDoesNotExist        = sdkerrors.Register(ModuleName, 10, "denom does not exist")
+	ErrUnableToCharge           = sdkerrors.Register(ModuleName, 11, "unable to charge for denom creation")
+	ErrDenomCreation            = sdkerrors.Register(ModuleName, 12, "unable to create denom")
```
sotnikov-s: this err is unused. as far as I can see it's not needed

x/tokenfactory/types/params.go
```diff
@@ -1,30 +1,87 @@
...
+	return nil
+}
+
+func validateFeeCollectorAddress(i interface{}) error {
+	v, ok := i.(string)
+	if !ok {
+		return fmt.Errorf("invalid parameter type: %T", i)
+	}
+
+	// Reserve might be explicitly empty in test environments
```
sotnikov-s: **Reserve** might be -> **Treasury** might be? also, I think we should remove all the reserve/treasury mentions in this PR: it's not treasury or reserve, it's fee collector

zavgorodnii:
i think that we need to set other genesis module parameters in the init script, right?

albertandrejev:
> i think that we need to set other genesis module parameters in the init script, right?

No, we can use defaults, but we need to implement upgrade procedure

sotnikov-s (dismissed):
LGTM. looking forward to seeing documentation and integration tests

proto/osmosis/tokenfactory/v1beta1/params.proto
```diff
@@ -4,6 +4,16 @@ package osmosis.tokenfactory.v1beta1;
 import "gogoproto/gogo.proto";
 
 option go_package = "github.com/neutron-org/neutron/x/tokenfactory/types";
+import "cosmos_proto/cosmos.proto";
+import "cosmos/base/v1beta1/coin.proto";
 
 // Params holds parameters for the tokenfactory module
-message Params {}
+message Params {
+    repeated cosmos.base.v1beta1.Coin denom_creation_fee = 1 [
```
NeverHappened, line 13: add comment?

x/tokenfactory/keeper/createdenom.go
```diff
@@ -64,3 +74,33 @@ func (k Keeper) validateCreateDenom(ctx sdk.Context, creatorAddr string, subdeno
 
 	return denom, nil
 }
+
+func (k Keeper) chargeForCreateDenom(ctx sdk.Context, creatorAddr string) (err error) {
```
NeverHappened: since  this function is private and exists only in `createdenom` file, how about renaming function to `chargeFee`?  (or is it osmosis copypaste)

wasmbinding/test/custom_message_test.go
```diff
@@ -147,6 +156,12 @@ func (suite *CustomMessengerTestSuite) TestCreateDenomMsg() {
 	suite.contractAddress = suite.InstantiateReflectContract(suite.ctx, suite.contractOwner, codeID)
 	suite.Require().NotEmpty(suite.contractAddress)
 
+	senderAddress := suite.ChainA.SenderAccounts[0].SenderAccount.GetAddress()
```
NeverHappened, line 159: also maybe add test somewhere that if no coins on a balance, CreateDenom fails?

NeverHappened:
also noticed in `genesis.go`:
		`fmt.Println("set auth meta", genDenom.GetDenom(), genDenom.GetAuthorityMetadata(), err)`
Probably use logger or get rid of

wasmbinding/test/custom_message_test.go
albertandrejev: Implemented but in another test https://github.com/neutron-org/neutron/blob/943fb820d141b758b77ab4a6661e8ec62957bc09/x/tokenfactory/keeper/createdenom_test.go#L17

NeverHappened (dismissed):

NeverHappened (dismissed):

NeverHappened (dismissed):
LGTM!
But:
I did not run it, hope that you checked that migration work as expected!

albertandrejev:
> LGTM! But: I did not run it, hope that you checked that migration work as expected!

Yes, I checked it!!!

app/upgrades/v0.4.4/upgrades.go
```diff
@@ -0,0 +1,55 @@
...
+		}
+
+		ctx.Logger().Info("Migrating SlashingKeeper Params...")
+		oldSlashingParams := keepers.SlashingKeeper.GetParams(ctx)
+		oldSlashingParams.SignedBlocksWindow = int64(36000)
+
+		keepers.SlashingKeeper.SetParams(ctx, oldSlashingParams)
+
+		ctx.Logger().Info("Migrating FeeBurner Params...")
+		s, ok := keepers.ParamsKeeper.GetSubspace(feeburnertypes.ModuleName)
```
sotnikov-s, line 35: why not
```go
oldFeeBurnerParams := keepers.FeeBurnerKeeper.GetParams(ctx)
```
and use its fields for `feeburnerDefaultParams.TreasuryAddress` and `feeburnerDefaultParams.NeutronDenom`?

app/upgrades/v0.4.4/upgrades.go
```diff
@@ -0,0 +1,55 @@
...
+		if !ok {
+			panic("global fee burner params subspace not found")
```
sotnikov-s: why panic, not return err?

proto/feeburner/params.proto
```diff
@@ -9,15 +9,10 @@ option go_package = "github.com/neutron-org/neutron/x/feeburner/types";
...
   string neutron_denom = 1;
-  // Defines reserve address
-  // TODO: it is important to remember that it used to be treasury_address here.
-  // A name conflict can lead to an error if the network is upgraded from
-  // a previous version to the current one (for example, on a testnet).
-  // This issue does not affect the mainnet, as the first version
-  // of neutron mainnet will include a proper name (reserve_address).
-  // If we still want to upgrade the network from the old version to the new one,
-  // we need to implement the handler to migrate treasury_address to reserve_address.
+  // Deprecated in v0.4.4. Is not used anymore
```
sotnikov-s, line 14: do we need to keep it?

x/tokenfactory/types/params.go
```diff
@@ -1,30 +1,88 @@
...
+	return nil
+}
+
+func validateFeeCollectorAddress(i interface{}) error {
+	v, ok := i.(string)
+	if !ok {
+		return fmt.Errorf("invalid parameter type: %T", i)
+	}
+
+	// Fee collector address might be explicitly empty in test environments
```
sotnikov-s, line 77: do we need to allow it? we don't know here whether it's a test environment or a prod one. so maybe we force having a valid address here and use some in the test environment?

x/tokenfactory/types/params.go
```diff
@@ -1,30 +1,88 @@
...
+	return paramtypes.ParamSetPairs{
+		paramtypes.NewParamSetPair(KeyDenomCreationFee, &p.DenomCreationFee, validateDenomCreationFee),
+		paramtypes.NewParamSetPair(KeyFeeCollectorAddress, &p.FeeCollectorAddress, validateFeeCollectorAddress),
+	}
+}
+
+func validateDenomCreationFee(i interface{}) error {
+	v, ok := i.(sdk.Coins)
+	if !ok {
+		return fmt.Errorf("invalid parameter type: %T", i)
```
sotnikov-s, line 61: would be nice to have the error like
```go
return fmt.Errorf("invalid denom creation fee type: expected sdk.Coin, got %T", i)
```

x/tokenfactory/types/params.go
```diff
@@ -1,30 +1,88 @@
...
+		return fmt.Errorf("invalid denom creation fee: %+v", i)
+	}
+
+	return nil
+}
+
+func validateFeeCollectorAddress(i interface{}) error {
+	v, ok := i.(string)
+	if !ok {
+		return fmt.Errorf("invalid parameter type: %T", i)
```
sotnikov-s, line 74: would be nice to have the error like
```go
return fmt.Errorf("invalid fee collector address type: expected string, got %T", i)
```

x/tokenfactory/types/params.go
```diff
@@ -1,30 +1,88 @@
...
+}
+
+func validateDenomCreationFee(i interface{}) error {
+	v, ok := i.(sdk.Coins)
+	if !ok {
+		return fmt.Errorf("invalid parameter type: %T", i)
+	}
+
+	if v.Validate() != nil {
+		return fmt.Errorf("invalid denom creation fee: %+v", i)
```
sotnikov-s: I'd add the validation error to the result error here

app/upgrades/v0.4.4/upgrades.go
zavgorodnii: because if you do it like that, it will panic during unmarshaling (you won't have the required new field in the old state)

app/upgrades/v0.4.4/upgrades.go
zavgorodnii: if we return an error here, the u[grade module will panic anyways. i'll fix it

proto/feeburner/params.proto
zavgorodnii: yes

x/tokenfactory/types/params.go
zavgorodnii: i am not sure i like it, but i don't think it's a big issue

x/tokenfactory/types/params.go
zavgorodnii: we treat it here same as in other modules. moving to technical debt

PR #225, opened 2023-04-28
Make disabling of CCV msg handler configurable

oldremez:
This PR allows configuring if arbitrary messages should be disabled while the chain is not interchain-secured.
The option "disable-ccv-handler" in `app.toml` being set to true disables message filtering for the test environment.

Passing tests show that disabling works properly (in the other case tests wouldn't pass).

To test if messages are filtered if the option is "false" or not even present in the config do the following:
1. Comment the last line of `init-neutrond.sh` (`sed -i -e '1s/^/disable-ccv-handler...`)
2. Run the chain via `make start`
3. Run the following command:
`neutrond tx bank send demowallet1 neutron1suhgf5svhu4usrurvxzlgn54ksxmn8gljarjtxqnapv8kjnp4nrstdxvff 1untrn --fees 125000untrn --keyring-backend test --home data/test-1 --chain-id test-1 -y -b block`
It should fail with `raw_log: tx contains unsupported message types at height` in the output.

### DIFF
--- a/app/ante_handler.go
+++ b/app/ante_handler.go
@@ -23,7 +23,7 @@ type HandlerOptions struct {
 	TXCounterStoreKey sdk.StoreKey
 }
 
-func NewAnteHandler(options HandlerOptions) (sdk.AnteHandler, error) {
+func NewAnteHandler(options HandlerOptions, disableCcvHandler bool) (sdk.AnteHandler, error) {
 	if options.AccountKeeper == nil {
 		return nil, sdkerrors.Wrap(sdkerrors.ErrLogic, "account keeper is required for AnteHandler")
 	}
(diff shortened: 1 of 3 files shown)


oldremez:
https://github.com/neutron-org/neutron-tests/actions/runs/4832851937/jobs/8612252984

foxpy (commented):
Looks legit, but I didn't test it :+1:

albertandrejev:
Please add new section to the config generation template. 
You may find this code useful https://github.com/cosmos/cosmos-sdk/blob/27c6be521a733cf657cd8d23e10e25b5bdae2d14/simapp/simd/cmd/root.go#L170
And also this one https://github.com/cosmos/cosmos-sdk/blob/27c6be521a733cf657cd8d23e10e25b5bdae2d14/server/config/toml.go

zavgorodnii (dismissed):

oldremez:
Closing this one since we done that on build time here https://github.com/neutron-org/neutron/pull/227

PR #227, opened 2023-04-30
ccv msg filter build tag

oldremez:
This PR allows configuring if arbitrary messages should be disabled while the chain is not interchain-secured at built time.
The build tag "skip_ccv_msg_filter" being set to true disables message filtering for the test environment.

Passing tests show that disabling works properly (in the other case tests wouldn't pass).

To test if messages are filtered if the option is "false" or not even present in the config do the following:
1. Comment the line 53 of `Makefile` (`build_tags_test_binary += skip_ccv_msg_filter`)
2. Run the chain via `make start`
3. Run the following command:
`neutrond tx bank send demowallet1 neutron1suhgf5svhu4usrurvxzlgn54ksxmn8gljarjtxqnapv8kjnp4nrstdxvff 1untrn --fees 125000untrn --keyring-backend test --home data/test-1 --chain-id test-1 -y -b block`
It should fail with `raw_log: tx contains unsupported message types at height` in the output.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@ FROM golang:1.20-bullseye
 RUN apt-get update && apt-get install -y jq
 EXPOSE 26656 26657 1317 9090
 COPY --from=app . /opt/neutron
-RUN cd /opt/neutron && make install
+RUN cd /opt/neutron && make install-test-binary
 WORKDIR /opt/neutron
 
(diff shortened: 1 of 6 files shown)


oldremez:
https://github.com/neutron-org/neutron-tests/actions/runs/4844906101

zavgorodnii (approved):

Issue #176, opened 2023-03-21
PR's from external contributors will always show failing ci


faddat:
Two jobs are causing this:

* the one that pings team members on telegram when a PR is made
* gobencher

PR #235, opened 2023-05-03
Use the latest ibc-go and interchain-security modules - Notional & CryptoCrew

faddat:
This PR updates ibc-go to v4.4.0 and interchain-security to the latest commit from the main branch.  

The goal is to comply with our mutually adopted new chain launch standards (to be updated "soonTM" with consumer chain launch standards.) 

Since state-breaking upgrades are much more challenging than genesis, we recommend that genesis include these changes, after conferring with Informal systems about weather or not 0a000ff can be a proper release tag, so that Neutron's go.mod file will be somewhat cleaner than it is in this PR.

* #176


faddat:
Should most likely be merged before launch, after speaking with Informal Systems

PR #239, opened 2023-05-03
remove starport stubs

faddat:
- remove starport stubs from Neutron
- format code

### DIFF
--- a/app/simulation_test.go
+++ b/app/simulation_test.go
@@ -34,8 +34,6 @@ type SimApp interface {
 }
 
 // BenchmarkSimulation run the chain simulation
-// Running using starport command:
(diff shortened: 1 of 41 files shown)


faddat:
This is mainly a cosmetic PR, and can safely be reviewed and merged post launch.

foxpy (changes requested):
Honestly this sounds pretty dangerous to merge for me. Are we 100% sure in future we are not going to use any of the tools which might depend on these commented lines?

faddat:
I would really hope that we don't.  Super bigly ser.

quasisamurai:
It looks like this PR completely breaks `ignite`
Снимок экрана 2023-05-24 в 11 31 07

faddat:
it totally should break ignite. 

That's the point....

PR #240, opened 2023-05-03
build own protobufs

faddat:
This PR gives neutron the ability to build its own .proto files instead of relying on Starport, which is badly outdated and has been known to result in stability issues caused by chain teams coding around it.


faddat:
To my surprise, this PR did not result in changes to the compiled .pb.go files, so this is also safe to review/merge post-launch.

swelf19:
Nice work.
> To my surprise, this PR did not result in changes to the compiled .pb.go files, so this is also safe to review/merge post-launch.

actually it does, looks like there is a typo in line
`cp -r github.com/neutron-org/neutron/x x/` instead of replacing  existing pg.go files, the command creates a new dir ./x/x
`cp -r github.com/neutron-org/neutron/x/* x/` seems a correct one
i made necessary updates in the pr https://github.com/neutron-org/neutron/pull/262

PR #252, opened 2023-05-19
add a state sync script and add syncing to readme.md

faddat:
This PR adds statesync script and adds state sync to the readme file so that users can easily join the neutron network.

### DIFF
--- a/contrib/statesync.bash
+++ b/contrib/statesync.bash
@@ -1,0 +1,73 @@
+#!/bin/bash
+# microtick and bitcanna contributed significantly here.
+# Pebbledb state sync script.
+# invoke like: bash contrib/statesync.bash
+
+## USAGE RUNDOWN
+# Not for use on live nodes
+# For use when testing.
+
+set -uxe
+
+# Set Golang environment variables.
+# ! Adapt as required, depending on your system configuration
+#export GOPATH=~/go
+#export PATH=$PATH:~/go/bin
+
+# Install with pebbledb (uncomment for incredible performance)
+go mod edit -replace github.com/tendermint/tm-db=github.com/baabeetaa/tm-db@pebble
+go mod tidy
+
+# go install -ldflags '-w -s -X github.com/cosmos/cosmos-sdk/types.DBBackend=pebbledb -X github.com/tendermint/tm-db.ForceSync=1' -tags pebbledb ./...
+
+# install (comment if using pebble for incredible performance)
+# go install ./...
+
+
+go install -ldflags '-w -s -X github.com/cosmos/cosmos-sdk/types.DBBackend=pebbledb -X github.com/tendermint/tm-db.ForceSync=1' -tags pebbledb ./...
+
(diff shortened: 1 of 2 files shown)


dragonfrond (approved):

foxpy (commented):
This looks cool actually

contrib/statesync.bash
```diff
@@ -0,0 +1,73 @@
...
+set -uxe
+
+# Set Golang environment variables.
+# ! Adapt as required, depending on your system configuration
+#export GOPATH=~/go
+#export PATH=$PATH:~/go/bin
+
+# Install with pebbledb (uncomment for incredible performance)
+go mod edit -replace github.com/tendermint/tm-db=github.com/baabeetaa/tm-db@pebble
+go mod tidy
```
albertandrejev, line 19: I am not sure that we need to modify sources in this script even in this way. It is good to use on local installation, but there is risk that this changes will fall into sources. And it makes neutrond client incompatible with available snapshots. This requires additional discussion and decision.

contrib/statesync.bash
```diff
@@ -0,0 +1,73 @@
...
+
+# go install -ldflags '-w -s -X github.com/cosmos/cosmos-sdk/types.DBBackend=pebbledb -X github.com/tendermint/tm-db.ForceSync=1' -tags pebbledb ./...
+
+# install (comment if using pebble for incredible performance)
+# go install ./...
+
+
+go install -ldflags '-w -s -X github.com/cosmos/cosmos-sdk/types.DBBackend=pebbledb -X github.com/tendermint/tm-db.ForceSync=1' -tags pebbledb ./...
```
albertandrejev, line 27: same here

contrib/statesync.bash
```diff
@@ -0,0 +1,73 @@
...
+export NEUTROND_NODE=tcp://127.0.0.1:8011
+export NEUTROND_P2P_MAX_NUM_INBOUND_PEERS=500
+export NEUTROND_RPC_PPROF_LADDR=127.0.0.1:6969
+
+# Fetch and set list of seeds from chain registry.
+NEUTROND_P2P_SEEDS=$(curl -s https://raw.githubusercontent.com/cosmos/chain-registry/master/neutron/chain.json | jq -r '[foreach .peers.seeds[] as $item (""; "\($item.id)@\($item.address)")] | join(",")')
+export NEUTROND_P2P_SEEDS
+
+# Start chain.
+neutrond start --x-crisis-skip-assert-invariants --iavl-disable-fastnode false
```
albertandrejev, line 73: I suspect that you forgot to put here something to modify `config.toml` using above variables.

contrib/statesync.bash
faddat: we don't, technically

contrib/statesync.bash
faddat: we don't need to modify config.toml

contrib/statesync.bash
faddat: make sense

PR #261, opened 2023-06-05
fix: `make start`: GRPC_PORT setting didn't work

foxpy:
The problem is that GRPC_PORT setting affects execution of `network/start.sh`, but, without using `export`, environment variables did only affect `network/init.sh`. This PR fixes this oversight.


ratik (approved):

sotnikov-s (approved):
utACK

quasisamurai (approved):

Issue #253, opened 2023-05-19
State sync


faddat:
![image](https://github.com/neutron-org/neutron/assets/7142025/e9967b8d-8f43-45d9-bfd6-11fb319dc2b0)

This is the state sync process.  All of the available endpoints are getting hammered too much to respond to a state sync request I think.

PR #265, opened 2023-06-07
Only statesync

faddat:
Adds a state sync script and directions for using it.  Closed the previous one because github was acting up.


foxpy (commented):
Thank you for your contribution! I am looking forward to merge this PR, but there is one issue about it:
- each time I try to run this script, node successfully syncs to a snapshot, but it doesn't receive any new blocks, so it gets stuck at snapshot height. is it by design or am I doing something wrong?

foxpy:
Oh, I've just noticed #253 

indeed, there is something wrong about peers/public RPCs, they are probably overloaded?

foxpy (approved):
Nevertheless, I think the script you have contributed is good enough

pr0n00gler (approved):

PR #267, opened 2023-06-14
fix epoch downtime

faddat:
The solution that Osmosis came up with for downtime at their daily epoch (caused by iterators) was IAVL fast node. 

While it isn't as good as what is going into SDK .50, it should be at least a 2x improvement on the iterators, and 
may provide more of an improvement. 

- fix the downtime at slash time


faddat:
context:

back in the day osmo epochs took 20min or even an hour.  This fixed that.

cmd/neutrond/root.go
```diff
@@ -273,3 +276,59 @@ func (ac appCreator) appExport(
...
+	// validators here.
+	//
+	// In summary:
+	// - if you leave srvCfg.MinGasPrices = "", all validators MUST tweak their
+	//   own app.toml config,
+	// - if you set srvCfg.MinGasPrices non-empty, validators CAN tweak their
+	//   own app.toml to override, or use this default value.
+	//
+	// In simapp, we set the min gas prices to 0.
+	srvCfg.BaseConfig.IAVLDisableFastNode = false
```
faddat, line 315: this line is the fix; validators may need to state sync to kick it into place.

network/init-neutrond.sh
```diff
@@ -620,4 +620,8 @@ if ! jq -e . "$GENESIS_PATH" >/dev/null 2>&1; then
     exit 1
 fi
 
+<<<<<<< HEAD
 echo "DAO $DAO_CONTRACT_ADDRESS"
+=======
+echo "DAO $DAO_CONTRACT_ADDRESS"
+>>>>>>> v1.0.2
```
foxpy, line 627: This sure looks wrong :cat:

testutil/test_helpers.go
```diff
@@ -26,7 +26,8 @@ import (
 
 	clienttypes "github.com/cosmos/ibc-go/v4/modules/core/02-client/types"
 	appProvider "github.com/cosmos/interchain-security/app/provider"
-	e2e "github.com/cosmos/interchain-security/testutil/integration"
+	e2e "github.com/cosmos/interchain-security/testutil/e2e"
```
foxpy, line 29:
```suggestion
	"github.com/cosmos/interchain-security/testutil/e2e"
```

pr0n00gler:
Resolved in v2.0.0 upgrade

PR #270, opened 2023-06-18
use gci

faddat:
This PR reduces the correct import ordering to exactly one and makes import ordering in Neutron fully deterministic.


foxpy:
Good job! I see that linter (apparently? maybe I am missing something?) doesn't support reordering by itself, did you actually correct each and every linter warning by hand? I am asking because I wanted to run linter myself and compare diff with your branch.

foxpy:
@faddat now we also have merge conflicts in your PR.

PR #271, opened 2023-06-19
feat: allowlist ccvconsumer params to allow change from government #ntrn-531

NeverHappened:
Adds all ics module params to be allowed to change through governance.

https://p2pvalidator.atlassian.net/browse/NTRN-531


NeverHappened:
https://github.com/neutron-org/neutron-tests/actions/runs/5321485346 (todo: check if successful)

foxpy (approved):
Sadly we cannot add integration tests for these parameters for now, hence I think I can approve it right away?

NeverHappened:
> Sadly we cannot add integration tests for these parameters for now, hence I think I can approve it right away?

why? what about these tests I added? https://github.com/neutron-org/neutron-integration-tests/pull/169/files
Maybe I don't know something

PR #294, opened 2023-07-26
feat: set commit timeout #ntrn-79

NeverHappened:
https://hadronlabs.atlassian.net/browse/NTRN-79

checked only by printing in server config in cmd, it's hard to discover cometbft settings in runtime

PR #296, opened 2023-07-31
Update Default Timeout Commit to 1s

bpiv400:
This PR replaces the use of `server.InterceptConfigsPreRunHandler` in the root command with a similar function (now defined locally in `cmd/neutrond`) that sets the default `timeout_commit` value to `1s`, which is the value that the Neutron team is encouraging all validators to use. 

Without this, `neutrond init` initializes a config that contains `timeout_commit=5s`. The combination of validators using the standard `timeout_commit` and the value recommended by the Neutron team is undermining the network's ability to have single round consensus on new blocks consistently.


go.mod
```diff
@@ -9,7 +9,7 @@ require (
 	github.com/confio/ics23/go v0.9.0
 	github.com/cosmos/admin-module v0.0.0-00010101000000-000000000000
 	github.com/cosmos/cosmos-proto v1.0.0-beta.3
-	github.com/cosmos/cosmos-sdk v0.45.15
+	github.com/cosmos/cosmos-sdk v0.46.11
```
mmikev, line 12: We can't do such an upgrade, please leave it as it is. We are going to upgrade directly to SDK 47 soon

pr0n00gler (changes requested):

pr0n00gler:
Thx, but since we will upgrade directly to SDK 47 timeout commit issue was resolved in https://github.com/neutron-org/neutron/pull/294

PR #302, opened 2023-08-12
fix: Add googleapis as dep in buf.yaml

apollo-sturdy:
Without this `buf generate` fails as it cannot find `import "google/api/annotations.proto"` in `proto/transfer/v1/query.proto`.


pr0n00gler (approved):

foxpy (approved):

foxpy:
@pr0n00gler let's merge it?

foxpy:
@apollo-sturdy thank you!

PR #303, opened 2023-08-15
Rework governance/ update admin-module NTRN-70

quasisamurai:
[TASK](https://hadronlabs.atlassian.net/browse/NTRN-70)

[Admin module PR](https://github.com/neutron-org/admin-module/pull/2)
[Neutron PR](https://github.com/neutron-org/neutron/pull/303)
[Neutron SDK PR](https://github.com/neutron-org/neutron-sdk/pull/107)
[Integration tests](https://github.com/neutron-org/neutron-integration-tests/pull/196)


app/app.go
```diff
@@ -688,17 +691,18 @@ func New(
...
 	app.CronKeeper.WasmMsgServer = wasmkeeper.NewMsgServerImpl(&app.WasmKeeper)
 	cronModule := cron.NewAppModule(appCodec, &app.CronKeeper)
 
-	if len(enabledProposals) != 0 {
-		app.AdminmoduleKeeper.Router().AddRoute(wasm.RouterKey, wasm.NewWasmProposalHandler(app.WasmKeeper, enabledProposals))
-	}
+	// TODO: enabled proposals?
+	//if len(enabledProposals) != 0 {
+	//	app.AdminmoduleKeeper.Router().AddRoute(wasm.RouterKey, wasm.NewWasmProposalHandler(app.WasmKeeper, enabledProposals))
+	//}
```
mmikev: TODO?

app/app.go
```diff
@@ -629,18 +629,21 @@ func New(
...
 	// register the proposal types
-	adminRouter := govv1beta1.NewRouter()
-	adminRouter.AddRoute(govtypes.RouterKey, govv1beta1.ProposalHandler).
+	adminRouterLegacy := govv1beta1.NewRouter()
+	adminRouterLegacy.AddRoute(govtypes.RouterKey, govv1beta1.ProposalHandler).
 		AddRoute(paramproposal.RouterKey, params.NewParamChangeProposalHandler(app.ParamsKeeper)).
 		AddRoute(upgradetypes.RouterKey, upgrade.NewSoftwareUpgradeProposalHandler(&app.UpgradeKeeper)).
 		AddRoute(ibchost.RouterKey, ibcclient.NewClientProposalHandler(app.IBCKeeper.ClientKeeper))
 
+	keeperModules := map[string]adminmodulemodulekeeper.RegisteredModuleUpdateParams{wasm.ModuleName: {UpdateParamsMsg: &wasmtypes.MsgUpdateParams{}}}
```
mmikev: There should be a comment what exactly this line does (basically we register here modules which params can be changed via admin-module)

mmikev: Also i guess there shouldn't be a wasm module only. But also all the modules we have on Neutron, except ICS

wasmbinding/message_plugin.go
```diff
@@ -399,60 +401,102 @@ func (m *CustomMessenger) performSubmitAdminProposal(ctx sdk.Context, contractAd
...
-			return nil, errors.Wrap(err, "failed to set content on ClearAdminProposal")
+		sdkMsg = &wasmtypes.MsgClearAdmin{
+			Sender:   authority,
+			Contract: p.Contract,
+		}
+	}
+
+	if proposal.ParamChangeNewProposal != nil {
+		p := proposal.ParamChangeNewProposal
+		myJsonString := fmt.Sprintf(`{"authority":"%s", "params": "%s"}`, authority, p.NewParams)
```
mmikev: I don't like using raw json string here.
I'm suggesting the following:
* make a field `UpdateParamsMsg` with type `sdk.Msg` in `ParamChangeNewProposal` structure;
* Here we must compare it with `m.AdminKeeper.RegisteredModulesUpdateParams[p.Module].UpdateParamsMsg` to be sure, that the proposed message is a message to update params for a particulare module;
* we must check that `UpdateParamsMsg.Signer()` equals to admin module address (because the module is registered as an authority for other modules);
* if everything is ok, pass the message to the admin-module;

So in that case to create a proposal for changing params in specific module, proposer must compose a proper `UpdateParams` msg for the module (where `authority` is an address of the `admin-module`), create a proposal in Neutron DAO with that message and after it passes, it is being executed and passed to admin-module.

@swelf19 wdyt?

pr0n00gler (changes requested):

app/app.go
```diff
@@ -688,17 +690,14 @@ func New(
...
-		authtypes.NewModuleAddress(govtypes.ModuleName).String(),
+		authtypes.NewModuleAddress(adminmodulemoduletypes.ModuleName).String(),
 		wasmOpts...,
 	)
 	wasmHooks.ContractKeeper = &app.WasmKeeper
 
 	app.CronKeeper.WasmMsgServer = wasmkeeper.NewMsgServerImpl(&app.WasmKeeper)
 	cronModule := cron.NewAppModule(appCodec, &app.CronKeeper)
 
-	if len(enabledProposals) != 0 {
```
mmikev, line 699: `enabledProposals` variable is not used anymore, remove it please

app/proposals_allowlisting.go
```diff
@@ -41,6 +36,20 @@ func isConsumerParamChangeWhitelisted(paramChanges []proposal.ParamChange) bool
 	return true
 }
 
+func isSdkMessageWhitelisted(msg sdk.Msg) bool {
```
mmikev, line 41: please add a comment what is the purpose of this method.

app/proposals_allowlisting.go
```diff
@@ -41,6 +36,20 @@ func isConsumerParamChangeWhitelisted(paramChanges []proposal.ParamChange) bool
...
+	switch msg.(type) {
```
mmikev, line 42: We definitely need to whitelist more messages here, but it's out of scope of this PR, will do it later

wasmbinding/message_plugin.go
```diff
@@ -295,6 +294,34 @@ func (m *CustomMessenger) submitTx(ctx sdk.Context, contractAddr sdk.AccAddress,
 }
 
 func (m *CustomMessenger) submitAdminProposal(ctx sdk.Context, contractAddr sdk.AccAddress, submitAdminProposal *bindings.SubmitAdminProposal) ([]sdk.Event, [][]byte, error) {
+	var data []byte
+	if submitAdminProposal.AdminProposal.ParamChangeProposal != nil || submitAdminProposal.AdminProposal.UpgradeProposal != nil || submitAdminProposal.AdminProposal.ClientUpdateProposal != nil {
```
mmikev: Please add a comment there why exactly we need that condition.

mmikev: Also i think it's cleaner to create a method `isLegacyProposal()` which does this check inside of it

wasmbinding/message_plugin.go
```diff
@@ -399,60 +399,56 @@ func (m *CustomMessenger) performSubmitAdminProposal(ctx sdk.Context, contractAd
...
-			Title:       p.Title,
-			Description: p.Description,
-			Contract:    p.Contract,
-		})
-		if err != nil {
-			return nil, errors.Wrap(err, "failed to set content on ClearAdminProposal")
-		}
+	var msg *admintypes.MsgSubmitProposal
+	var sdkMsgs []sdk.Msg
+	var sdkMsg sdk.Msg
```
mmikev:
```suggestion
	var (
	    msg *admintypes.MsgSubmitProposal
	    sdkMsgs []sdk.Msg
	    sdkMsg sdk.Msg
	)
	
```

wasmbinding/message_plugin.go
```diff
@@ -399,60 +399,56 @@ func (m *CustomMessenger) performSubmitAdminProposal(ctx sdk.Context, contractAd
...
+	cdc := m.AdminKeeper.Codec()
+	err = cdc.UnmarshalInterfaceJSON([]byte(proposal.ProposalExecuteMessage.Message), &sdkMsg)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to unmarshall incoming sdk message")
+	}
+
+	signers := sdkMsg.GetSigners()
+	if len(signers) != 1 {
+		return nil, errors.Wrap(err, "should be 1 signer")
```
mmikev:
```suggestion
		return nil, errors.Wrap(sdkerrors.ErrorInvalidSigner, "should be 1 signer")
```

wasmbinding/message_plugin.go
```diff
@@ -399,60 +399,56 @@ func (m *CustomMessenger) performSubmitAdminProposal(ctx sdk.Context, contractAd
...
+	}
+	if !signers[0].Equals(authority) {
+		return nil, errors.Wrap(err, "authority in incoming msg is not equal to admin module")
```
mmikev:
```suggestion
		return nil, errors.Wrap(sdkerrors.ErrUnauthorized, "authority in incoming msg is not equal to admin module")
```

wasmbinding/message_plugin.go
mmikev: These two conditions differ only in the method that should called (`m.performSubmitAdminProposalLegacy()` or `m.performSubmitAdminProposal()`). 

Logging and unmarshalling logic is the same for both conditions. Let's refactor to make it easier to read?

pr0n00gler (changes requested):

app/app.go
```diff
@@ -170,23 +169,6 @@ var (
 	EnableSpecificProposals = ""
 )
 
-// GetEnabledProposals parses the ProposalsEnabled / EnableSpecificProposals values to
-// produce a list of enabled proposals to pass into wasmd app.
-func GetEnabledProposals() []wasm.ProposalType {
```
swelf19, line 175: do you think its safe to remove `EnableSpecificProposals` and `ProposalsEnabled` variables. and also remove all mentions in Makefile

app/app.go
```diff
@@ -443,7 +424,7 @@ func New(
 		authtypes.ProtoBaseAccount,
 		maccPerms,
 		sdk.GetConfig().GetBech32AccountAddrPrefix(),
-		authtypes.NewModuleAddress(govtypes.ModuleName).String(),
+		authtypes.NewModuleAddress(adminmodulemoduletypes.ModuleName).String(),
```
NeverHappened: i feel like import names everywhere should be `adminmoduletypes` and `adminmodulekeeper`

wasmbinding/message_plugin.go
```diff
@@ -797,6 +777,12 @@ func (m *CustomMessenger) validateProposalQty(proposal *bindings.AdminProposal)
 
 	return fmt.Errorf("more than one admin proposal type is present in message")
```
sotnikov-s: please refactor it using switch statement

wasmbinding/message_plugin.go
```diff
@@ -295,7 +294,40 @@ func (m *CustomMessenger) submitTx(ctx sdk.Context, contractAddr sdk.AccAddress,
 }
 
 func (m *CustomMessenger) submitAdminProposal(ctx sdk.Context, contractAddr sdk.AccAddress, submitAdminProposal *bindings.SubmitAdminProposal) ([]sdk.Event, [][]byte, error) {
-	response, err := m.performSubmitAdminProposal(ctx, contractAddr, submitAdminProposal)
+	var data []byte
+	err := m.validateProposalQty(&submitAdminProposal.AdminProposal)
+	if err != nil {
+		return nil, nil, errors.Wrap(err, "failed to validate proposal quantity")
```
sotnikov-s: it's not failed to validate, it's the other way around: there was a validation and it was processed from the beginning to the end, but the argument didn't pass the validation. so I think it would be more meaningful just to return a "invalid proposal quantity" text here

wasmbinding/message_plugin.go
```diff
@@ -295,7 +294,40 @@ func (m *CustomMessenger) submitTx(ctx sdk.Context, contractAddr sdk.AccAddress,
...
+	}
+	// here we handle pre-sdk47 style of proposals: param change & upgrade/cancel
+	if m.isLegacyProposal(&submitAdminProposal.AdminProposal) {
+		resp, err := m.performSubmitAdminProposalLegacy(ctx, contractAddr, submitAdminProposal)
+		if err != nil {
+			ctx.Logger().Debug("performSubmitAdminProposalLegacy: failed to submitAdminProposal",
+				"from_address", contractAddr.String(),
+				"creator", contractAddr.String(),
```
sotnikov-s: do we need to log both `from_address` and `creator` given a fact they are the same?

wasmbinding/message_plugin.go
```diff
@@ -295,7 +294,40 @@ func (m *CustomMessenger) submitTx(ctx sdk.Context, contractAddr sdk.AccAddress,
...
+			return nil, nil, errors.Wrap(err, "failed to submit admin proposal legacy")
+		}
+		data, err = json.Marshal(resp)
+		if err != nil {
+			ctx.Logger().Error("json.Marshal: failed to marshal submitAdminProposalLegacy response to JSON",
+				"from_address", contractAddr.String(),
+				"creator", contractAddr.String(),
+				"error", err,
+			)
+			return nil, nil, errors.Wrap(err, "marshal json failed")
```
sotnikov-s, line 326: we never return any events from this func. can we remove the first return value?

wasmbinding/message_plugin.go
```diff
@@ -295,7 +294,40 @@ func (m *CustomMessenger) submitTx(ctx sdk.Context, contractAddr sdk.AccAddress,
 }
 
 func (m *CustomMessenger) submitAdminProposal(ctx sdk.Context, contractAddr sdk.AccAddress, submitAdminProposal *bindings.SubmitAdminProposal) ([]sdk.Event, [][]byte, error) {
```
swelf19: The only field from whole `bindings.SubmitAdminProposal` you are using is `AdminProposal`, may be its cleaner pass the field itself, not a whole struct

wasmbinding/message_plugin.go
```diff
@@ -399,60 +400,53 @@ func (m *CustomMessenger) performSubmitAdminProposal(ctx sdk.Context, contractAd
...
-		})
-		if err != nil {
-			return nil, errors.Wrap(err, "failed to set content on ClearAdminProposal")
-		}
+	signers := sdkMsg.GetSigners()
+	if len(signers) != 1 {
+		return nil, errors.Wrap(sdkerrors.ErrorInvalidSigner, "should be 1 signer")
+	}
+	if !signers[0].Equals(authority) {
+		return nil, errors.Wrap(sdkerrors.ErrUnauthorized, "authority in incoming msg is not equal to admin module")
```
swelf19, line 439: i'm not sure, authentication does make any sense this place.
1) We(chain) trust proposer
2) Proposer constructs a msg and can put any value in "signer" field.
3) only admin(dao) can execute this proposals

From my point of view, right place to do a validation is ` func (m AdminMsgServer) SubmitProposal(...)` method 
Because in general use you are not limited with wambindings, any admin is able to interact with `AdminMsgServer`
Also, this particular validation happens only after proposal has passed dao voting stage, am i right? It would be nice to simulate proposal during creation

wasmbinding/message_plugin.go
quasisamurai: no, bc there is such requirement  in `DispatchMsg` func

wasmbinding/message_plugin.go
quasisamurai: makes no sense for me.  it actually will reduce the code readability

app/app.go
quasisamurai: why it might be unsafe? wasm.ProposalType is basically deprecated

wasmbinding/message_plugin.go
```diff
@@ -294,42 +293,65 @@ func (m *CustomMessenger) submitTx(ctx sdk.Context, contractAddr sdk.AccAddress,
...
 		"from_address", contractAddr.String(),
-		"creator", contractAddr.String(),
 	)
 	return nil, [][]byte{data}, nil
 }
 
-func (m *CustomMessenger) performSubmitAdminProposal(ctx sdk.Context, contractAddr sdk.AccAddress, submitAdminProposal *bindings.SubmitAdminProposal) (*admintypes.MsgSubmitProposalResponse, error) {
-	msg := admintypes.MsgSubmitProposal{Proposer: contractAddr.String()}
-	proposal := submitAdminProposal.AdminProposal
+func (m *CustomMessenger) performSubmitAdminProposalLegacy(ctx sdk.Context, contractAddr sdk.AccAddress, adminProposal *bindings.AdminProposal) (*admintypes.MsgSubmitProposalLegacyResponse, error) {
```
sotnikov-s, line 359: god damn github multiline changes. here's another switch improvement from me :) couldn't make it the suggest changes way, so simply attaching a screenshot. 

I don't like the way we handle here non-nil proposal fields: check each of them sequentially with an ordinary if, not if-else statement. so from the first glance it looks like we can possibly mutate the msg content multiple times. I know that we validate it outside of the func so here it should be only one of the fields, but it's not good to split logic between the funcs in such a way. this simple improvement will make this concrete func self-sufficient in regard of proposal fields handling:
image

<details closed>
<summary>here's the whole func:</summary>

```golang
func (m *CustomMessenger) performSubmitAdminProposalLegacy(ctx sdk.Context, contractAddr sdk.AccAddress, adminProposal *bindings.AdminProposal) (*admintypes.MsgSubmitProposalLegacyResponse, error) {
	proposal := adminProposal
	msg := admintypes.MsgSubmitProposalLegacy{Proposer: contractAddr.String()}

	switch {
	case proposal.ParamChangeProposal != nil:
		p := proposal.ParamChangeProposal
		err := msg.SetContent(&paramChange.ParameterChangeProposal{
			Title:       p.Title,
			Description: p.Description,
			Changes:     p.ParamChanges,
		})
		if err != nil {
			return nil, errors.Wrap(err, "failed to set content on ParameterChangeProposal")
		}
	case proposal.UpgradeProposal != nil:
		p := proposal.UpgradeProposal
		err := msg.SetContent(&ibcclienttypes.UpgradeProposal{
			Title:       p.Title,
			Description: p.Description,
			Plan: softwareUpgrade.Plan{
				Name:   p.Plan.Name,
				Height: p.Plan.Height,
				Info:   p.Plan.Info,
			},
			UpgradedClientState: p.UpgradedClientState,
		})
		if err != nil {
			return nil, errors.Wrap(err, "failed to set content on UpgradeProposal")
		}
	case proposal.ClientUpdateProposal != nil:
		p := proposal.ClientUpdateProposal
		err := msg.SetContent(&ibcclienttypes.ClientUpdateProposal{
			Title:              p.Title,
			Description:        p.Description,
			SubjectClientId:    p.SubjectClientId,
			SubstituteClientId: p.SubstituteClientId,
		})
		if err != nil {
			return nil, errors.Wrap(err, "failed to set content on ClientUpdateProposal")
		}
	default:
		return nil, errors.Wrapf(sdkerrors.ErrInvalidRequest, "unexpected legacy admin proposal structure: %+v", proposal)
	}

	if err := msg.ValidateBasic(); err != nil {
		return nil, errors.Wrap(err, "failed to validate incoming SubmitAdminProposal message")
	}

	response, err := m.Adminserver.SubmitProposalLegacy(sdk.WrapSDKContext(ctx), &msg)
	if err != nil {
		return nil, errors.Wrap(err, "failed to submit proposal")
	}

	ctx.Logger().Debug("submit proposal legacy processed in msg server",
		"from_address", contractAddr.String(),
	)

	return response, nil
}
```

</details>

wasmbinding/message_plugin.go
```diff
@@ -399,60 +394,52 @@ func (m *CustomMessenger) performSubmitAdminProposal(ctx sdk.Context, contractAd
...
-		}
+	signers := sdkMsg.GetSigners()
+	if len(signers) != 1 {
+		return nil, errors.Wrap(sdkerrors.ErrorInvalidSigner, "should be 1 signer")
+	}
+	if !signers[0].Equals(authority) {
+		return nil, errors.Wrap(sdkerrors.ErrUnauthorized, "authority in incoming msg is not equal to admin module")
 	}
+	sdkMsgs = append(sdkMsgs, sdkMsg)
+	msg, err = admintypes.NewMsgSubmitProposal(sdkMsgs, contractAddr)
```
sotnikov-s: you can simplify the func by removing the sdkMsgs variable and simply passing the sdkMsg in this method as
```golang
admintypes.NewMsgSubmitProposal([]sdk.Msg{sdkMsg}, contractAddr)
```

sotnikov-s: unhandled error

wasmbinding/message_plugin.go
sotnikov-s: I agree on the fact such a validation belongs to the module, message plugin should make only the basic validation which is applicable for it and simply proxy messages to corresponding modules. We should have admin module specific validation in the admin module itself and only there so we 1) don't copy+paste -> maintain it more easy 2) don't put module specific logic to message plugin

wasmbinding/message_plugin.go
```diff
@@ -294,42 +293,65 @@ func (m *CustomMessenger) submitTx(ctx sdk.Context, contractAddr sdk.AccAddress,
...
 
-func (m *CustomMessenger) submitAdminProposal(ctx sdk.Context, contractAddr sdk.AccAddress, submitAdminProposal *bindings.SubmitAdminProposal) ([]sdk.Event, [][]byte, error) {
-	response, err := m.performSubmitAdminProposal(ctx, contractAddr, submitAdminProposal)
+func (m *CustomMessenger) submitAdminProposal(ctx sdk.Context, contractAddr sdk.AccAddress, adminProposal *bindings.AdminProposal) ([]sdk.Event, [][]byte, error) {
+	var data []byte
+	err := m.validateProposalQty(adminProposal)
+	if err != nil {
+		return nil, nil, errors.Wrap(err, "failed to validate proposal quantity")
+	}
+	// here we handle pre-sdk47 style of proposals: param change & upgrade/cancel
```
sotnikov-s: > param change & upgrade/cancel

`param change, upgrade, client update`?

app/app.go
```diff
@@ -612,7 +593,7 @@ func New(
 		app.BankKeeper,
 		// 25% of rewards should be sent to the redistribute address
 		rewardsaddressprovider.NewFixedAddressRewardsAddressProvider(app.AccountKeeper.GetModuleAddress(ccvconsumertypes.ConsumerRedistributeName)),
-		authtypes.NewModuleAddress(govtypes.ModuleName).String(),
+		authtypes.NewModuleAddress(adminmodulemoduletypes.ModuleName).String(),
```
NeverHappened: maybe `adminmoduletypes`?

wasmbinding/message_plugin.go
```diff
@@ -399,60 +394,55 @@ func (m *CustomMessenger) performSubmitAdminProposal(ctx sdk.Context, contractAd
...
+
+	return response, nil
+}
+
+func (m *CustomMessenger) performSubmitAdminProposal(ctx sdk.Context, contractAddr sdk.AccAddress, adminProposal *bindings.AdminProposal) (*admintypes.MsgSubmitProposalResponse, error) {
+	proposal := adminProposal
+	authority := authtypes.NewModuleAddress(admintypes.ModuleName)
+	var (
+		msg     *admintypes.MsgSubmitProposal
+		sdkMsgs []sdk.Msg
```
NeverHappened: unused

app/app.go
```diff
@@ -114,10 +113,10 @@ import (
 	tokenfactorykeeper "github.com/neutron-org/neutron/x/tokenfactory/keeper"
 	tokenfactorytypes "github.com/neutron-org/neutron/x/tokenfactory/types"
 
-	adminmodulemodule "github.com/cosmos/admin-module/x/adminmodule"
+	"github.com/cosmos/admin-module/x/adminmodule"
 	adminmodulecli "github.com/cosmos/admin-module/x/adminmodule/client/cli"
-	adminmodulemodulekeeper "github.com/cosmos/admin-module/x/adminmodule/keeper"
-	adminmodulemoduletypes "github.com/cosmos/admin-module/x/adminmodule/types"
+	adminmodulekeeper "github.com/cosmos/admin-module/x/adminmodule/keeper"
+	admintypes "github.com/cosmos/admin-module/x/adminmodule/types"
```
NeverHappened: adminmoduletypes

NeverHappened (commented):
There are lots of `app.GetEnabledProposals(),` in testing code that needs to be fixed

pr0n00gler (approved):

sotnikov-s (approved):

PR #304, opened 2023-08-18
feat: ictx + transfer with limited sudo call #NTRN-85

swelf19:
In order to prevent malicious contract to overuse ibc sudo call, and run heavy computing at relayer's expense, we run sudo call with certain amount of gas(dao controlled) - `SudoCallLimit`
We run sudo call with `cacheCtx` with limited amount of gas
`_, err = im.ContractManagerKeeper.SudoTimeout(cacheCtx, senderAddress, packet)`
If the contract tries to consume more gas than allowed with `SudoCallLimit`, then panic is initiated by `wasmvm`
We converts the `out of gas` panic(and only this), and process it later as general error returned from the contract, i.e. suppress the error, save failure to the storage, pay fees to relayer.
In case of any other kind of panic, we treat it as relayers fault, do not recover it. Transaction aborted.

test run: https://github.com/neutron-org/neutron-tests/actions/runs/5939595766


swelf19:
https://github.com/neutron-org/neutron-tests/actions/runs/6010032342

x/transfer/ibc_handlers.go
```diff
@@ -138,25 +75,55 @@ func (im IBCModule) HandleTimeout(ctx sdk.Context, packet channeltypes.Packet, r
...
+
+// createCachedContext creates a cached context for handling Sudo calls to CosmWasm smart-contracts.
+// If there is an error during Sudo call, we can safely revert changes made in cached context.
+func (im IBCModule) createCachedContext(ctx sdk.Context) (sdk.Context, func(), sdk.GasMeter) {
+	cacheCtx, writeFn := ctx.CacheContext()
+	sudoLimit := im.ContractManagerKeeper.GetParams(ctx).SudoCallGasLimit
+	gasMeter := sdk.NewGasMeter(sudoLimit)
+	cacheCtx = cacheCtx.WithGasMeter(gasMeter)
+	return cacheCtx, writeFn, gasMeter
+}
```
mmikev: We have two identical methods: one in ICTX module and another one in transfer module. Is there any way to avoid copy-paste and make these methods reusable on both modules?

pr0n00gler (changes requested):

network/init-neutrond.sh
```diff
@@ -637,12 +637,11 @@ set_genesis_param signed_blocks_window                   "\"$SLASHING_SIGNED_BLO
 set_genesis_param min_signed_per_window                  "\"$SLASHING_MIN_SIGNED\","                   # slashing
 set_genesis_param slash_fraction_double_sign             "\"$SLASHING_FRACTION_DOUBLE_SIGN\","         # slashing
 set_genesis_param slash_fraction_downtime                "\"$SLASHING_FRACTION_DOWNTIME\""             # slashing
-set_genesis_param minimum_gas_prices                     "$MIN_GAS_PRICES,"                            # globalfee
-set_genesis_param max_total_bypass_min_fee_msg_gas_usage "\"$MAX_TOTAL_BYPASS_MIN_FEE_MSG_GAS_USAGE\"" # globalfee
+set_genesis_param minimum_gas_prices                     "$MIN_GAS_PRICES,"                             # globalfeeset_genesis_param max_total_bypass_min_fee_msg_gas_usage "\"$MAX_TOTAL_BYPASS_MIN_FEE_MSG_GAS_USAGE\"" # globalfee
```
NeverHappened: this line seems confusing - it merged into one?

app/upgrades/sdk47/upgrades.go
```diff
@@ -12,6 +12,7 @@ import (
 	upgradetypes "github.com/cosmos/cosmos-sdk/x/upgrade/types"
 	v6 "github.com/cosmos/ibc-go/v7/modules/apps/27-interchain-accounts/controller/migrations/v6"
 	"github.com/neutron-org/neutron/app/upgrades"
+	contractmanagermoduletypes "github.com/neutron-org/neutron/x/contractmanager/types"
```
NeverHappened: how about just `contractmanagertypes`?

neutronutils/utils.go
```diff
@@ -0,0 +1,11 @@
+package neutronutils
+
+import sdk "github.com/cosmos/cosmos-sdk/types"
+
+// CreateCachedContext creates a cached context for with a limited gas meter.
```
NeverHappened:
```suggestion
// CreateCachedContext creates a cached context with a limited gas meter.
```

x/contractmanager/types/params.go
```diff
@@ -7,14 +7,18 @@ import (
...
 
+const DefaultSudoCallGasLimit = uint64(1_000_000)
+
 // ParamKeyTable the param key table for launch module
 func ParamKeyTable() paramtypes.KeyTable {
 	return paramtypes.NewKeyTable().RegisterParamSet(&Params{})
 }
 
 // NewParams creates a new Params instance
 func NewParams() Params {
```
NeverHappened: maybe make this function accept sudoGasLimit as an argument, and use it in DefaultParams()?

pr0n00gler (approved):

x/transfer/ibc_handlers_test.go
```diff
@@ -96,25 +94,21 @@ func TestHandleAcknowledgement(t *testing.T) {
 	require.NoError(t, err)
 	p.Data = tokenBz
 
-	// error during SudoResponse non contract
+	//// error during SudoResponse non contract
```
NeverHappened: too many `//` :)

NeverHappened (changes requested):
All looks great! Few minor changes requested + need to resolve conflicts

swelf19:
https://github.com/neutron-org/neutron-tests/actions/runs/6039808703

swelf19:
https://github.com/neutron-org/neutron-tests/actions/runs/6046723887

NeverHappened (approved):
LGTM!

pr0n00gler (approved):

app/upgrades/sdk47/upgrades.go
```diff
@@ -91,6 +92,15 @@ func CreateUpgradeHandler(
 			return nil, err
 		}
 
+		ctx.Logger().Info("Setting sudo callback limit...")
```
sotnikov-s, line 84: would be nice to have migration code for each module in a dedicated private func (e.g. as the `migrateCronParams` below) for better readability and structure. so please move either your addition (related to the `contractmanager` module) or all of them (+ `feeburner` and `builder`) if you feel this way

app/upgrades/sdk47/upgrades.go
```diff
@@ -91,6 +92,15 @@ func CreateUpgradeHandler(
...
+		cmParams := contractmanagertypes.Params{
+			SudoCallGasLimit: 1_000_000,
```
sotnikov-s: would be cleaner to use the `DefaultSudoCallGasLimit` you've defined

neutronutils/errors/errors.go
```diff
@@ -0,0 +1,21 @@
+package errors
+
+import (
+	"cosmossdk.io/errors"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+// OutOfGasRecovery converts `out of gas` panic into an error
+// leave unprocessed any other kinds of panics
```
sotnikov-s: leave -> leaving (otherwise sounds like a recommendation or something)

neutronutils/utils.go
```diff
@@ -0,0 +1,11 @@
+package neutronutils
+
+import sdk "github.com/cosmos/cosmos-sdk/types"
+
+// CreateCachedContext creates a cached context with a limited gas meter.
+func CreateCachedContext(ctx sdk.Context, gasLimit uint64) (sdk.Context, func(), sdk.GasMeter) {
+	cacheCtx, writeFn := ctx.CacheContext()
+	gasMeter := sdk.NewGasMeter(gasLimit)
+	cacheCtx = cacheCtx.WithGasMeter(gasMeter)
+	return cacheCtx, writeFn, gasMeter
```
sotnikov-s: gasMeter is accessible via `cacheCtx.GasMeter()`. what's the point of returning it either accessing via this method? imo from the caller point of view it would be more concise (2 returned values instead of 3) and clean (explicit gas meter getting instead of some implicit value returned) to use the method

x/transfer/ibc_handlers.go
```diff
@@ -96,41 +30,43 @@ func (im IBCModule) HandleAcknowledgement(ctx sdk.Context, packet channeltypes.P
 	if err != nil {
 		return errors.Wrapf(sdkerrors.ErrInvalidAddress, "failed to decode address from bech32: %v", err)
 	}
+	if !im.ContractManagerKeeper.HasContractInfo(ctx, senderAddress) {
+		return nil
+	}
 
-	cacheCtx, writeFn, newGasMeter := im.createCachedContext(ctx)
-	defer im.outOfGasRecovery(ctx, newGasMeter, senderAddress, packet, data, contractmanagertypes.Ack, &ack)
+	cacheCtx, writeFn, newGasMeter := neutronutils.CreateCachedContext(ctx, im.ContractManagerKeeper.GetParams(ctx).SudoCallGasLimit)
```
sotnikov-s: I think it makes sense to move `CreateCachedContext` after fee distribution to emphasise 1) the general context usage in fee distribution, not the cacheCtx, and 2) limited cacheCtx gas meter usage only in scope of sudo call. if you see eye to eye with me, please apply this change in all IBC handlers

x/transfer/ibc_handlers.go
```diff
@@ -96,41 +30,43 @@ func (im IBCModule) HandleAcknowledgement(ctx sdk.Context, packet channeltypes.P
 	if err != nil {
 		return errors.Wrapf(sdkerrors.ErrInvalidAddress, "failed to decode address from bech32: %v", err)
 	}
+	if !im.ContractManagerKeeper.HasContractInfo(ctx, senderAddress) {
```
sotnikov-s: 1. why is it needed?
2. why we return a nil err here?
3. why we don't do the same in interchaintx handlers?

neutronutils/utils.go
```diff
@@ -0,0 +1,11 @@
+package neutronutils
```
sotnikov-s: I don't think I like the dedicated package for such sudo call specific stuff as limited cached context creation and out of gas panic to error conversion. please consider moving it to contractmanager module:
- make `CreateCachedContext` a method of the `ContractManagerKeeper`, maybe use `Params.SudoCallGasLimit` inside instead of accepting it as a parameter, and maybe rename to smth like `SudoCachedContext`
- move the `OutOfGasRecovery` to the `contractmanager` package and also give it a more Sudo call related name

sotnikov-s: or even better: why do we even expose all these `contractmanager` specific stuff? why don't encapsulate both the 1) limited cached context and 2) out of gas panic in the Sudo methods of the `ContractManagerKeeper`, and handle panics and call `writeFn` inside?

sotnikov-s:
also please:
1. respect the linter errors re files formatting;
2. create a PR to the neutron-docs repo because these changes are definitely one of those we don't want to forget about to reflect there.

x/transfer/ibc_handlers.go
swelf19: Transfer module can be used by both, either a regular account or a contract. In case it's a regular account we do not need to call sudo, we do early exit.
interchaintxs module can be used by contract account only, we dont need need to do early exit.

sotnikov-s: don't we need to distribute fees before early exit?

swelf19: Regular user does not pay ibc fees, only contract does. Basically fees is a compensation for calling sudo handler by relayer. There is no callback, when origin of the ibc msg is regular account.

neutronutils/utils.go
swelf19: good thoughts, i've opened a dedicated pr https://github.com/neutron-org/neutron/pull/313 to make the changes.

swelf19:
> also please:
> 
> 1. respect the linter errors re files formatting;
> 2. create a PR to the neutron-docs repo because these changes are definitely one of those we don't want to forget about to reflect there.

1. I want to fix the linter errors after merge all major PRs into `update-sdk47` branch, cause now linter create too much noise with "deprecated" errors, which should be fixed in corresponding PRs
2. will do

swelf19:
https://github.com/neutron-org/neutron-tests/actions/runs/6236180528

swelf19:
https://github.com/neutron-org/neutron-tests/actions/runs/6237857634

sotnikov-s (approved):

pr0n00gler (approved):

PR #313, opened 2023-09-08
feat: interachain sudo callback refactoring

swelf19:
- [x] extracted "an executing sudo in cached limited context" into middleware
- [x] removed redundant "Is it a contract address" checks
- [x] merge SudoResponse/SudoError/SudoTimeout into single handler
- [x] added requirement to pay fee to register interchain account 

https://github.com/neutron-org/neutron-tests/actions/runs/6196055940

related PRs:
https://github.com/neutron-org/neutron-sdk/pull/112
https://github.com/neutron-org/neutron-dev-contracts/pull/27
https://github.com/neutron-org/neutron-integration-tests/pull/204


Makefile
```diff
@@ -123,7 +123,7 @@ build-static-linux-amd64: go.sum $(BUILDDIR)/
 	$(DOCKER) cp neutronbinary:/bin/neutrond $(BUILDDIR)/neutrond-linux-amd64
 	$(DOCKER) rm -f neutronbinary
 
-install-test-binary: check_version go.sum
```
mmikev: Why is `check_version` removed?

x/contractmanager/ibc_middleware_test.go
```diff
@@ -0,0 +1,83 @@
+package contractmanager
+
+//
+//import (
```
mmikev: Why is it all commented out?

pr0n00gler (dismissed):

proto/neutron/interchaintxs/v1/tx.proto
```diff
@@ -26,6 +27,10 @@ message MsgRegisterInterchainAccount {
   string connection_id = 2 [ (gogoproto.moretags) = "yaml:\"connection_id\"" ];
   string interchain_account_id = 3
       [ (gogoproto.moretags) = "yaml:\"interchain_account_id\"" ];
+  repeated cosmos.base.v1beta1.Coin register_fee = 4 [
+    (gogoproto.nullable) = false,
+    (gogoproto.castrepeated) = "github.com/cosmos/cosmos-sdk/types.Coins"
+  ];
```
sotnikov-s, line 33: what's the point of having this kind of parameter for the message? can't we just use the `register_fee` from the module params?

x/contractmanager/keeper/failure.go
```diff
@@ -71,30 +68,12 @@ func (k Keeper) GetFailure(ctx sdk.Context, contractAddr sdk.AccAddress, id uint
...
-			}
-		}
-	case types.Timeout:
-		if _, err := k.SudoTimeout(ctx, contractAddr, *failure.Packet); err != nil {
-			return errors.Wrapf(types.FailedToResubmitFailure, "cannot resubmit failure ack timeout; failureId = %d; err = %s", failure.Id, err)
-		}
-	default:
-		return errors.Wrapf(types.IncorrectAckType, "cannot resubmit failure with incorrect ackType = %s", failure.GetAckType())
+	if _, err := k.wasmKeeper.Sudo(ctx, contractAddr, failure.SudoPayload); err != nil {
+		return errors.Wrapf(types.FailedToResubmitFailure, "cannot resubmit failure ack; failureId = %d; err = %s", failure.Id, err)
```
sotnikov-s: cannot resubmit failure **ack** -> cannot resubmit failure

x/contractmanager/keeper/sudo.go
```diff
@@ -20,118 +20,37 @@ func (k Keeper) HasContractInfo(ctx sdk.Context, contractAddress sdk.AccAddress)
...
-			"error", err, "contract_address", senderAddress)
-		return nil, fmt.Errorf("failed to Sudo: %v", err)
+		return nil, fmt.Errorf("failed to marshal MessageOnChanOpenAck: %v", err)
 	}
-
-	return resp, nil
+	return m, nil
 }
 
 func (k Keeper) SudoOnChanOpenAck(
```
sotnikov-s: it's not used anymore, is it?

x/contractmanager/keeper/sudo.go
```diff
@@ -20,118 +20,37 @@ func (k Keeper) HasContractInfo(ctx sdk.Context, contractAddress sdk.AccAddress)
 	return k.wasmKeeper.HasContractInfo(ctx, contractAddress)
```
sotnikov-s, line 20: probably not in scope of this PR, just want to share my observations:

I think we should remove the `HasContractInfo` method from the `contractmanager.Keeper` interface for it's just a proxy, no specific logic inside it. this could 1) make the `HasContractInfo` calls less ambiguous (why do we ask for a contract info from the `contractmanager` module, not from the `wasm` one, given the fact it's the `wasm` module which owns contracts) and 2) simplify `contractmanager.Keeper` interface, e.g. make it cleaner, more `contractmanager` module specialisation-oriented

sotnikov-s: another suggestion: I don't see any reasons of `SudoKVQueryResult` and `SudoTxQueryResult` in being placed in the `contractmanager` module. I suggest moving them to the `interchainqueries` module — to the only place they are used and to the place IMO they belong to in terms of module responsibility and scope

x/contractmanager/keeper/sudo.go
```diff
@@ -20,118 +20,37 @@ func (k Keeper) HasContractInfo(ctx sdk.Context, contractAddress sdk.AccAddress)
...
-
-	if !k.wasmKeeper.HasContractInfo(ctx, senderAddress) {
-		if request.SourcePort == types.TransferPort {
-			// we want to allow non contract account to send the assets via IBC Transfer module
-			// we can determine the originating module by the source port of the request packet
-			return nil, nil
-		}
-		k.Logger(ctx).Debug("SudoError: contract not found", "senderAddress", senderAddress)
-		return nil, fmt.Errorf("%s is not a contract address and not the Transfer module", senderAddress)
+func PrepareOpenAckCallbackMessage(details types.OpenAckDetails) ([]byte, error) {
```
sotnikov-s, line 45: IMO the `PrepareOpenAckCallbackMessage` func belongs to the `interchaintxs` module

x/contractmanager/types/module.go
```diff
@@ -0,0 +1,9 @@
+package types
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+type Sudo interface {
+	Sudo(ctx sdk.Context, contractAddress sdk.AccAddress, msg []byte) ([]byte, error)
+}
```
sotnikov-s: the interface type is not used

x/contractmanager/types/sudo.go
```diff
@@ -25,30 +23,26 @@ type MessageKVQueryResult struct {
...
 // transaction failed with a timeout.
-type MessageTimeout struct {
-	Timeout struct {
-		Request channeltypes.Packet `json:"request"`
-	} `json:"timeout"`
+type MessageSudoCallback struct {
+	Response *ResponseSudoPayload `json:"response,omitempty"`
+	Error    *ErrorSudoPayload    `json:"error,omitempty"`
+	Timeout  *TimeoutPayload      `json:"timeout,omitempty"`
+}
```
sotnikov-s: please update the description

x/contractmanager/keeper/sudo.go
sotnikov-s: the `PrepareSudoCallbackMessage` func also seems not at the right place for me, but I couldn't come up with a better one since it's used by two modules (not by a single one as `PrepareOpenAckCallbackMessage`). but since they are quite similar, I think we can either keep the both here or find better places for both of them. a better place for `PrepareOpenAckCallbackMessage` is obvious — `interchaintxs` module. any ideas on `PrepareSudoCallbackMessage`?

x/interchaintxs/keeper/keeper.go
```diff
@@ -36,20 +40,46 @@ func NewKeeper(
...
+
+	treasury := k.feeBurnerKeeper.GetParams(ctx).TreasuryAddress
+	treasuryAddress, err := sdk.AccAddressFromBech32(treasury)
+	if err != nil {
+		return errors.Wrapf(sdkerrors.ErrInvalidAddress, "failed to convert treasury, bech32 to AccAddress: %v: %v", treasury, err)
+	}
+
+	err = k.bankKeeper.SendCoins(ctx, payer, treasuryAddress, fee)
+	if err != nil {
+		return errors.Wrapf(err, "failed send fee(%v) from %v to %v", fee, payer, treasury)
```
sotnikov-s: better to use `%s` for all three params

x/interchaintxs/keeper/keeper.go
```diff
@@ -36,20 +40,46 @@ func NewKeeper(
 	memKey storetypes.StoreKey,
 	channelKeeper types.ChannelKeeper,
 	icaControllerKeeper types.ICAControllerKeeper,
-	contractManagerKeeper types.ContractManagerKeeper,
+	contractManagerKeeper types.WasmKeeper,
```
sotnikov-s: contractManagerKeeper -> sudoKeeper

x/interchaintxs/types/params.go
```diff
@@ -12,29 +14,20 @@ var _ paramtypes.ParamSet = (*Params)(nil)
 var (
 	KeyMsgSubmitTxMaxMessages     = []byte("MsgSubmitTxMaxMessages")
 	DefaultMsgSubmitTxMaxMessages = uint64(16)
+	DefaultRegisterFee            = sdk.NewCoins(sdk.NewCoin(params.DefaultDenom, sdk.NewInt(1000)))
```
sotnikov-s, line 17: isn't it too little?

proto/neutron/interchaintxs/v1/tx.proto
sotnikov-s: ah, I remember that we discussed an advantage of having such a parameter: if the fee increases one day, the smart contracts with the previously set fee will stop working thus making their owners to increase the fee manually thus forcing the owners to, so to say, acknowledge the change. otherwise a significant change of this parameter could bring an unexpected and implicit operational costs change.

whereas I was the one who arose this point, now I don't think it makes sense: this so to say acknowledgement will bring huge operational load on protocols running their contracts on neutron and particularly for the DAO-controlled ones: they will have to pass a `update_contract_params` call through governance process to increase the fee even if it changed slightly. I think that making a simple force charge without requiring contracts to pass the value explicitly is totally ok: if this parameter ever changes significantly, I think all the protocol owners will become aware of it because such a change will be only possible via a proposal on Neutron

pr0n00gler (approved):

PR #334, opened 2023-10-24
enable backwards compat feature: only contracts after migration are charged for ice creation NTRN-135

quasisamurai:
[TASK](https://hadronlabs.atlassian.net/browse/NTRN-135)

We need changes in ICTX module related to register_fee backwards compatible.

In the upgrade handler of Neutron v1.1.0 we must take the last uploaded code_id of wasm contracts and save it to the storage.

We make register_fee field optional.

When a contract calls RegisterInterchainAccount method, Neutron compares code_id of the calling contract with the saved one:

If it’s less or equal to the saved one - we allow contract to not pay fees for the ICA registration;

if code_id is greater - register_fee is required.

- register_fee is optional field in Neutron CW SDK and in the RegisterInterchainAccount message;

- last code_id is saved into the storage in the upgrade handler for v1.1.0;

- contracts with code_id less or equal then the saved one do not pay fees for ICA registration;

- contract with code_id greater then the saved one pay fees for ICA registration;

- documentation is updated;

- tests are implemented;


x/interchaintxs/keeper/keeper.go
```diff
@@ -92,3 +92,19 @@ func (k Keeper) ChargeFee(ctx sdk.Context, payer sdk.AccAddress, fee sdk.Coins)
...
+	bytes := store.Get(types.LastCodeIdBeforeUpgrade)
+
+	if bytes == nil {
+		k.Logger(ctx).Debug("Last code id key don't exists, GetLastCodeID returns 0")
+		return 0
+	}
+	return sdk.BigEndianToUint64(bytes)
+}
+
+func (k Keeper) SetLastCodeIDBeforeUpgrade(ctx sdk.Context, id uint64) {
```
mmikev: Unused method

app/upgrades/nextupgrade/upgrades.go
```diff
@@ -238,6 +240,13 @@ func setInterchainTxsParams(ctx sdk.Context, paramsKeepers paramskeeper.Keeper,
 
 	bz := codec.MustMarshal(&currParams)
 	store.Set(interchaintxstypes.ParamsKey, bz)
+
+	wasmStore := ctx.KVStore(wasmStoreKey)
+	bzWasm := wasmStore.Get(wasmtypes.KeyLastCodeID)
+	if bzWasm == nil {
+		return fmt.Errorf("last code ID not found during the upgrade")
+	}
```
mmikev: There is no public method in wasmd keeper to use instead of this raw work with the storage?

app/upgrades/nextupgrade/upgrades_test.go
```diff
@@ -159,3 +163,56 @@ func (suite *UpgradeTestSuite) TestAdminModuleUpgrade() {
...
+
+	// register with fees
+	jsonStringAfterUpgrade := `{"register": {"connection_id":"connection-1","interchain_account_id":"test-3"}}`
+	byteEncodedMsgAfterUpgrade := []byte(jsonStringAfterUpgrade)
+	_, err = contractKeeper.Execute(ctx, contractAddressAfterUpgrade, sdk.AccAddress("neutron1weweewe"), byteEncodedMsgAfterUpgrade, sdk.NewCoins(sdk.NewCoin("untrn", sdk.NewInt(1000))))
+	suite.Require().NoError(err)
+
+	// failed register due lack of fees (fees required)
+	_, err = contractKeeper.Execute(ctx, contractAddressAfterUpgrade, sdk.AccAddress("neutron1weweewe"), byteEncodedMsgAfterUpgrade, nil)
+	suite.Error(err)
```
mmikev: Can we check here that we get exactly fee related error?

x/interchaintxs/keeper/msg_server.go
```diff
@@ -47,8 +47,10 @@ func (k Keeper) RegisterInterchainAccount(goCtx context.Context, msg *ictxtypes.
 		return nil, errors.Wrapf(ictxtypes.ErrNotContract, "%s is not a contract address", msg.FromAddress)
 	}
 
-	if err := k.ChargeFee(ctx, senderAddr, msg.RegisterFee); err != nil {
-		return nil, errors.Wrapf(err, "failed to charge fees to pay for RegisterInterchainAccount msg: %s", msg)
+	if k.sudoKeeper.GetContractInfo(ctx, senderAddr).CodeID >= k.GetLastCodeIDBeforeUpgrade(ctx) {
```
mmikev: Let's leave a small comment here why we do this

x/interchaintxs/types/keys.go
```diff
@@ -18,7 +18,9 @@ const (
 )
 
 const (
-	prefixParamsKey = iota + 1
+	prefixParamsKey               = iota + 1
+	prefixLastCodeIdBeforeUpgrade = iota + 2
```
mmikev: The same here, a little comment to describe the purpose of the contstant would be very nice in the future

pr0n00gler (changes requested):

x/contractmanager/types/expected_keepers.go
```diff
@@ -1,12 +1,14 @@
...
 
 import (
+	wasmtypes "github.com/CosmWasm/wasmd/x/wasm/types"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 )
 
 // WasmKeeper defines the expected interface needed to cam cosmwasm contracts.
 type WasmKeeper interface {
 	HasContractInfo(ctx sdk.Context, contractAddress sdk.AccAddress) bool
+	GetContractInfo(ctx sdk.Context, contractAddress sdk.AccAddress) *wasmtypes.ContractInfo
```
sotnikov-s, line 11: unused method

x/interchaintxs/keeper/keeper.go
```diff
@@ -92,3 +92,14 @@ func (k Keeper) ChargeFee(ctx sdk.Context, payer sdk.AccAddress, fee sdk.Coins)
 func (k Keeper) GetAuthority() string {
 	return k.authority
 }
+
+func (k Keeper) GetLastCodeIDBeforeUpgrade(ctx sdk.Context) (codeID uint64) {
```
sotnikov-s: I don't like the naming. which upgrade? what if we have more than one upgrade with some code ID? why do we even mention some upgrade in module keeper's interface? let's name it something like `GetLastFreeRegisterICACodeID` (there could definitely be a better naming for it, just an example) — says for itself what kind of ID it returns and hints at the context/reason for that

x/interchaintxs/keeper/msg_server.go
sotnikov-s: why `>=`? we will charge the contracts with src code uploaded before upgrade this way, won't we? seems like here a strict `>` is required

sotnikov-s (changes requested):

app/upgrades/nextupgrade/upgrades_test.go
```diff
@@ -159,3 +163,56 @@ func (suite *UpgradeTestSuite) TestAdminModuleUpgrade() {
...
+	)
+
+	suite.Require().True(ccvConsumerSubspace.Has(ctx, ccvconsumertypes.KeyRewardDenoms))
+	suite.FundAcc(sdk.AccAddress("neutron1weweewe"), sdk.NewCoins(sdk.NewCoin("untrn", sdk.NewInt(10000))))
+	// emulate mainnet/testnet state
+	ccvConsumerSubspace.Set(ctx, ccvconsumertypes.KeyRewardDenoms, &[]string{params.DefaultDenom})
+
+	var denomsBefore []string
+	ccvConsumerSubspace.Get(ctx, ccvconsumertypes.KeyRewardDenoms, &denomsBefore)
+	suite.Require().Equal(denomsBefore, []string{params.DefaultDenom})
```
sotnikov-s: you sure it's related to this test case?

app/upgrades/nextupgrade/upgrades_test.go
```diff
@@ -159,3 +163,56 @@ func (suite *UpgradeTestSuite) TestAdminModuleUpgrade() {
...
+	contractKeeper := keeper.NewDefaultPermissionKeeper(app.WasmKeeper)
+	// store contract just to increase code_id
+	_ = suite.StoreTestCode(ctx, sdk.AccAddress("neutron1_ica"), "testdata/neutron_interchain_txs.wasm")
```
sotnikov-s: why is this needed?

app/upgrades/nextupgrade/upgrades_test.go
```diff
@@ -159,3 +163,56 @@ func (suite *UpgradeTestSuite) TestAdminModuleUpgrade() {
...
+	upgrade := upgradetypes.Plan{
+		Name:   nextupgrade.UpgradeName,
+		Info:   "some text here",
+		Height: 100,
+	}
+	app.UpgradeKeeper.ApplyUpgrade(ctx, upgrade)
+
+	lastCodeID := app.InterchainTxsKeeper.GetLastCodeIDBeforeUpgrade(ctx)
+	// ensure that wasm module stores next code id
+	suite.Require().Equal(lastCodeID, codeIDBefore+1)
```
sotnikov-s, line 189: why you expect the next code if it's named the **last** one?

x/contractmanager/types/expected_keepers.go
quasisamurai: It's not.  This method is required by task itself (get the contract code id by its address)

app/upgrades/nextupgrade/upgrades_test.go
quasisamurai: to emulate that we have more than 1 contract stored before upgrade. Doesn't mandatory, but I find this at least not superfluous

upd. nvm, we already have enough contracts stored. deleted this

app/upgrades/nextupgrade/upgrades_test.go
quasisamurai: Because the keyword `last` is  inherited from wasm module. Actually, it's more like `next` id

x/interchaintxs/keeper/msg_server.go
quasisamurai: because as you noticed in the next comment, "_why you expect the next code if it's named the last one?_".  This is related. On a practical level, this value is more like "next id", [here](https://github.com/neutron-org/neutron/blob/7e91506fb6d733163bab7a4a20960558c1674d97/app/upgrades/nextupgrade/upgrades_test.go#L198) it is tested

app/upgrades/nextupgrade/upgrades_test.go
quasisamurai: firstly, I left it because every upgrade test had these lines. now removed, thx

x/contractmanager/types/expected_keepers.go
sotnikov-s: I can see that you've added this method to two expected wasm keepers: on the interchaintxs module's side and on the contractmanager module's side. the method is used in interchaintxs, but is not in contractmanager
image

x/interchaintxs/keeper/msg_server.go
```diff
@@ -47,8 +47,11 @@ func (k Keeper) RegisterInterchainAccount(goCtx context.Context, msg *ictxtypes.
 		return nil, errors.Wrapf(ictxtypes.ErrNotContract, "%s is not a contract address", msg.FromAddress)
 	}
 
-	if err := k.ChargeFee(ctx, senderAddr, msg.RegisterFee); err != nil {
-		return nil, errors.Wrapf(err, "failed to charge fees to pay for RegisterInterchainAccount msg: %s", msg)
+	// if contract is stored after [last] upgrade, we're not going charge fees for register ICA
+	if k.sudoKeeper.GetContractInfo(ctx, senderAddr).CodeID >= k.GetLastFreeRegisterICACodeID(ctx) {
```
sotnikov-s: a reminder to improve naming re Last-Current-FromID

sotnikov-s (approved):

quasisamurai:
https://github.com/neutron-org/neutron-tests/actions/runs/6745812174/job/18338461190

quasisamurai:
https://github.com/neutron-org/neutron-tests/actions/runs/6745812174

PR #336, opened 2023-10-25
fix: modify interchain keeper's accepted parameteres

Lockwarr:
Following our discussion - let interchaintxs keeper accept a more generalized interface.
This way, other chains that reuse the interchaintxs module would not be required to also use feeburner module, instead
they can implement GetTreasury function in the module that maintains the treasury address for their chain and pass it 
to the interchaintxs keeper.

cc @swelf19 @pr0n00gler


swelf19 (approved):

pr0n00gler (approved):

pr0n00gler:
https://github.com/neutron-org/neutron-tests/actions/runs/6682014928

x/feeburner/keeper/keeper.go
```diff
@@ -142,3 +142,9 @@ func (k Keeper) Logger(ctx sdk.Context) log.Logger {
 func (k Keeper) FundCommunityPool(ctx sdk.Context, amount sdk.Coins, sender sdk.AccAddress) error {
 	return k.bankKeeper.SendCoinsFromAccountToModule(ctx, sender, authtypes.FeeCollectorName, amount)
 }
+
+// GetTreasury is used to satisfy the interface requirement for interchaintxs module.
+// The single point of truth for the treasury address is in the feeburner module params.
+func (k Keeper) GetTreasury(ctx sdk.Context) string {
+	return k.GetParams(ctx).TreasuryAddress
+}
```
sotnikov-s: I don't like the fact that we're referring to interchaintx module in the feeburner module. it's not much of a modularity if we couple modules so tight. we need to find a self-sufficient way to let the interchaintx module know about the treasury. if we're aiming to make the module more independent, it must not be literally _treasury_ in the interchaintx's module context, it could be just an address where the fees go — that's how it is from interchaintx module's prospective.

I understand that this is not an issue of your contribution specifically, it's an issue of our design in the first place. I propose to try to rethink this matter. I personally will think about it later today/tomorrow and share my thoughts, but now I just want to highlight that the initial idea of making interchaintxs module more general is not reached yet. please think about it and share your thoughts/concerns/disapproval, I'll be happy to read it.

sotnikov-s (changes requested):

x/feeburner/keeper/keeper.go
sotnikov-s: I've got some time to phrase my concern and here's what I don't like in the current version of `interchaintxs` and `feeburner` modules:
1. `interchaintxs` module is aware of somewhat called treasury that is actually used as a fee collector.
2. `feeburner` module has `GetTreasury` method whereas there is already a way to get its treasury address via `GetParams` method

the above implies that we've beaten explicit and direct dependency `interchaintxs->feeburner`, but we've got another, opposite, implicit but still very strict one: `feeburner->interchaintxs` (`feeburner` now has a method made in accordance with and for needs of `interchaintxs` module requirements)

my suggestions are:
1. remove `GetTreasury` from `feeburner`
2. define the current fee collector addr getter func and add it to `interchaintxs` module's fields
3. use the `feeburnerKeeper.GetParams().TreasuryAddress` as the value for the field

I've attached a couple of PoC screens that will help you understand me:
telegram-cloud-photo-size-2-5332321895531532965-y
telegram-cloud-photo-size-2-5332321895531532966-y
telegram-cloud-photo-size-2-5332321895531532967-y
telegram-cloud-photo-size-2-5332321895531532968-y
telegram-cloud-photo-size-2-5332321895531532980-y

please note that, once again, it's a rough PoC, there is still a number of misprints e.g. `treasury` naming still occurs in interchaintxs module's code. please let me know what you think

Lockwarr: Hey, I understand your reason. And surely I can implement it this way. 

On a side note - about this "initial idea of making interchaintxs module more general":
I can see this module living in the ibc-go repository as I think they are working on a [interchaintxs](https://github.com/cosmos/ibc-go/tree/main/modules/apps/callbacks) module of their own.
So I think it will be nice to push in the same direction and have one module to be reusable by all cosmos chains if it's possible.

Maybe it will be useful if I share some concerns that we've gone through as a different blockchain that utilizes the interchaintxs module(from neutron@v1.0.4): 
First, we had to also include contractmanager and feeburner modules on our chain if we wanted to utilize interchaintxs. Maybe it's worth it to see if it should be required to also have contractmanager and feeburner modules in order to use interchaintxs module. 
We also don't have the ability yet to resubmit failed acknowledgements (I think this is managed by contractmanager)
Maybe it would be nice to have an option to not accept some messages from relayers that will be failing with X error, or at least a parameter of the interchaintxs module to indicate if failing acks should be sent to contractmanager.

Lockwarr: Hey, I understand your reason. And surely I can implement it this way. 



About this - "initial idea of making interchaintxs module more general":
I can see this module living in the ibc-go repository as I think they are working on a [interchaintxs](https://github.com/cosmos/ibc-go/tree/main/modules/apps/callbacks) module of their own.
So I think it will be nice to push in the same direction and have one module to be reusable by all cosmos chains if it's possible.

Maybe it will be useful if I share some concerns that we've gone through as a different blockchain that utilizes the interchaintxs module(from neutron@v1.0.4): 
First, we had to also include contractmanager and feeburner modules on our chain if we wanted to utilize interchaintxs. Maybe it's worth it to see if it should be required to also have contractmanager and feeburner modules in order to use interchaintxs module. 
We also don't have the ability yet to resubmit failed acknowledgements (I think this is managed by contractmanager)
Maybe it would be nice to have an option to not accept some messages from relayers that will be failing with X error, or at least a parameter of the interchaintxs module to indicate if failing acks should be sent to contractmanager.
Also, sometimes we're seeing 'out of gas' errors for some transactions which have submessages. This is a problem because the transaction gets commited partially and we have to continue from the 2nd or 3rd submessage, wherever the gas ran out.

I have not experimented with the ibc-callbacks module at all yet, so I'm not sure about the challenges they face.

mmikev: Hey!

We'll move eventually to the ibc-go callbacks implementation in one of the next releases, but not in v1.1.0 (the closest release) since it's feature freezed and we are preparing for the upgrade.

> First, we had to also include contractmanager and feeburner modules on our chain if we wanted to utilize interchaintxs. Maybe it's worth it to see if it should be required to also have contractmanager and feeburner modules in order to use interchaintxs module.

We are planning to refactor this to make those modules optional in the future.

> We also don't have the ability yet to resubmit failed acknowledgements (I think this is managed by contractmanager)

It's fixed in v1.1.0 release of Neutron (the release is not ready yet, but you can see relevant changes in `update-sdk47` branch, particularly [Resubmit Failure message](https://github.com/neutron-org/neutron/blob/f4a8acde11cde1103d5010bc04d3d18971978184/x/contractmanager/keeper/failure.go#L70))

> Also, sometimes we're seeing 'out of gas' errors for some transactions which have submessages. This is a problem because the transaction gets commited partially and we have to continue from the 2nd or 3rd submessage, wherever the gas ran out.

That's pretty weird, could you provide more details about that? (DM me them, please)

So, about the suggestions, let's implement it as @sotnikov-s described to make things a bit cleaner and reusable for your purposes, and we will refactor this completely in the future releases.
Wdyt?

Lockwarr: Okay, I'll be implementing the changes for this PR

x/interchaintxs/keeper/keeper.go
```diff
@@ -76,15 +76,15 @@ func (k Keeper) ChargeFee(ctx sdk.Context, payer sdk.AccAddress, fee sdk.Coins)
 		return errors.Wrapf(sdkerrors.ErrInsufficientFee, "provided fee is less than min governance set ack fee: %s < %s", fee, params.RegisterFee)
 	}
 
-	treasury := k.feeBurnerKeeper.GetParams(ctx).TreasuryAddress
-	treasuryAddress, err := sdk.AccAddressFromBech32(treasury)
+	feeCollector := k.getFeeCollectorAddr(ctx)
+	feeCollectorAddress, err := sdk.AccAddressFromBech32(feeCollector)
 	if err != nil {
-		return errors.Wrapf(sdkerrors.ErrInvalidAddress, "failed to convert treasury, bech32 to AccAddress: %s: %s", treasury, err.Error())
+		return errors.Wrapf(sdkerrors.ErrInvalidAddress, "failed to convert fee collecor, bech32 to AccAddress: %s: %s", feeCollector, err.Error())
```
sotnikov-s:
```suggestion
		return errors.Wrapf(sdkerrors.ErrInvalidAddress, "failed to convert fee collector, bech32 to AccAddress: %s: %s", feeCollector, err.Error())
```

sotnikov-s (approved):

pr0n00gler:
https://github.com/neutron-org/neutron-tests/actions/runs/6682014928

pr0n00gler (approved):

PR #342, opened 2023-11-01
Audit Fix: Fix coins.add usage during deposit

jcompagni10:
Every sdk.Coin.Add happening in DepositCore results in multiple traversals over all coins
in sharesIssued , and also involves allocation of additional data structures, thus resulting in quadratic
scaling of the DepositCore runtime (because it invokes sdk.Coins Add() for each issued share).
As a fix, we instead aggregate the coins as a slice and then perform aggregation and validation only once. 
We also move coin sending logic to the end of the deposit function for further efficiency.


pr0n00gler (approved):
LGTM, but please attach a testrun

utils/bank.go
```diff
@@ -105,3 +106,24 @@ func FilteredPaginateAccountBalances(
 
 	return res, nil
 }
+
+// SanitizeCoins takes an unsorted list of coins and sorts them, removes coins with amount zero and combines duplicate coins
+func SanitizeCoins(coins []sdk.Coin) sdk.Coins {
```
sotnikov-s, line 111: have you seen the [NewCoins](https://github.com/neutron-org/cosmos-sdk/blob/a02d35388763a73f65cab154a9e546079b73fb05/types/coin.go#L180) func from cosmos-sdk? it does exactly the same according to the comment: removes zero coins and sorts the rest. if you have, any reasons you decided to write your own algorithm for it (I presume there could be some subtle differences)? if you have not, do you think we can reuse it?

jcompagni10: NewCoins doesn't remove duplicates

sotnikov-s (approved):

jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/6816533411

PR #344, opened 2023-11-01
Audit Fix: prevent depositor or LP from creating unfair spreads 

jcompagni10:
Due to rounding in favor of the maker when trading against an LP or LO position
small amounts of liquidity can unfairly benefit the maker at the expense of the taker.

Unfortunately the obvious solutions of either preventing small deposits, or stopping takers from trading against these positions either do not work or create problematic DOS vectors.

Instead, we have switched our rounding logic to favor the taker. This also helps mitigate a number of existing DOS/dusting vectors. In order to protect makers, we also add a  check as part of swap to prevent takers from swapping very small amounts where rounding would work significantly in their favor.


x/dex/keeper/liquidity.go
```diff
@@ -40,6 +40,21 @@ func (k Keeper) Swap(
...
+		// Nonethless, once the remainingTakerDenom gets small enough to start causing unfair swaps
+		// it is much simpler to just abort.
+		if inAmount.IsZero() || isUnfairTruePrice(inAmount, outAmount, liq) {
+			// If they've already swapped just end the swap
+			if remainingTakerDenom.LT(maxAmountTakerDenom) {
+				break
+			}
+			// If they have not swapped anything return informative error
+			return sdk.Coin{}, sdk.Coin{}, false, types.ErrSwapAmountTooSmall
+		}
```
andrey-kuprianov: I believe this creates a DOS vector. If, due to some problem (a corner case in `PrecDec` for example) there is a possibility to inject a liquidity which would cause `isUnfairTruePrice()` to evaluate to `true`, while there is a lot of other valid liquidity after that, all swaps would terminate at this "poisonous" liquidity. I would say the right approach here is to ignore the problematic one, and continue:

```suggestion
		if inAmount.IsZero() || isUnfairTruePrice(inAmount, outAmount, liq) {
			continue
		}
```

x/dex/keeper/liquidity.go
```diff
@@ -107,3 +127,14 @@ func (k Keeper) SaveLiquidity(sdkCtx sdk.Context, liquidityI types.Liquidity) {
 		panic("Invalid liquidity type")
 	}
 }
+
+var MaxTrueMakerSpread = math_utils.MustNewPrecDecFromStr("0.005")
```
andrey-kuprianov: Would it be better to move this magic constant to the module params? This would make it possible to easily change it, if the need arises.

x/dex/keeper/liquidity.go
jcompagni10: I'm not sure I agree here @andrey-kuprianov. 
1.) Given the nature of rounding behavior and other arithmetic, I am reasonably confident there is no "magic" liquidity amount that would always yield an UnfairTruePrice for swap amounts above a relatively low bound (~1000 utokens). 
2.) If there is indeed a problematic corner-case with precDec, it actually seems like beneficial behavior to halt trading for a pair until the problem can be understood and resolved.
3.) The downside to this change is that all swaps will become more expensive because they will have to iterate through a potentially large number of pools before they reach the normal termination case. If there is indeed a possibility of creating poisonous pools this unbounded iteration represents a similarly problematic DOS risk. 
4.) I would not label this a DOS vector, worst case scenario it renders a pool un-tradeable until more liquidity is deposited in the "poison pool" but it does not carry a risk of halting the chain or otherwise degrading performance.

Nonetheless, I do agree that this highlights the importance of good monitoring here so the problem can be quickly brought to our attention and resolved.

x/dex/keeper/liquidity.go
jcompagni10: 👍

x/dex/keeper/liquidity.go
andrey-kuprianov: Ok, this of course also makes sense. Those are two different priorities: `continue` prioritizes for liveness, while `break` prioritizes for safety. I agree with your considerations above: if you are reasonably sure in the math computations, then prioritizing for safety and fulfilling the secondary objective of making the swaps cheaper seems to be the right choice. And monitoring is always a good idea! 🙌

PR #345, opened 2023-11-02
audit small fixes NTRN-136

quasisamurai:
[TASK](https://hadronlabs.atlassian.net/browse/NTRN-136)

[Admin module PR](https://github.com/neutron-org/admin-module/pull/5)

- [x] There is a wrong log message in HandleTimeout (it refers to acknowledgement instead of the timeout): [here](https://github.com/neutron-org/neutron/blob/261f47c30dcfc7cd51eef2b78bd770abd059208b/x/transfer/ibc_handlers.go#L73-L75);

- [x] The entire logic of the [DispatchMessage](https://github.com/neutron-org/neutron/blob/b15a3aa47ec9cb91a4d41557124643d837a70053/wasmbinding/message_plugin.go#L86-L142) function is contained within the if-branch. A more elegant approach would involve placing the opposite condition inside the if statement.

- [x] Inside of the each module’s module.go file, there is a ConsensusVersion() function which returns a sequence number for state-breaking change of the module which should be incremented for each consensus-breaking change introduced by the module, such as [here](https://github.com/neutron-org/neutron/blob/112b268777c3a23a9c66b952b161d469ddffe80d/x/contractmanager/module.go#L160). The current implementation directly returns the consensus version as a hard-coded value, which could potentially lead to inconsistencies if not updated properly during migrations or code updates. We recommend introducing a constant to track the version.

- [x] The current implementation hard-codes the global fees and denom values [here](https://github.com/neutron-org/neutron/blob/4f285beee5b00b94ae9ee03c7e28089db65038f9/app/upgrades/nextupgrade/upgrades.go#L249-L256) and [here](https://github.com/neutron-org/neutron/blob/4f285beee5b00b94ae9ee03c7e28089db65038f9/app/upgrades/nextupgrade/upgrades.go#L289-L290), which might not be optimal for maintainability and flexibility. Consider introducing constant values for fees, and if possible, add denominations dynamically.

- [x] There is a [superfluous assignment](https://github.com/neutron-org/neutron/blob/261f47c30dcfc7cd51eef2b78bd770abd059208b/x/transfer/keeper/keeper.go#L55-L56) of the memo field. After creating the new transferMsg, there is no need to assign it’s memo field again.

- [x] Remove mention of a proposal_id which was not created yet from an error message [here](https://github.com/neutron-org/admin-module/blob/d5cd5ae2230c0b9cb540267acfc4e1cb5f55395b/x/adminmodule/keeper/proposal.go#L13-L13);


pr0n00gler:
Wen test run 😈

quasisamurai:
https://github.com/neutron-org/neutron-tests/actions/runs/6784096134

quasisamurai:
https://github.com/neutron-org/neutron-tests/actions/runs/6784096134

swelf19 (approved):

PR #352, opened 2023-11-07
[NTRN-142] feat: enhance contract failure with error

sotnikov-s:
**task:** https://hadronlabs.atlassian.net/browse/NTRN-142

**This PR:**
- adds `sudo._contract_address`, `sudo.failure_id` and `sudo.error` events emission on failed Sudo calls;
- adds `AddressFailure` method to `contractmanager`'s `Keeper` which returns a single failure by given contract address and failure ID;
- adds `Error` field to `contractmanager`'s `Failure` storage item which contains redacted error occurred on a Sudo call;
- adds `neutrond q contractmanager failure-details [address] [failure-id]` CLI query which returns full error message related to a failure by given contract address and failure ID. The query retrieves the error message from events emitted on failed Sudo.

**How to test this code:**
1. checkout to update-sdk47 related branches for [integration tests](https://github.com/neutron-org/neutron-integration-tests/pull/174) and [ICQ relayer](https://github.com/neutron-org/neutron-query-relayer/pull/80), download dev contracts from [sdk-47 related branch](https://github.com/neutron-org/neutron-dev-contracts/pull/29);
2. build neutron from this branch;
3. start cosmopark and run `interchaintx.test.ts` outside docker:
```bash
make start-cosmopark (from setup folder)
# wait ~20sec for initialisation
NO_DOCKER=1 ./node_modules/jest/bin/jest.js -b src/testcases/run_in_band/interchaintx.test.ts` (from repo root)
```
4. run `neutrond q contractmanager failures` and see 5 stored failures with redacted errors like codespace: wasm, code: 5;
5. run `neutrond q contractmanager failure-details neutron1m0z0kk0qqug74n9u9ul23e28x5fszr628h20xwt6jywjpp64xn4qatgvm0 1` (also for `2`, `3`, `4` and `5` failure ID) and see the original error messages.


x/contractmanager/client/cli/query_failure.go
```diff
@@ -23,19 +26,19 @@ func CmdFailures() *cobra.Command {
...
 			}
 
-			params := &types.QueryFailuresRequest{
+			params := &contractmanagertypes.QueryFailuresRequest{
 				Address:    address,
 				Pagination: pageReq,
 			}
 
-			res, err := queryClient.Failures(context.Background(), params)
+			res, err := queryClient.Failures(cmd.Context(), params)
```
mmikev, line 41: Shouldn't we pass `clientCtx` here?

x/contractmanager/client/cli/query_failure.go
```diff
@@ -49,3 +52,60 @@ func CmdFailures() *cobra.Command {
...
+			}
+
+			clientCtx, err := client.GetClientQueryContext(cmd)
+			if err != nil {
+				return err
+			}
+
+			queryClient := contractmanagertypes.NewQueryClient(clientCtx)
+			if _, err = queryClient.AddressFailure(
+				cmd.Context(),
```
mmikev, line 78: And same here about `clientCtx`

x/contractmanager/ibc_middleware.go
```diff
@@ -34,9 +36,16 @@ func (k SudoLimitWrapper) Sudo(ctx sdk.Context, contractAddress sdk.AccAddress,
 		// maybe later we'll retrieve actual errors from events
 		resp, err = k.WasmKeeper.Sudo(cacheCtx, contractAddress, msg)
 	}()
-	if err != nil {
-		// the contract either returned an error or panicked with `out of gas`
-		k.contractManager.AddContractFailure(ctx, contractAddress.String(), msg)
+	if err != nil { // the contract either returned an error or panicked with `out of gas`
+		failure := k.contractManager.AddContractFailure(ctx, contractAddress.String(), msg, redactError(err).Error())
```
mmikev: Maybe we should redact an error inside '.AddContractFailure' method?
What if someone forget about non determenism and passes a full error here as an argument?

x/contractmanager/client/cli/query_failure.go
sotnikov-s: `clientCtx` is another kind of context as far as I understand. `cmd.Context()` returns the general kind context from golang stdlib whereas `clientCtx` is context in some different terms, and it is not usable as an argument because neither it itself nor one of its fields implement context interface. also it's kind of reasonable to wrap CLI queries into context of cmd execution (`cmd.Context()`) so queries are killed once the parent context (cmd) is killed/interrupted

mmikev: Understandable

x/contractmanager/ibc_middleware.go
sotnikov-s: yeah it kind of makes sense and I've even added this as one of the **Left to be done** points. but I got unsure about it just 30 minutes ago while driving home by taxi: it's useful to have such under-the-hood redaction from the from the sudo wrapper's side, but it will be too much for cases when we have a deterministic error passed to the `AddContractFailure` method, e.g. in the InitGenesis func we have concrete errors passing there and they will be all formatted as `codespace: undefined, code: 1` (the default value for errors returned from `redactError`). I think we should not move `redactError` inside the `AddContractFailure` but expect the error to be deterministic on the input, but we definitely should make a giant warning about that somewhere around the `AddContractFailure` method

sotnikov-s:
https://github.com/neutron-org/neutron-tests/actions/runs/6823402494

Test run agaist the `feat/upd-47` integration tests branch because we decided to postpone emitted events + failure details querying testing until an RPC handler is implemented for these needs. querying failure details via CLI in tests is such an ugly pain in the ass.

The test run checks whether the feature breaks something in the code. For feature testing please refer to the "How to test this code" section and make the test manually.

pr0n00gler:
https://github.com/neutron-org/neutron-tests/actions/runs/6853768496

PR #355, opened 2023-11-07
Fix: small improvements

pr0n00gler:
https://hadronlabs.atlassian.net/browse/NTRN-154

1. Fixed in another PR;
2. Fixed;
3. Fixed;
4. No way to fix that without refactoring of test suite. Suggest to leave it as it is due to a time pressure;
5. Fixed in https://github.com/duality-labs/duality/blob/9a208a52d5b9fcb1b684f590f0d8bdb635a69cbf/x/dex/keeper/core.go#L327
6. This PR https://github.com/neutron-org/neutron/pull/344 will use the swap error, so we leave it unchanged for now.


jcompagni10:
Just for clarification, 5.) is referring to this https://github.com/duality-labs/duality/blob/9a208a52d5b9fcb1b684f590f0d8bdb635a69cbf/x/dex/keeper/core.go#L327
It was resolved as part of an earlier PR.

With regards to 6.) This PR https://github.com/neutron-org/neutron/pull/344 will use the swap error, so I think it's better to leave it unchanged for now.

jcompagni10:
otherwise, LGTM

PR #357, opened 2023-11-08
fix: msg multihop validation #ntrn-157

NeverHappened:
https://hadronlabs.atlassian.net/browse/NTRN-157


jcompagni10 (approved):

x/dex/types/message_multi_hop_swap.go
```diff
@@ -70,6 +70,20 @@ func (msg *MsgMultiHopSwap) ValidateBasic() error {
 		return ErrMissingMultihopRoute
 	}
 
+	for _, route := range msg.Routes {
```
jcompagni10, line 74: A bit of a nitpick, but we're already iterating through the routes below, would be a bit cleaner to combine all checks into a single loop.

jcompagni10 (approved):

jcompagni10 (approved):

PR #378, opened 2023-11-28
v2: release

pr0n00gler:


quasisamurai (approved):
lgtm

app/app.go
```diff
@@ -21,9 +21,9 @@ import (
 	"github.com/neutron-org/neutron/docs"
 
 	"github.com/neutron-org/neutron/app/upgrades"
+	v030 "github.com/neutron-org/neutron/app/upgrades/v0.3.0"
```
sotnikov-s, line 24: a minor suggestion: let's make this one and the rest as `v0_3_0`? otherwise it could be a little pain in the ass in the future when we have releases like 2.0.15 as `2015`. we'd better stick with a more readable pattern for versions

sotnikov-s (approved):
we also need to update pion-1 upgrade docs for validators correspondingly

app/app.go
mmikev: Linter doesn't like this

NeverHappened (approved):
lgtm! you didn't change the old upgrade names, so it should be fine right?just too make sure there is no repeated migrations because of the renames

swelf19 (approved):

PR #385, opened 2023-12-02
change limitOrder expiration to int64

jcompagni10:
In order to make limit order expirationTime available to CW contracts it must be stored as an integer. 

In the previous code the time.time value was overloaded to also express the type of limit order (GTC, JIT, GoodTil). The integer time representation of time cannot be overloaded, so additional metadata about the limit order also has to be stored. 
This is probably a beneficial improvement, but it does make for a larger changeset. 

Note: int64 must be used (instead of uint64) because ctx.BlockTime.Unix() returns an int64. As part of the testing setup this integer value is frequently negative.


proto/neutron/dex/tx.proto
```diff
@@ -105,11 +105,8 @@ message MsgPlaceLimitOrder {
       (gogoproto.jsontag) = "amount_in"
   ];
   LimitOrderType order_type = 8;
-  // expirationTime is only valid iff orderType == GOOD_TIL_TIME.
-  google.protobuf.Timestamp expiration_time = 9 [
-                                             (gogoproto.stdtime) = true,
-                                             (gogoproto.nullable) = true
-                                             ];
+  // Unix time in seconds. Expiration_time is only valid iff orderType == GOOD_TIL_TIME.
+  int64 expiration_time = 9;
```
sotnikov-s, line 109: as a`google.protobuf.Timestamp` it was `(gogoproto.nullable) = true`, but as an `int64` now it's not marked as optional. I also can see that on the CLI side it's optional. so just to make sure, it's treated as optional with the `0` value passed as none, right?

I need to be sure on such things because I'm developing stargate helpers to be used on smart contracts' side and such things should be obvious for users. I think I'll also ask you to review optionality of request/response fields one day :) for now I simply rely on the `gogoproto.nullable` tags and mark the `true` ones as optional, will double check it against cli commands, but as the last check I'll need your help

x/dex/client/cli/tx_place_limit_order.go
```diff
@@ -18,9 +17,9 @@ import (
 func CmdPlaceLimitOrder() *cobra.Command {
 	cmd := &cobra.Command{
 		//nolint:lll
-		Use:     "place-limit-order [receiver] [token-in] [token-out] [tick-index] [amount-in] ?[order-type] ?[expirationTime] ?(--max-amout-out)",
+		Use:     "place-limit-order [receiver] [token-in] [token-out] [tick-index] [amount-in] ?[order-type] ?[expirationTimeUnix] ?(--max-amout-out)",
 		Short:   "Broadcast message PlaceLimitOrder",
-		Example: "place-limit-order alice tokenA tokenB [-10] tokenA 50 GOOD_TIL_TIME '01/02/2006 15:04:05' --max-amount-out 20 --from alice",
+		Example: "place-limit-order alice tokenA tokenB [-10] tokenA 50 GOOD_TIL_TIME 1701429212 --max-amount-out 20 --from alice",
```
sotnikov-s: > place-limit-order alice tokenA tokenB [-10] **tokenA** 50 GOOD_TIL_TIME 1701429212 --max-amount-out 20 --from alice

the second **tokenA** must be a misprint

proto/neutron/dex/tx.proto
sotnikov-s: also: you sure about the decision of using int64 here instead of a uint? you can compare and cast between uint and int. I only don't understand why you use negative timestamps in tests

proto/neutron/dex/limit_order_tranche.proto
```diff
@@ -46,15 +47,13 @@ message LimitOrderTranche {
...
-                                             (gogoproto.nullable) = true
-                                             ];
+  int64 expiration_time = 6;
   string price_taker_to_maker = 7 [
                 (gogoproto.moretags)   = "yaml:\"price_taker_to_maker\"",
                 (gogoproto.customtype) = "github.com/neutron-org/neutron/utils/math.PrecDec",
                 (gogoproto.nullable)   = false,
                 (gogoproto.jsontag) = "price_taker_to_maker"
                 ];
+  LimitOrderType Type = 8;
```
mmikev: Should it be added into the current release or we can wait for the next one?

pr0n00gler (changes requested):
State migrations must be implemented to change already saved structs in the state since:
* `expiration_time` has a new type now;
* `Type` field is added to `LimitOrderTrancheKey` struct

proto/neutron/dex/tx.proto
jcompagni10: This is correct. 0 is treated as a nil value. 

Uint64 would be the more logical choice. But as part of ibc-testing the chain is init-ed with a nil time value(https://github.com/cosmos/interchain-security/blob/512753529b7fea6f825c1bfdb1242dd4207b2366/legacy_ibc_testing/testing/app.go#L149-L154) , which gets converted to "0000-00-00T00:00:00" so the UNIX timestamp used to commit the first block is -62135596800. To avoid ending up with negative values we'd have to either fix IBC-testing or significantly re-write the test fixtures.

x/dex/client/cli/tx_place_limit_order.go
jcompagni10: fixed

proto/neutron/dex/limit_order_tranche.proto
jcompagni10: Unfortunately, switching away from time.time requires we add this additional type data to tell if a limit order is JIT or goodTil. Before we used a zero time value to denote JIT limit orders and a nil value to denote non-expiring limit orders. Since uint64 cannot be nil we need to the extra field to retain this data. 

That being said, I realize this is a fairly large change at this stage. Up to you if it makes sense to merge or figure out a fix on the CW for now. @pr0n00gler

jcompagni10:
> State migrations must be implemented to change already saved structs in the state since:
> 
> * `expiration_time` has a new type now;
> * `Type` field is added to `LimitOrderTrancheKey` struct

If we merge now will be updating the testnet or just restarting fresh?

swelf19:
> If we merge now will be updating the testnet or just restarting fresh?

We already released v2.0.0 and this pr brings breaking changes, so i think upgrade with migration handlers just the right way to do it. Even more, there are already tranches in testnet are waiting to be migrated

jcompagni10:
> > If we merge now will be updating the testnet or just restarting fresh?
> 
> We already released v2.0.0 and this pr brings breaking changes, so i think upgrade with migration handlers just the right way to do it. Even more, there are already tranches in testnet are waiting to be migrated

Got it. Will do

proto/neutron/dex/limit_order_tranche.proto
sotnikov-s: I like the way to distinguish between limit order types by a designated field, not by obscure expiration time values. also: please be careful with field names in proto:

LimitOrderType **Type** = 8; should be **type** or even better **order_type** as in other messages

jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/7106058429

pr0n00gler:
This PR is API breaking and will be merged in the next breaking changes release

PR #390, opened 2023-12-06
Add grpc option `IncludePoolData` to QueryUserDeposits

jcompagni10:
In order to facilitate easier state computation for the duality frontend additional pool data needs to be included in the UserDepositsQuery. See https://github.com/duality-labs/duality-web-app/issues/493 for more details


jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/7647484992

pr0n00gler (dismissed):

proto/neutron/dex/deposit_record.proto
```diff
@@ -18,4 +19,11 @@ message DepositRecord {
   int64 lower_tick_index = 4;
   int64 upper_tick_index = 5;
   uint64 fee = 6;
+  string total_shares = 7 [
+    (gogoproto.moretags) = "yaml:\"total_shares\"",
+    (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Int",
```
swelf19, line 24: `github.com/cosmos/cosmos-sdk/types.Int` is deprecated and it is an alias to `cosmossdk.io/math.Int`. But it's not necessary to fix it right now.  We want to move sometime, but in standalone task

x/dex/client/cli/query_user_deposits.go
```diff
@@ -8,6 +8,7 @@ import (
 	"github.com/neutron-org/neutron/v2/x/dex/types"
 )
 
+// TODO: add optional IncludePoolData arg
```
swelf19: is it hard to do, if not, may be it worth to

x/dex/keeper/grpc_query_user_deposits.go
```diff
@@ -70,3 +74,14 @@ func (k Keeper) UserDepositsAll(
 		Pagination: pageRes,
 	}, nil
 }
+
+func (k Keeper) addPoolData(ctx sdk.Context, record *types.DepositRecord) *types.DepositRecord {
+	pool, found := k.GetPool(ctx, record.PairId, record.CenterTickIndex, record.Fee)
+	if found {
+		record.Pool = pool
+	}
+	supply := k.bankKeeper.GetSupply(ctx, pool.GetPoolDenom())
```
swelf19, line 85: i think it must never happens, where there is no pool found? but still, it looks strange, if no pool found you try to GetPoolDenom from nil pointer. We can change if statement
```
if !found {
return
}
```
otherwise fill the data

x/dex/keeper/grpc_query_user_deposits.go
```diff
@@ -56,6 +56,10 @@ func (k Keeper) UserDepositsAll(
 					Fee:             poolMetadata.Fee,
 				}
 
+				if req.IncludePoolData {
+					depositRecord = k.addPoolData(ctx, depositRecord)
```
swelf19: looks a bit confusing to me, semantic looks like `append` function, may be it because you return the same pointer

swelf19 (commented):
1) pls regenerate openapi
2) neutron-sdk's proto files has to be regenerated https://github.com/neutron-org/neutron-sdk/tree/main/packages/neutron-sdk/src/proto_types

sotnikov-s:
manually regenerated proto for neutron-sdk here: https://github.com/neutron-org/neutron-sdk/pull/127

x/dex/client/cli/query_user_deposits.go
sotnikov-s: must be a piece of cake: just add an argument to the cmd and pass it to the QueryAllUserDepositsRequest struct

x/dex/keeper/grpc_query_user_deposits.go
sotnikov-s: also probably it makes sense to return an error e.g. "pool data not found" to the calls with `req.IncludePoolData=true` in this case. so the users explicitly asked to get pool data get an error if no data found, and not just a response with no pool data. just a suggestion, not sure how to make it the right way from the user's perspective

x/dex/keeper/grpc_query_user_deposits.go
sotnikov-s: you mean that the `depositRecord = ` part could be removed since depositRecord is passed as a pointer and mutations are applied to the record without an additional assignment? if yes, I agree

pr0n00gler (dismissed):

NeverHappened (dismissed):
Please rerun integration tests, otherwise lgtm

swelf19 (dismissed):

jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/7879623104

jcompagni10:
NOTE: openapi.yml has been generated with ignite v28.2.0

NeverHappened (approved):

swelf19 (approved):

PR #400, opened 2023-12-21
NTRN-188 refactor dex module keeper

quasisamurai:
[TASK](https://hadronlabs.atlassian.net/browse/NTRN-188)

Get [poolByID ](https://github.com/neutron-org/neutron/blob/main/x/dex/keeper/pool.go#L99)is currently inefficient.It currently makes an unnecessary call to k.GetPoolIDByParams as part of k.GetPool

This code should be refactored.

I also refactored InitPool & SetPool functions because I can


x/dex/keeper/pool.go
```diff
@@ -30,26 +30,50 @@ func (k Keeper) InitPool(
...
-	poolMetadata := types.PoolMetadata{PairId: pairID, Tick: centerTickIndexNormalized, Fee: fee}
+	poolID := k.initializePoolMetadata(ctx, pairID, centerTickIndexNormalized, fee)
 
-	// Get current poolID
+	k.storePoolID(ctx, poolID, pairID, centerTickIndexNormalized, fee)
+
+	return types.NewPool(pairID, centerTickIndexNormalized, fee, poolID)
+}
+
+func (k Keeper) initializePoolMetadata(
```
jcompagni10: maybe move this to keeper/pool_metadata.go ?

x/dex/keeper/pool.go
```diff
@@ -85,7 +118,6 @@ func (k Keeper) GetPool(
 	case lowerTickFound && !upperTickFound:
 		upperTick = types.NewPoolReservesFromCounterpart(lowerTick)
 	case !lowerTickFound && !upperTickFound:
-		// Pool has already been initialized before so we can safely assume that pool creation doesn't throw an error
```
jcompagni10, line 88: Maybe just personal preference, but I like comments explaining why we chose to use a method that panics. Otherwise, it can look like a mistake to someone later.

x/dex/keeper/pool.go
```diff
@@ -30,26 +30,50 @@ func (k Keeper) InitPool(
...
 
-	// Store poolMetadata
 	k.SetPoolMetadata(ctx, poolMetadata)
 
-	// Create a reference so poolID can be looked up by poolMetadata
+	k.incrementPoolCount(ctx)
+	return poolID
+}
+
+func (k Keeper) storePoolID(
```
jcompagni10: Can we call this storePoolIDRef? Seems a bit more clear.

jcompagni10 (changes requested):

x/dex/keeper/pool.go
```diff
@@ -85,7 +98,7 @@ func (k Keeper) GetPool(
 	case lowerTickFound && !upperTickFound:
 		upperTick = types.NewPoolReservesFromCounterpart(lowerTick)
 	case !lowerTickFound && !upperTickFound:
-		// Pool has already been initialized before so we can safely assume that pool creation doesn't throw an error
+		// Pool has already been initialized before, пso we can safely assume that pool creation doesn't throw an error
```
mmikev:
```suggestion
		// Pool has already been initialized before, so we can safely assume that pool creation doesn't throw an error
```

pr0n00gler (dismissed):

pr0n00gler (approved):

jcompagni10 (approved):

PR #438, opened 2024-02-15
feat: return dust for multihop swap to sender #ntrn-187

NeverHappened:
send all dust collected on multihop route to the receiver

https://hadronlabs.atlassian.net/browse/NTRN-187


x/dex/keeper/integration_multihopswap_test.go
```diff
@@ -70,6 +71,90 @@ func (s *DexTestSuite) TestMultiHopSwapSingleRoute() {
...
+
+	s.assertDexBalanceWithDenomInt("TokenA", math.NewInt(59_841))
+	s.assertDexBalanceWithDenomInt("TokenB", math.NewInt(1_000_000))
+	s.assertDexBalanceWithDenomInt("TokenC", math.NewInt(1_000_000))
+	s.assertDexBalanceWithDenomInt("TokenD", math.NewInt(999_999))
+}
+
+// this is a case where Pool.Swap() (B <-> C) gives up not the full available tokens amount and leaves 1,
+// and because of that swap fails overall (1 token * 403 price > 2 allowed for order to be filled)
+func (s *DexTestSuite) TestBugWithLiquidityCheck() {
```
NeverHappened: Hey @NicholasDotSol @jcompagni10 

This test demonstrates behavior that I'm not sure is correct.
So we setup pools, such that:
- A is much cheaper than B
- B is much more expensive than C

In this scenario with multihop swap, on a step with B to C exchange we have:
1,002TokenB and try to exchange it to TokenC.
Inside the `Pool.Swap()` it will calculate that we swap 1,001TokenB to 22,015,000C. As you see that's one token less that we wanted to swap.
This have consequnces:
- We have different real price of TokenB to TokenC that we could have (and should?)
- We leave 1TokenB, and in multihop swap scenario it will try to swap it using next pool (because the condition for orderFilled=true is not satisfied). In this test however there is no more pools left, so it just fails the multihop with error "Insufficient liquidity" that is definitely incorrect - pool had enough liquidity.
So question is: what do you think - is it correct behavior? What can we do with multihop scenario?

jcompagni10: Will look more closely in a bit. But At first glance if we try to swap 60_000_000TokenA we should get out 1002 tokenB but the pool is only funded with 1TokenB

x/dex/keeper/multihop_swap.go
```diff
@@ -98,69 +99,83 @@ func (k Keeper) RunMultihopRoute(
...
 	if err != nil {
-		return sdk.Coin{}, err
+		return sdk.Coin{}, sdk.Coin{}, err
 	}
 	if !orderFilled {
-		return sdk.Coin{}, types.ErrInsufficientLiquidity
+		return sdk.Coin{}, sdk.Coin{}, types.ErrInsufficientLiquidity
+	}
+
+	dust = sdk.Coin.Sub(sdk.NewCoin(swapAmountTakerDenom.Denom, amountIn), swapAmountTakerDenom)
```
mmikev, line 177: Why do we need to do a substraction here?
Isn't `swapAmountTakerDenom` already a dust amount?

x/dex/keeper/integration_multihopswap_test.go
```diff
@@ -70,6 +70,83 @@ func (s *DexTestSuite) TestMultiHopSwapSingleRoute() {
 	s.assertDexBalanceWithDenom("TokenD", 0)
 }
 
+func (s *DexTestSuite) TestMultiHopSwapSingleRouteWithDust() {
```
mmikev, line 73: Let's add a testcase where Alice receives not only TokenA dust, but also dust of other tokens

pr0n00gler (changes requested):

x/dex/keeper/multihop_swap.go
NeverHappened: I think no, [here](https://github.com/neutron-org/neutron/blob/0085ad29a76a11fd5d124b774e1b06b64e716cc0/x/dex/keeper/liquidity.go#L89) you can see it returns INPUT - DUST

x/dex/keeper/integration_multihopswap_test.go
NeverHappened: There is such a testcase just below (TestMultiHopSwapSingleRouteWithManyDustTokens)

x/dex/keeper/multihop_swap.go
mmikev: Ah, so basically it's the amount that **was** swapped

mmikev: Not quite good naming and comment coverage honestly

x/dex/keeper/integration_multihopswap_test.go
mmikev: Ah, didn't see that

x/dex/keeper/core.go
```diff
@@ -289,6 +292,21 @@ func (k Keeper) MultiHopSwapCore(
...
+		err = k.bankKeeper.SendCoinsFromModuleToAccount(
+			ctx,
+			types.ModuleName,
+			receiverAddr,
+			bestRoute.dust,
+		)
+		if err != nil {
+			return sdk.Coin{}, fmt.Errorf("failed to send dust back to user: %w", err)
+		}
+	}
```
sotnikov-s: we can optimise the transfer by sending both coinOut and dust in a single operation:
```go
err = k.bankKeeper.SendCoinsFromModuleToAccount(
	ctx,
	types.ModuleName,
	receiverAddr,
	append(bestRoute.dust, bestRoute.coinOut), <<<<<<<<<<<<<
)
```

jcompagni10 (dismissed):
Lgtm

x/dex/keeper/multihop_swap.go
```diff
@@ -98,69 +101,83 @@ func (k Keeper) RunMultihopRoute(
...
+) (sdk.Coins, sdk.Coin, func(), error) {
 	routeData, err := k.HopsToRouteData(ctx, route.Hops)
 	if err != nil {
-		return sdk.Coin{}, nil, err
+		return sdk.Coins{}, sdk.Coin{}, nil, err
 	}
 	currentPrice := math_utils.OnePrecDec()
 
 	var currentOutCoin sdk.Coin
+	var currentDust sdk.Coin
```
swelf19: A small suggestion, rename all `current` to `step`, i think the `step` name  fits better for variables usage.

swelf19 (dismissed):

pr0n00gler (dismissed):

swelf19 (dismissed):

x/dex/keeper/integration_multihopswap_test.go
```diff
@@ -70,6 +70,83 @@ func (s *DexTestSuite) TestMultiHopSwapSingleRoute() {
...
+	_, err := s.msgServer.MultiHopSwap(s.GoCtx, msg)
+	s.Assert().Nil(err)
+
+	// THEN alice gets out 1 TokenD
+
+	// 2_000_000_000 - 600_000_000 (swap in) + 1_468_096 (dust)
+	s.assertAccountBalanceWithDenomInt(s.alice, "TokenA", math.NewInt(1_400_031_907)) // alice balance - spent + received dust
+	s.assertAccountBalanceWithDenomInt(s.alice, "TokenB", math.NewInt(347))
+	s.assertAccountBalanceWithDenomInt(s.alice, "TokenC", math.NewInt(0))
+	s.assertAccountBalanceWithDenomInt(s.alice, "TokenD", math.NewInt(24)) // received TokenD after swap
```
sotnikov-s, line 142: I wonder if it's good in terms of UX — to have TokenB received when making a TokenA -> TokenD swap. have you guys considered making backwards swaps for intermediate token dust? e.g. in this example we could swap all the TokenB dust back to TokenA and as a result have only TokenA and TokenD on the user's wallet

x/dex/keeper/integration_multihopswap_test.go
```diff
@@ -273,7 +350,7 @@ func (s *DexTestSuite) TestMultiHopSwapMultiRouteFindBestRoute() {
 
 	// THEN swap succeeds through route A<>B, B<>E, E<>X
 
-	s.assertAccountBalanceWithDenom(s.alice, "TokenA", 0)
+	s.assertAccountBalanceWithDenomInt(s.alice, "TokenA", math.NewInt(2))
```
sotnikov-s:
```suggestion
	s.assertAccountBalanceWithDenomInt(s.alice, "TokenA", math.NewInt(2)) // dust left
```

x/dex/keeper/integration_multihopswap_test.go
NeverHappened: Interesting suggestion, not sure if needed.
Cons is that there is no guarantee we can swap back until the TokenA. 
Also can be a little taxing on the efficiency?

x/dex/keeper/integration_multihopswap_test.go
NeverHappened: Oh this was just mistake in commenting. Indeed if you swap 60000000 you should get 1002 tokenB.

sotnikov-s (dismissed):

swelf19 (dismissed):

pr0n00gler (dismissed):

PR #443, opened 2024-02-20
fix: make sure order is fulfilled on 1 token left

NeverHappened:
We can have few potential issues with swap now:
- one token left and swap does not get fulfilled (or swaps from the other pool), even though we have needed liquidity;
- user needs to swap 1 very expensive token but cannot do it;
- we can have unfair price error (need to check if this problem exists).


x/dex/keeper/liquidity.go
```diff
@@ -97,6 +96,7 @@ func (k Keeper) Swap(
 		), orderFilled, nil
 }
 
+// what's the point of this function if we just write cache every time?
 func (k Keeper) SwapWithCache(
```
NeverHappened: what's the point of this function if we just write cache every time?

jcompagni10: Need to confirm but I I think there might be some minor gas savings if we traverse pools multiple times. But it’s largely unnecessary at this point

x/dex/types/pool_test.go
```diff
@@ -207,3 +207,33 @@ func TestCalcGreatestMatchingRatio1SidedPool1SidedToken1B(t *testing.T) {
 	assert.Equal(t, math.NewInt(0), trueAmount0)
 	assert.Equal(t, math.NewInt(10), trueAmount1)
 }
+
+// This failing test shows that if we swap 1 to many tokens,
+// we'll get 0A to xB tokens in result.
+// This is intended behaviour because of the rounding leftovers
+func TestSwapRoundingWithExpensiveToCheap(t *testing.T) {
```
jcompagni10, line 214: Can we add either a liquidity test or a  limit order integration test with these same numbers to ensure that this sort of order will not actually succeed

jcompagni10 (changes requested):

x/dex/keeper/liquidity.go
```diff
@@ -43,20 +42,12 @@ func (k Keeper) Swap(
 
 		inAmount, outAmount := liq.Swap(remainingTakerDenom, remainingMakerDenom)
 
-		// If (due to rounding) the actual price given to the maker is demonstrably unfair
-		// we do not save the results of the swap and we exit.
-		// While the decrease in price quality for the maker is semi-linear with the amount
-		// being swapped, it is possible that the next swap could yield a "fair" price.
-		// Nonethless, once the remainingTakerDenom gets small enough to start causing unfair swaps
-		// it is much simpler to just abort.
-		if inAmount.IsZero() || isUnfairTruePrice(params.MaxTrueTakerSpread, inAmount, outAmount, liq) {
```
NeverHappened, line 52: Do we also want to delete `inAmount.IsZero()` check? Does it allow for possibility of zero amount exchanges?

jcompagni10:
Let's pause on merging this. I am working on a slightly larger fix that will also resolve the issues here.

pr0n00gler:
Replaced by https://github.com/neutron-org/neutron/pull/479

PR #448, opened 2024-02-26
release: v3 preparation

swelf19:


swelf19:
https://github.com/neutron-org/neutron-tests/actions/runs/8080473232

app/upgrades/v2.0.2/constants.go
```diff
@@ -1,7 +1,7 @@
 package v202
 
 import (
-	"github.com/neutron-org/neutron/v2/app/upgrades"
+	"github.com/neutron-org/neutron/v3/app/upgrades"
```
mmikev, line 4: i don't think we need to change package version in v2 migration code

pr0n00gler (approved):

PR #458, opened 2024-03-06
️fix: load genesis path from config when executue newApp

### LINKED ISSUE #457: There is a problem in the process of loading ChainID from genesis in newApp.
There is a problem in the process of loading ChainID from genesis in newApp.

SunSet0418:
This PR modifies the path to load the Genesis file to obtain the path from the configuration file to set chain-id when running newApp.

close #457

### DIFF
--- a/cmd/neutrond/root.go
+++ b/cmd/neutrond/root.go
@@ -234,7 +234,8 @@ func (ac appCreator) newApp(
 	chainID := cast.ToString(appOpts.Get(flags.FlagChainID))
 	if chainID == "" {
 		// fallback to genesis chain-id
-		appGenesis, err := tmtypes.GenesisDocFromFile(filepath.Join(homeDir, "config", "genesis.json"))
+		genDocFile := filepath.Join(homeDir, cast.ToString(appOpts.Get("genesis_file")))
+		appGenesis, err := tmtypes.GenesisDocFromFile(genDocFile)
 		if err != nil {
 			panic(err)
 		}
@@ -243,7 +244,7 @@ func (ac appCreator) newApp(
 	}
 
 	return app.New(logger, db, traceStore, true, skipUpgradeHeights,
-		cast.ToString(appOpts.Get(flags.FlagHome)),
+		homeDir,
 		cast.ToUint(appOpts.Get(server.FlagInvCheckPeriod)),
 		ac.encCfg,
 		appOpts,


pr0n00gler:
Could you please describe a problem a bit more?
What case the PR solves exactly?

SunSet0418:
@pr0n00gler 
The genesis path included in the configuration file and the genesis path hard-coded in the code are different.

First, in the process of executing newApp, chain-id is read from hard-coded genesis. There is already a genesis path in the configuration file, so there is no need to hard-code the genesis path.

Furthermore, when the node is executed after newApp, genesis is read from the genesis path in the configuration file.

Isn't it strange that one app references different file paths containing the same data?

sotnikov-s:
> @pr0n00gler The genesis path included in the configuration file and the genesis path hard-coded in the code are different.
> 
> First, in the process of executing newApp, chain-id is read from hard-coded genesis. There is already a genesis path in the configuration file, so there is no need to hard-code the genesis path.
> 
> Furthermore, when the node is executed after newApp, genesis is read from the genesis path in the configuration file.
> 
> Isn't it strange that one app references different file paths containing the same data?

Hello! This is a bit complicated to understand what you mean. I'd be grateful if you could make the following, it would really help me grasp the problem:
1. describe the issue (better on the issue page you linked) using a common issue report flow: what you expected, what you got, how to reproduce
2. add links to the code where those incorrect operations take place

thanks!

SunSet0418:
As you all know, we can set the genesis_file path through config.toml.
```
# Path to the JSON file containing the initial validator set and other meta data
genesis_file = "genesis/mainnet.json"
```
However, in this code the Genesis file is loaded only from the fixed `$DAEMON_HOME/config/genesis.json` path.
https://github.com/neutron-org/neutron/blob/4d29321b8ab10d00782e82dec3267417dd7a40e0/cmd/neutrond/root.go#L237

If I set the `genesis_file` path in config.toml to `$DAEMON_HOME/genesis/mainnet.json` instead of `$DAEMON_HOME/config/genesis.json`, neutrond cannot be run.

So, we changed it to use the `genesis_file` path loaded from config.toml.

SunSet0418:
Hello :) @pr0n00gler @pr0n00gler 

As you all know, we can set the `genesis_file` path through config.toml.
```
# Path to the JSON file containing the initial validator set and other meta data
genesis_file = "genesis/mainnet.json"
```
However, in this code the Genesis file is loaded only from the fixed `$DAEMON_HOME/config/genesis.json` path.
https://github.com/neutron-org/neutron/blob/4d29321b8ab10d00782e82dec3267417dd7a40e0/cmd/neutrond/root.go#L237

If I set the `genesis_file` path in config.toml to `$DAEMON_HOME/genesis/mainnet.json` instead of `$DAEMON_HOME/config/genesis.json`, neutrond cannot be run.
스크린샷 2024-05-02 오후 2 44 13

So, we changed it to use the `genesis_file` path loaded from config.toml.

SunSet0418:
The same PR was uploaded to gaia, merged, and reflected in v16.
https://github.com/cosmos/gaia/pull/2971

pr0n00gler (approved):

PR #463, opened 2024-03-12
Feat/add calc withdraw

jcompagni10:
Add calculate withdrawable amount to QueryUserLimitOrderTranche

This is needed so that CW contracts can check how much they can withdraw from a limit order. Specifically needed for astroport integration.


jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/8254660656

NeverHappened (commented):
integration tests failed. please use sdk-47 branch for neutron-dao

x/dex/keeper/grpc_query_limit_order_tranche_user.go
```diff
@@ -46,14 +47,34 @@ func (k Keeper) LimitOrderTrancheUserAll(
...
+			TradePairId:           tradePairID,
+			TickIndexTakerToMaker: tickIndex,
+			TrancheKey:            trancheUser.TrancheKey,
+		},
+	)
+
+	if !found {
+		return math.ZeroInt(), status.Error(codes.NotFound, "Tranche not found")
+	}
+	withdrawableShares, _ := tranche.Withdraw(&trancheUser)
```
NeverHappened: How about small refactoring in tranche? we can make new function tranche#Withdrawable (or some kind of this), that will only return stuff and do not modify data. And tranche#Withdraw will use this function and just change `t.ReservesTakerDenom = reservesTokenOutDec.Sub(amountOutTokenOut).TruncateInt()`.
That way there is some boundaries between read and write functions.

proto/neutron/dex/query.proto
```diff
@@ -133,10 +133,17 @@ message QueryParamsResponse {
 message QueryGetLimitOrderTrancheUserRequest {
   string address = 1;
   string tranche_key = 2;
+  bool calc_withdrawable_shares = 3;
```
mmikev, line 136: i suppose this is not API breaking cause it's just a fiiled addition, but could you confirm pls?

x/dex/client/cli/query_limit_order_tranche_user.go
```diff
@@ -45,18 +45,23 @@ func CmdListLimitOrderTrancheUser() *cobra.Command {
 
 func CmdShowLimitOrderTrancheUser() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:     "show-limit-order-tranche-user [address] [tranche-key]",
+		Use:     "show-limit-order-tranche-user [address] [tranche-key] ?(--calc-withdraw)",
 		Short:   "shows a LimitOrderTrancheUser",
-		Example: "show-limit-order-tranche-user TRANCHEKEY123 alice",
+		Example: "show-limit-order-tranche-user alice TRANCHEKEY123",
```
mmikev: I think it's better to use valid neutron address as an example instead of `alice`

proto/neutron/dex/query.proto
jcompagni10: Correct. If a value is not supplied it will default to false and the pre-existing behavior will be maintained

jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/8473082006

pr0n00gler (approved):

PR #472, opened 2024-03-27
Feat/add price to limit order

jcompagni10:
Allow limit orders to be placed by specifying a price instead of a tick
Tests still breaking because of pending rounding changes. 

neutron-sdk changes: https://github.com/neutron-org/neutron-sdk/pull/143
integration-test changes: https://github.com/neutron-org/neutron-integration-tests/pull/294


jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/8702351288

x/dex/types/price.go
```diff
@@ -1,14 +1,20 @@
...
 )
 
 // NOTE: 559_680 is the highest possible tick at which price can be calculated with a < 1% error
 // when using 26 digit decimal precision (via prec_dec).
 // The error rate for very negative ticks approaches zero, so there is no concern there
-const MaxTickExp uint64 = 559_680
+const (
+	MaxTickExp uint64 = 559_680
+	MinPrice   string = "0.000000000000000000000000495"
+	MaxPrice   string = "2020125331305056766452345.127500016657360222036663651"
```
mmikev, line 17: How was this number derived?

x/dex/keeper/msg_server.go
```diff
@@ -103,12 +103,19 @@ func (k MsgServer) PlaceLimitOrder(
 	if err != nil {
 		return &types.MsgPlaceLimitOrderResponse{}, err
 	}
+	tickIndex := msg.TickIndexInToOut
+	if msg.PriceInToOut != nil {
+		tickIndex, err = types.CalcTickIndexFromPrice(*msg.PriceInToOut)
```
mmikev: Maybe it's better to allow using of only one field? Either `.TickIndexInToOut` or `.PriceInToOut`?

proto/neutron/dex/tx.proto
```diff
@@ -92,11 +92,14 @@ enum LimitOrderType {
 }
 
 message MsgPlaceLimitOrder {
+  // TODO: fix wire numbers
```
mmikev: What exactly do you mean by todo?

pr0n00gler (changes requested):

x/dex/types/price.go
jcompagni10: this is just the price of 1.0001^559,680 and 1.0001^-559,680

proto/neutron/dex/tx.proto
jcompagni10: Just a note to make all the wire numbers sequential at some point. Very low priority

x/dex/keeper/msg_server.go
jcompagni10: This is enforced in validate basic. Since the plan is to get rid of TickIndexInToOut it doesn't make sense to use a proto ONEOF field

pr0n00gler:
there are a couple of conflicts 🙈

proto/neutron/dex/tx.proto
```diff
@@ -115,6 +118,12 @@ message MsgPlaceLimitOrder {
     (gogoproto.nullable) = true,
     (gogoproto.jsontag) = "max_amount_out"
   ];
+  string price_in_to_out = 11 [
```
mmikev: Also please:
1. add `price_in_to_out` field to `MsgPlaceLimitOrder` in our [wasm bindings](https://github.com/neutron-org/neutron/blob/6364e0025538acc1efc82d7197dc2a2da373fa1b/wasmbinding/bindings/msg.go#L263).
2. reflect field addition in Neutron CW SDK [itself](https://github.com/neutron-org/neutron-sdk/blob/4fc7cfd63b685753d62e64125bc30e7512a911b1/packages/neutron-sdk/src/bindings/dex/msg.rs#L51).
3. and you will need to regenerate protobufs there to make stargate msgs up to date like i did it [here](https://github.com/neutron-org/neutron-sdk/pull/139).
4. And reflect changes in neutron-dev-contracts (like it was done [here](https://github.com/neutron-org/neutron-dev-contracts/pull/45))
5. And restart integration tests to see that everything is ok with bindings and stargate

pr0n00gler (changes requested):

x/dex/utils/math.go
```diff
@@ -66,3 +66,25 @@ func MustSafeUint64ToInt64(in uint64) (out int64) {
 
 	return safeInt64
 }
+
+func Log(val, base math_utils.PrecDec) (math_utils.PrecDec, error) {
+	valueFloat64, err := val.Float64()
```
swelf19: 1) how is it even possible, we fail to convert valid PrecDec to float64, i think val.MustFloat64() is more suitable here
2) It's a bit out of scope, is it ok we loose some occurace by converting PrecDec -> f64

x/dex/types/price.go
```diff
@@ -1,14 +1,20 @@
...
 
 import (
+	sdkerrors "cosmossdk.io/errors"
+
 	math_utils "github.com/neutron-org/neutron/v3/utils/math"
 	"github.com/neutron-org/neutron/v3/x/dex/utils"
 )
 
 // NOTE: 559_680 is the highest possible tick at which price can be calculated with a < 1% error
 // when using 26 digit decimal precision (via prec_dec).
```
swelf19:
```suggestion
// when using 27 digit decimal precision (via prec_dec).
```

x/dex/types/price_test.go
```diff
@@ -1,11 +1,73 @@
...
+		{
+			desc: "GT MaxTickExp",
+			tick: int64(types.MaxTickExp) + 1,
+		},
+		{
+			desc: "LT TickExp",
+			tick: -1*int64(types.MaxTickExp) - 1,
+		},
+	} {
+		t.Run(tc.desc, func(t *testing.T) {
```
swelf19, line 62: 1) Could you pls describe it a little, what we testing here.
2) I can see `types.CalcPrice` returns only `types.ErrTickOutsideRange` error, but, in case it changes in future, may be it's safier to change `else` to `esle if err1==nil`, cause it's not obvious from the test, `types.ErrTickOutsideRange` the onlt possible error. Or just add some comments pls.

x/dex/client/cli/tx_place_limit_order.go
```diff
@@ -65,16 +67,29 @@ func CmdPlaceLimitOrder() *cobra.Command {
...
-				maxAmountOutIntP = &maxAmountOutInt
+			}
+
+			priceArg, err := cmd.Flags().GetString(FlagPrice)
+			if err != nil {
+				return err
+			}
+
+			var priceDec math_utils.PrecDec
+			if maxAmountOutArg != "" {
```
swelf19: You want to test `priceArg`, right?

x/dex/types/message_place_limit_order_test.go
```diff
@@ -6,13 +6,20 @@ import (
...
+	math_utils "github.com/neutron-org/neutron/v3/utils/math"
+
 	"github.com/neutron-org/neutron/v3/testutil/common/sample"
 	dextypes "github.com/neutron-org/neutron/v3/x/dex/types"
 )
 
 func TestMsgPlaceLimitOrder_ValidateBasic(t *testing.T) {
 	ZEROINT := math.ZeroInt()
 	ONEINT := math.OneInt()
+	FIVEDEC := math_utils.NewPrecDec(6)
```
swelf19: are you sure its five

x/dex/types/message_place_limit_order_test.go
```diff
@@ -109,7 +116,44 @@ func TestMsgPlaceLimitOrder_ValidateBasic(t *testing.T) {
...
+				Receiver:       sample.AccAddress(),
+				TokenIn:        "TokenA",
+				TokenOut:       "TokenB",
+				LimitSellPrice: &HUGEDEC,
+				AmountIn:       math.OneInt(),
+			},
+			err: dextypes.ErrPriceOutsideRange,
+		},
+		{
+			name: "price > maxPrice",
```
swelf19: description and error do match
"price > maxPrice" vs err: dextypes.ErrInvalidPriceAndTick

x/dex/types/message_place_limit_order_test.go
```diff
@@ -119,6 +163,40 @@ func TestMsgPlaceLimitOrder_ValidateBasic(t *testing.T) {
 				AmountIn:         math.OneInt(),
 			},
 		},
+
+		{
+			name: "valid msg price",
```
swelf19: "valid msg price" and "valid msg price > 1" have the same test arguments, is it intended to be?

wasmbinding/bindings/msg.go
```diff
@@ -261,13 +261,18 @@ type Dex struct {
...
+	Creator  string `json:"creator,omitempty"`
+	Receiver string `json:"receiver,omitempty"`
+	TokenIn  string `json:"token_in,omitempty"`
+	TokenOut string `json:"token_out,omitempty"`
+	// Deprecated: tick_index_in_to_out will be removed in future release; limit_sell_price should be used instead.
+	TickIndexInToOut int64  `json:"tick_index_in_to_out,omitempty"`
+	LimitSellPrice   string `json:"limit_sell_price,omitempty"`
+	// LimitSellPriceExp can be used to shift LimitSellPrice to a smaller decimal by an order of 10^LimitSellPriceExp
+	// ie. LimitSellPrice: 1.123 && LimitSellPriceExp: 5 => 1.123 / (10^5) = 0.00001123
+	LimitSellPriceExp uint64   `json:"limit_sell_price_exp,omitempty"`
```
swelf19: What the aim of the field, why cant you set desired price with a single `LimitSellPrice` instead of `LimitSellPrice`+`LimitSellPriceExp`. Why the field introduced only for wasm bindings?

swelf19 (dismissed):

x/dex/client/cli/tx_place_limit_order.go
```diff
@@ -91,7 +106,8 @@ func CmdPlaceLimitOrder() *cobra.Command {
 				amountInInt,
 				orderType,
 				goodTil,
-				maxAmountOutIntP,
+				&maxAmountOutInt,
+				&priceDec,
```
swelf19: since you declare variables as `var priceDec math_utils.PrecDec` and `var maxAmountOutInt math.Int` the arguments never nil, does not it brake some logic inside PlaceOrder handler?

x/dex/utils/math.go
jcompagni10: 1. Given that I am writing this as a general method it is possible for a preDec to overflow a float64. I prefer to keep it as is. 
2. Generally speaking yes, but for the specific case I have written a bunch of tests to ensure that within ticklimits there is not loss precision

x/dex/types/price_test.go
jcompagni10: I've added additional comments to the test

x/dex/types/message_place_limit_order_test.go
jcompagni10: haha oops

jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/9025686170

x/dex/utils/math.go
swelf19: > 1. Given that I am writing this as a general method it is possible for a preDec to overflow a float64. I prefer to keep it as is.

Actually it can not, float64 a way wider range. It silently truncates original number keeps only precise number of digits (and loose precision), but it does not throw any error. That the reason why i suggest MustFloat64 (not silently ignore an error), it saves you a couple of lines of unnecessary error checks. But it's your decision

x/dex/utils/math.go
```diff
@@ -66,3 +67,58 @@ func MustSafeUint64ToInt64(in uint64) (out int64) {
...
+	log := logValue / logBase
+
+	logAsPrecDec, err := math_utils.NewPrecDecFromStr(strconv.FormatFloat(log, 'f', -1, 64))
+	if err != nil {
+		return math_utils.ZeroPrecDec(), fmt.Errorf("error converting float64 value to string: %v", err)
+	}
+	return logAsPrecDec, nil
+}
+
+var scientificNotationRE = regexp.MustCompile(`^([\d\.]+)(E([\-\+])(\d+))?$`)
```
swelf19: what's about small 'e', it's pretty common case i think.
And second pretty common case, you can omit '+' sign for positive power.

x/dex/utils/math_test.go
```diff
@@ -0,0 +1,99 @@
...
+			in:   "0.1E-0",
+			out:  math_utils.MustNewPrecDecFromStr("0.1"),
+		},
+		{
+			desc: "valid decimal with E-1",
+			in:   "189.3E-1",
+			out:  math_utils.MustNewPrecDecFromStr("18.93"),
+		},
+		{
+			desc: "valid decimal with E+10",
```
swelf19:
```suggestion
			desc: "valid decimal with E-10",
```

wasmbinding/message_plugin.go
```diff
@@ -246,6 +248,15 @@ func (m *CustomMessenger) dispatchDexMsg(ctx sdk.Context, contractAddr sdk.AccAd
 			t := time.Unix(int64(*(dex.PlaceLimitOrder.ExpirationTime)), 0)
 			msg.ExpirationTime = &t
 		}
+
+		if limitPriceStr := dex.PlaceLimitOrder.LimitSellPrice; limitPriceStr != "" {
+			limitPriceDec, err := dexutils.ParsePrecDecScientificNotation(limitPriceStr)
+			if err != nil {
+				return nil, nil, err
```
sotnikov-s: please add some description to the error e.g. by using errors.Wrap/errors.Wrapf

x/dex/keeper/msg_server.go
```diff
@@ -116,12 +116,19 @@ func (k MsgServer) PlaceLimitOrder(
 	if err != nil {
 		return &types.MsgPlaceLimitOrderResponse{}, err
 	}
+	tickIndex := msg.TickIndexInToOut
+	if msg.LimitSellPrice != nil {
+		tickIndex, err = types.CalcTickIndexFromPrice(*msg.LimitSellPrice)
+		if err != nil {
+			return &types.MsgPlaceLimitOrderResponse{}, err
```
sotnikov-s: please add some description to the error e.g. by using errors.Wrap/errors.Wrapf

x/dex/types/errors.go
```diff
@@ -189,4 +189,21 @@ var (
...
+		ModuleName,
+		1156,
+		"Invalid price; 0.00000000000000000000000050 < PRICE > 2020125331305056766451886.728",
+	)
+
+	ErrInvalidPriceAndTick = sdkerrors.Register(
+		ModuleName,
+		1157,
+		"Cannot specify LimitSellPrice and TickIndexInToOut",
+	)
```
sotnikov-s: I find the error message confusing. What do you think about rephrasing it e.g. like
```
Only LimitSellPrice or TickIndexInToOut should be specified
```
or
```
Cannot specify both LimitSellPrice and TickIndexInToOut
```

x/dex/types/price.go
```diff
@@ -1,14 +1,20 @@
...
 // NOTE: 559_680 is the highest possible tick at which price can be calculated with a < 1% error
-// when using 26 digit decimal precision (via prec_dec).
+// when using 27 digit decimal precision (via prec_dec).
 // The error rate for very negative ticks approaches zero, so there is no concern there
-const MaxTickExp uint64 = 559_680
+const (
+	MaxTickExp uint64 = 559_680
+	MinPrice   string = "0.000000000000000000000000495"
+	MaxPrice   string = "2020125331305056766452345.127500016657360222036663651"
+)
```
sotnikov-s: please move the comment inside the const brackets just above the MaxTickExp definition? otherwise the comment is applied to all the consts defined in the block:
image

should be like that:
image

x/dex/types/price.go
```diff
@@ -24,6 +30,29 @@ func CalcPrice(relativeTickIndex int64) (math_utils.PrecDec, error) {
 	return math_utils.OnePrecDec().Quo(utils.BasePrice().Power(uint64(relativeTickIndex))), nil
 }
 
+func CalcTickIndexFromPrice(price math_utils.PrecDec) (int64, error) {
+	if IsPriceOutOfRange(price) {
+		return 0, ErrPriceOutsideRange
+	}
```
sotnikov-s, line 37: this check is already done in the Validate method. can be removed I guess?

jcompagni10: I prefer both. It feels like a problem waiting to happen if we make code changes in the future and call this function from elsewhere and the check doesn't exist.

x/dex/utils/math_test.go
```diff
@@ -0,0 +1,109 @@
...
+			out:  math_utils.MustNewPrecDecFromStr("18.93"),
+		},
+		{
+			desc: "valid decimal with E-10",
+			in:   "101234567890.1E-10",
+			out:  math_utils.MustNewPrecDecFromStr("10.12345678901"),
+		},
+		{
+			desc: "valid with lowercase 'e'",
+			in:   "10.1E-2",
```
swelf19:
```suggestion
			in:   "10.1e-2",
```

swelf19 (approved):

sotnikov-s (approved):
LGTM. please resolve conflicts and it seems to be good to go

jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/9280203744

pr0n00gler:
Integration tests are not passing due to some linter error in dev-contracts, please fix them

sotnikov-s (approved):
please attach a successful test run

jcompagni10:
> please attach a successful test run

https://github.com/neutron-org/neutron-tests/actions/runs/9280203744/attempts/4

jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/9280203744

pr0n00gler (approved):

PR #473, opened 2024-03-28
fix: remove default fee in tick update event

dib542:
The default fee attribute here is incorrectly set (it assumes the value of the `TickIndex` which may not be a valid fee at all).

It is better to be removed from the base event and only optionally added on line 208: this helps us distinguish between an intentionally set Fee of `0` and no fee set at all (eg. the tick is reserves in a tranche without a fee as used in line 221).


jcompagni10 (approved):
lgtm

pr0n00gler (approved):

PR #474, opened 2024-04-01
feat: slinky #ntrn-250

NeverHappened:
https://hadronlabs.atlassian.net/browse/NTRN-250

Slinky integration (includes two modules oracle && marketmap)
More docs here: https://docs.skip.money/slinky/overview/

PRs:
- Neutron: https://github.com/neutron-org/neutron/pull/474
  Modules and keepers integration into app
- Integration-tests: https://github.com/neutron-org/neutron-integration-tests/pull/281
  Basic test for querying the prices and oracle sidecar service integration into tests
- Neutronjsplus: https://github.com/neutron-org/neutronjsplus/pull/28
  Generated proto types, added price queries for oracle, proposal to adding marketmap (add currency pairs)


.tool-versions
```diff
@@ -1 +1 @@
-golang 1.21.6
+golang 1.21.1
```
mmikev: Should be 1.22.x, no?

app/app.go
```diff
@@ -269,6 +294,9 @@ func init() {
 	}
 
 	DefaultNodeHome = filepath.Join(userHomeDir, "."+Name)
+
+	// TODO: this is a hack to make unit tests pass. Remove that after marketmap module fixes DefaultParams() call
+	marketmaptypes.DefaultMarketAuthority = "neutron1hxskfdxpp5hqgtjj6am6nkjefhfzj359x0ar3z"
```
mmikev: Please, remove the string literal and replace it with a derivation from a module name

app/app.go
```diff
@@ -716,6 +751,22 @@ func New(
...
+		authtypes.NewModuleAddress(adminmoduletypes.ModuleName),
+	)
+	marketmapModule := marketmap.NewAppModule(appCodec, app.MarketMapKeeper)
+
+	oracleKeeper := oraclekeeper.NewKeeper(runtime.NewKVStoreService(keys[oracletypes.StoreKey]),
+		appCodec,
+		app.MarketMapKeeper,
+		authtypes.NewModuleAddress(adminmoduletypes.ModuleName))
+	app.OracleKeeper = &oracleKeeper
+	oracleModule := oracle.NewAppModule(appCodec, *app.OracleKeeper)
```
mmikev, line 770: 1) `NewKeeper` returns **a structure** on L761;
2) `app.OracleKeeper` is a pointer to a structure;
3) `NewAppModule` requires a structure, so you deref the pointer;

Maybe we should suggest Skip to use pointers everywhere?

app/app.go
```diff
@@ -995,6 +1054,8 @@ func New(
 		feeBurnerModule,
 		cronModule,
 		dexModule,
+		// marketmapModule,
+		// oracleModule,
```
mmikev: Please remove a commented code

app/upgrades/nextupgrade/constants.go
```diff
@@ -1,23 +1,27 @@
...
 	globalfeetypes "github.com/neutron-org/neutron/v3/x/globalfee/types"
+	marketmaptypes "github.com/skip-mev/slinky/x/marketmap/types"
+	oracletypes "github.com/skip-mev/slinky/x/oracle/types"
 )
 
 const (
 	// UpgradeName defines the on-chain upgrade name.
-	UpgradeName = "NextUpgrade"
+	UpgradeName        = "nextupgrade"
+	AdminModuleAddress = "neutron1hxskfdxpp5hqgtjj6am6nkjefhfzj359x0ar3z"
```
mmikev: Please derive it from a module address

pr0n00gler (changes requested):
Please fix linter issues

.tool-versions
sotnikov-s: to match the version specified in go.mod

```suggestion
golang 1.22.1
```

app/app.go
```diff
@@ -1059,12 +1120,65 @@ func New(
...
+	// Connect to the oracle service (default timeout of 5 seconds).
+	go func() {
+		// TODO: change background to something cancellable across the app?
+		if err := app.oracleClient.Start(context.Background()); err != nil {
+			app.Logger().Error("failed to start oracle client", "err", err)
+			panic(err)
+		}
+
+		app.Logger().Info("started oracle client", "address", cfg.OracleAddress)
+	}()
```
sotnikov-s: 1. I see you start oracle client, but you don't stop it. I believe a deferred Stop call is missing..?
2. there's no reason to print "started oracle client" because the very same logging is made inside the oracle client's Start method: https://github.com/skip-mev/slinky/blob/4c45a1de0482d2d1f52e2a174dcbc86db942dec7/service/clients/oracle/client.go#L152
image
3. what's the point of running client in a goroutine here? why don't just make it in the sync way and crash in case oracle doesn't work? I think this is the expected behaviour — if I was a validator, I'd prefer to not have a running node at all instead of a running node with the oracle implicitly down

app/app.go
```diff
@@ -1059,12 +1120,65 @@ func New(
...
+
+	// Create special kind of store to implement ValidatorStore interfaces for ConsumerKeeper (as we don't have StakingKeeper)
+	ccvconsumerCompatKeeper := voteweighted.NewCCVConsumerCompatKeeper(app.ConsumerKeeper)
+
+	// Create the proposal handler that will be used to fill proposals with
+	// transactions and oracle data.
+	proposalHandler := proposals.NewProposalHandler(
```
sotnikov-s: what do you think about adding a number of elaborations to the code (variable names, comments, import names)?

1. rename the `handler` introduced at the line 1117 to `blockSdkProposalHandler` and rename the `proposalHandler` introduced at the line 1164 to `slinkyProposalHandler` or `oracleProposalHandler` (similar to `oraclePreBlockHandler` introduced below). motivation is to 1) emphasise the origins and designations of the handlers; 2) improve readability (now it's hard to tell apart the difference between `handler` and `proposalHandler`
2. rename the `"github.com/skip-mev/slinky/abci/proposals"` import (used at the line 1664 as `proposals.NewProposalHandler`) to, say, `slinkyproposals` or `oracleproposals` to, again, emphasise the origin of the import and the function being called

network/init.sh
```diff
@@ -14,6 +14,10 @@ RPCPORT=${RPCPORT:-26657}
 RESTPORT=${RESTPORT:-1317}
 ROSETTA=${ROSETTA:-8081}
 
+ORACLE_ADDRESS=${ORACLE_ADDRESS:-localhost:8080}
+ORACLE_METRICS_ENABLED=${ORACLE_METRICS_ENABLED:-true}
+ORACLE_CLIENT_TIMEOUT=${ORACLE_CLIENT_TIMEOUT:-2s}
```
sotnikov-s: 2s seems too much given 1s local/cloud block time. let's switch to something lesser? like 500ms

network/init.sh
```diff
@@ -79,6 +83,17 @@ sed -i -e 's#"tcp://localhost:1317"#"tcp://0.0.0.0:'"$RESTPORT"'"#g' "$CHAIN_DIR
 sed -i -e 's#"tcp://0.0.0.0:1317"#"tcp://0.0.0.0:'"$RESTPORT"'"#g' "$CHAIN_DIR/config/app.toml"
 sed -i -e 's#":8080"#":'"$ROSETTA_1"'"#g' "$CHAIN_DIR/config/app.toml"
 
+# NOTE: we provide prometheus_server_address but it does not matter, since we don't use it, just need non-empty value
```
sotnikov-s, line 86: what do you mean we don't user it? I see a `true` default value above for `ORACLE_METRICS_ENABLED`. am I missing something?

network/init-neutrond.sh
```diff
@@ -719,6 +719,8 @@ set_genesis_param proposer_fee                          "\"0.25\""
 set_genesis_param escrow_account_address                "\"$DAO_CONTRACT_ADDRESS_B64\","                  # builder(POB)
 set_genesis_param sudo_call_gas_limit                   "\"1000000\""                                     # contractmanager
 set_genesis_param max_gas                               "\"1000000000\""                                  # consensus_params
+set_genesis_param vote_extensions_enable_height         "\"1\""                                           # consensus_params
+set_genesis_param market_authority                      "\"neutron1hxskfdxpp5hqgtjj6am6nkjefhfzj359x0ar3z\","                      # marketmap
```
sotnikov-s: what is this address? why not a const?

sotnikov-s (changes requested):

network/init.sh
mmikev: Is it related to block time?

sotnikov-s: well, this timeout [is used](https://github.com/skip-mev/slinky/blob/6530aaa172d11fd5bc35019dead6fddc5bd86218/abci/ve/vote_extension.go#L136) in ExtendVote in each block, and each validator makes a request to oracle with this timeout. so, if oracle is unavailable, node will be stuck until the timeout is reached. correct me if I'm wrong

mmikev: Nice, that's a good catch actually.

We need to keep this in mind when we will create recommended params of oracle to vals

sotnikov-s: it should be low for sure as far as I understand. a validator should better submit a nil vote extension rather than become stuck waiting for either long timeout or parent context cancellation

app/app.go
NeverHappened: lets do that, i felt that when wrote the code

network/init.sh
NeverHappened: hm, interesting, i would investigate into this

network/init.sh
NeverHappened: yes, metrics enabled, but they're also having a "prometheus_server_address" value, that is for starting a second prometheus server (i don't know what for), and it should be non-empty because of validation.
But it's not used in our scenario, so we provide some non empty value here

network/init-neutrond.sh
NeverHappened: this is admin module address, value used for the check of who can change data inside marketmap module.
not sure if constant makes sense here.

app/app.go
NeverHappened: about point number 3 - app.oracleClient.Start is calll syncronous? and if it fails to run, it will return err != nil and crush. Or I'm missing something

sotnikov-s: it is sync according to the implementation: https://github.com/skip-mev/slinky/blob/4c45a1de0482d2d1f52e2a174dcbc86db942dec7/service/clients/oracle/client.go#L108. it is sync, but it [doesn't wait](https://github.com/skip-mev/slinky/blob/4c45a1de0482d2d1f52e2a174dcbc86db942dec7/service/clients/oracle/client.go#L126) until the connection is established unless we configure it with an option to perform [BlockingDial](https://github.com/skip-mev/slinky/blob/4c45a1de0482d2d1f52e2a174dcbc86db942dec7/service/clients/oracle/options.go#L9). I think it makes sense to create a context with some timeout and pass it to the `oracleClient.Start` method and use the BlockingDial

sotnikov-s: please doublecheck my suggestions and make a kind of (manual?) test covering the scenarios of unavailable oracle

sotnikov-s (approved):

pr0n00gler (approved):

NeverHappened:
https://github.com/neutron-org/neutron-tests/actions/runs/8858537920

PR #499, opened 2024-04-24
chore: explicit message validation #NTRN-257

joldie777:
Task: https://hadronlabs.atlassian.net/browse/NTRN-257

List of changes:
1. Added message validation in each MsgServer handler.
2. Implemented unit tests.


joldie777:
Test run: https://github.com/neutron-org/neutron-tests/actions/runs/8823699280

x/contractmanager/keeper/msg_server_test.go
```diff
@@ -0,0 +1,44 @@
...
+	"github.com/stretchr/testify/require"
+
+	"github.com/neutron-org/neutron/v3/testutil/contractmanager/keeper"
+	"github.com/neutron-org/neutron/v3/x/contractmanager/types"
+)
+
+func TestMsgUpdateParamsValidate(t *testing.T) {
+	k, ctx := keeper.ContractManagerKeeper(t, nil)
+
+	tests := []struct {
```
mmikev, line 15: What do you think of adding a successful testcase here and everywhere below?

joldie777: I decided not to add them since this would entail the need to mock methods that will be called after successful validation pass. Also it seems like these cases are covered in already implemented handler tests.
But I can add them if it's needed.

pr0n00gler:
LGTM, but please resolve conflicts

joldie777:
> LGTM, but please resolve conflicts

Done

joldie777:
Test run: https://github.com/neutron-org/neutron-tests/actions/runs/8940453918

pr0n00gler (approved):

x/globalfee/types/tx.go
```diff
@@ -0,0 +1,35 @@
+package types
```
swelf19, line 1: It seems that we don't need this entire file except ValidateBasic. All these methods were part of interfaces, for example LegacyMsg. But in the current version of SDK they are not necessary. And if we don’t use them explicitly, then we don’t have to implement them.

x/interchaintxs/keeper/msg_server_test.go
```diff
@@ -29,6 +30,73 @@ import (
...
+			types.ErrEmptyInterchainAccountID,
+		},
+		{
+			"long interchain account id",
+			types.MsgRegisterInterchainAccount{
+				FromAddress:         testutil.TestOwnerAddress,
+				ConnectionId:        "connection-id",
+				InterchainAccountId: string(make([]byte, 48)),
+			},
+			types.ErrLongInterchainAccountID,
```
swelf19, line 88: may be out of scope, but it's not the best error value. Additional "got X, want Y" would be nice.

x/dex/keeper/msg_server.go
```diff
@@ -26,6 +27,10 @@ func (k MsgServer) Deposit(
 	goCtx context.Context,
 	msg *types.MsgDeposit,
 ) (*types.MsgDepositResponse, error) {
+	if err := msg.ValidateBasic(); err != nil {
+		return nil, errors.Wrap(err, "failed to validate MsgDeposit")
+	}
+
```
sotnikov-s: in the task description it is said that
> Now the logic is deprecated and msg.ValidateBasic() is not being called automatically anymore.

but in the docs a different statement is made:
image

so, is ValidateBasic called twice now? once in the BaseApp, once here in Deposit?

mmikev: This is a good catch, i missed that paragraph.

Cosmos SDK still [calls](https://github.com/cosmos/cosmos-sdk/blob/c4d9a495052b78a02723ed1df1d336efe83a4e8d/baseapp/baseapp.go#L620) method if a struct implements `HasValidateBasic` interface.

I think we can rename all `ValidateBasic` methods to `Validate` to remove the implementation and thus the method will be called only once in our MsgServers

x/globalfee/types/tx.go
joldie777: Redundant methods are removed.

x/interchaintxs/keeper/msg_server_test.go
joldie777: Additional information has been added to the error.

x/dex/keeper/msg_server.go
joldie777: All ValidateBasic() module message methods have been renamed to Validate().

swelf19 (approved):

quasisamurai (approved):

x/interchainqueries/keeper/keeper_test.go
```diff
@@ -404,7 +397,7 @@ func (suite *KeeperTestSuite) TestUpdateInterchainQuery() {
 
 	for i, tt := range tests {
 		tt := tt
-		suite.Run(fmt.Sprintf("Case %s, %d/%d tests", tt.name, i, len(tests)), func() {
+		suite.Run(fmt.Sprintf("Case %s, %d/%d tests", tt.name, i+1, len(tests)), func() {
```
sotnikov-s, line 400: since you started this, could you please replace `i` with `i+1` in other similar places in the project? 
- x/interchainqueries/keeper/grpc_query_test.go
- x/interchainqueries/keeper/process_block_results_test.go

x/cron/types/tx.go
```diff
@@ -27,7 +27,7 @@ func (msg *MsgUpdateParams) GetSignBytes() []byte {
 	return ModuleCdc.MustMarshalJSON(msg)
 }
 
-func (msg *MsgUpdateParams) ValidateBasic() error {
+func (msg *MsgUpdateParams) Validate() error {
```
sotnikov-s, line 30: let's add a similar check to security address?
image
```go
sdk.AccAddressFromBech32(msg.Params.SecurityAddress)
```

x/feeburner/types/tx.go
```diff
@@ -27,7 +27,7 @@ func (msg *MsgUpdateParams) GetSignBytes() []byte {
 	return ModuleCdc.MustMarshalJSON(msg)
 }
 
-func (msg *MsgUpdateParams) ValidateBasic() error {
+func (msg *MsgUpdateParams) Validate() error {
```
sotnikov-s, line 30: same here with msg.Params.TreasuryAddress

x/feerefunder/types/tx.go
```diff
@@ -27,7 +27,7 @@ func (msg *MsgUpdateParams) GetSignBytes() []byte {
 	return ModuleCdc.MustMarshalJSON(msg)
 }
 
-func (msg *MsgUpdateParams) ValidateBasic() error {
+func (msg *MsgUpdateParams) Validate() error {
```
sotnikov-s, line 30: let's check here that msg.Params.Fees are roughly valid?
- RecvFee is empty
- AckFee and TimeoutFee are not empty

x/tokenfactory/types/msgs.go
```diff
@@ -74,7 +74,7 @@ func NewMsgMintTo(sender string, amount sdk.Coin, mintToAddress string) *MsgMint
 
 func (m MsgMint) Route() string { return RouterKey }
 func (m MsgMint) Type() string  { return TypeMsgMint }
-func (m MsgMint) ValidateBasic() error {
+func (m MsgMint) Validate() error {
```
sotnikov-s, line 77: let's add a `sdk.AccAddressFromBech32(m. MintToAddress)` check here

x/tokenfactory/types/msgs.go
```diff
@@ -116,7 +116,7 @@ func NewMsgBurnFrom(sender string, amount sdk.Coin, _ string) *MsgBurn {
 
 func (m MsgBurn) Route() string { return RouterKey }
 func (m MsgBurn) Type() string  { return TypeMsgBurn }
-func (m MsgBurn) ValidateBasic() error {
+func (m MsgBurn) Validate() error {
```
sotnikov-s, line 124: let's add a `sdk.AccAddressFromBech32(m. BurnFromAddress)` check here

x/tokenfactory/types/tx.go
```diff
@@ -27,7 +27,7 @@ func (msg *MsgUpdateParams) GetSignBytes() []byte {
 	return ModuleCdc.MustMarshalJSON(msg)
 }
 
-func (msg *MsgUpdateParams) ValidateBasic() error {
+func (msg *MsgUpdateParams) Validate() error {
```
sotnikov-s, line 30: same: let's check msg.Params.FeeCollectorAddress here

sotnikov-s (commented):
since we've already touched some Validate methods logic, let's improve some more of them?

x/tokenfactory/types/msgs.go
sotnikov-s: no, we cannot just to that because [if the address is empty the sender address is used](https://github.com/neutron-org/neutron/blob/1c2388b321f215a024923aefc49cb7a929a8dfad/x/tokenfactory/keeper/msg_server.go#L72-L74). we should do the check only if the sender addr is not an empty string

x/tokenfactory/types/msgs.go
sotnikov-s: no, we cannot just to that because [if the address is empty the sender address is used](https://github.com/neutron-org/neutron/blob/1c2388b321f215a024923aefc49cb7a929a8dfad/x/tokenfactory/keeper/msg_server.go#L108-L110). we should do the check only if the sender addr is not an empty string

x/interchainqueries/keeper/keeper_test.go
joldie777: Done

x/cron/types/tx.go
joldie777: Done

x/feeburner/types/tx.go
joldie777: Done

x/feerefunder/types/tx.go
joldie777: Done

x/tokenfactory/types/msgs.go
joldie777: Done

x/tokenfactory/types/msgs.go
joldie777: Done

x/tokenfactory/types/tx.go
joldie777: Done

sotnikov-s (approved):

PR #514, opened 2024-05-08
Feat: enable vote extensions in upgrade handler

pr0n00gler:
This PR adds a simple method to enable Vote Extensions functionality (Slinky requires it) in v4 upgrade handler

### DIFF
--- a/app/upgrades/v4.0.0/upgrades.go
+++ b/app/upgrades/v4.0.0/upgrades.go
@@ -4,8 +4,11 @@ import (
 	"context"
 	"fmt"
 
+	comettypes "github.com/cometbft/cometbft/proto/tendermint/types"
 	adminmoduletypes "github.com/cosmos/admin-module/x/adminmodule/types"
 	authtypes "github.com/cosmos/cosmos-sdk/x/auth/types"
+	consensuskeeper "github.com/cosmos/cosmos-sdk/x/consensus/keeper"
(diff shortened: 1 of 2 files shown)


app/upgrades/v4.0.0/upgrades.go
```diff
@@ -57,3 +66,24 @@ func setMarketMapParams(ctx sdk.Context, marketmapKeeper *marketmapkeeper.Keeper
...
 }
+
+func enableVoteExtensions(ctx sdk.Context, consensusKeeper *consensuskeeper.Keeper) error {
+	oldParams, err := consensusKeeper.Params(ctx, &types.QueryParamsRequest{})
+	if err != nil {
+		return err
+	}
+
+	// we need to enable VoteExtensions for Slinky
+	oldParams.Params.Abci = &comettypes.ABCIParams{VoteExtensionsEnableHeight: ctx.BlockHeight()}
```
aljo242:
```suggestion
	oldParams.Params.Abci = &comettypes.ABCIParams{VoteExtensionsEnableHeight: ctx.BlockHeight() + 4}
```

we would prefer that the VE enabled height is set to some offset after the upgrade.

This will make it easier for all to debug the upgrade in case there are issues.  Only the state will be set here, all of that can be validated for a few blocks, and then the VEs can kick in.

swelf19 (approved):

NeverHappened (approved):

PR #518, opened 2024-05-09
Fix: missing proto signer

pr0n00gler:
This PR adds missing proto option to define signer of a message


pr0n00gler:
https://github.com/neutron-org/neutron-tests/actions/runs/9015204675

PR #519, opened 2024-05-09
fix: use basicmodulemanager as recommended in sdk-50 upgrade guide

Lockwarr:
Initial commit, needs review - check the changes for anything missing

 if I overwrite the 'transfer' key in the newBasicModuleManager with neutron/x/transfer appModule, tests are failing with:
 `test panicked: failed to encode client state: *tendermint.ClientState does not have a registered interface`
 And if I don't, test are failing with an error that neutron.MsgTransfer TypeURL is not registered.
 
Edit: Since neutron/x/transfer module uses the storeKey and ModuleName from native ibctransfer, maybe we have a conflict with this new basicModuleManager?


pr0n00gler:
Thanks, it will be merged in: https://github.com/neutron-org/neutron/pull/607

PR #527, opened 2024-05-14
Feat/Behind Enemy Lines enforcement

jcompagni10:
Prevent users from depositing behind enemy lines. 

This is critical for prevent a DOS vector where users deposit many ticks of dust causing swaps to have to touch a large number of ticks. By preventing users from depositing at a price cheaper than the opposing token price (ie. on the wrong side of the spread), we greatly limit the surface area for ticks that can be dusted. Most ticks near the spread will already have liquidity so dusting is impossible or limit to a small set of un-used fee tiers. While users can still place dust towards the back of the order book this will not effect swaps. 

Additionally, this maintains a nice invariant that all liquidity in the dex exists across a partitioned spread with no overlap in prices.


jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/9276692068

x/dex/keeper/core.go
```diff
@@ -49,11 +49,22 @@ func (k Keeper) DepositCore(
...
 		if err := k.ValidateFee(ctx, fee); err != nil {
-			return nil, nil, nil, err
+			return nil, nil, nil, failedDeposits, err
 		}
-		autoswap := !options[i].DisableAutoswap
+
+		if k.IsPoolBehindEnemyLines(ctx, pairID, tickIndex, fee, amount0, amount1) {
+			err = sdkerrors.Wrapf(types.ErrDepositBehindEnemyLines,
+				"deposit failed at tick %d fee %d", tickIndex, fee)
+			if options.FailTxOn_BEL {
```
NeverHappened: why not FailTxOnBEL? kinda breaks the naming

x/dex/keeper/core.go
```diff
@@ -49,11 +49,22 @@ func (k Keeper) DepositCore(
 		amount1 := amounts1[i]
 		tickIndex := tickIndices[i]
 		fee := fees[i]
+		options := options[i]
```
NeverHappened: maybe better `option := options[i]`
to avoid rewriting base var

proto/neutron/dex/tx.proto
```diff
@@ -29,6 +29,7 @@ service Msg {
 
 message DepositOptions {
   bool disable_autoswap = 1;
+  bool fail_tx_on_BEL = 2;
```
NeverHappened: why not fail_tx_on_bel?

joldie777 (approved):

NicholasDotSol (approved):

NeverHappened (changes requested):

pr0n00gler (approved):

quasisamurai (approved):

jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/9276692068

pr0n00gler:
Resolve conflicts, pls

pr0n00gler (approved):

PR #531, opened 2024-05-15
NTRN-282 dex telemetry

quasisamurai:
[TASK](https://hadronlabs.atlassian.net/browse/NTRN-282)

- [x] Gas per swap
- [x] Number limit orders placed
- [x] Total amount withdrawn from LP and LO
- [x] Total Number of expired limit orders on orderbook
- [x] Total tickLiquidities


x/dex/keeper/core.go
```diff
@@ -210,6 +210,7 @@ func (k Keeper) WithdrawCore(
 			receiverAddr,
```
jcompagni10, line 226: Need to add incWithdraw for `coin0`

x/dex/keeper/core.go
```diff
@@ -299,6 +300,8 @@ func (k Keeper) MultiHopSwapCore(
 		bestRoute.route,
 	))
 
+	gasConsumed(ctx)
```
jcompagni10: Not sure what this is tracking. 
In the initial spec I had talked about tracking gas per swap which would measure gas used by calls into `k.Swap`

x/dex/keeper/keeper.go
```diff
@@ -45,3 +57,47 @@ func (k Keeper) Logger(ctx sdk.Context) log.Logger {
 func (k Keeper) GetAuthority() string {
 	return k.authority
 }
+
+func incWithdrawnAmount(coins sdk.Coins) {
+	for _, coin := range coins {
+		divisor := big.NewInt(6)
```
jcompagni10: Why are we dividing by 6?

x/dex/keeper/core.go
jcompagni10: https://www.notion.so/hadron/Duality-Monitoring-a9c194e4e81f490dbf97ca39fee7ad2e?pvs=4#bcdc3d47c3a74c9f9fe2a0d994ec4406

x/dex/keeper/limit_order_expiration.go
```diff
@@ -152,5 +152,6 @@ func (k Keeper) PurgeExpiredLimitOrders(ctx sdk.Context, curTime time.Time) {
 		}
 
 		k.RemoveLimitOrderExpirationByKey(ctx, iterator.Key())
+		incExpiredOrders()
```
jcompagni10: Not sure I understand this metric either. I think we want to track active expiring limit orders. 
Probably have one metric here https://github.com/neutron-org/neutron/blob/feat%2Fdex-telemetry/x/dex/keeper/limit_order_tranche.go#L223
And then decrement where you have the metric above.

x/dex/keeper/msg_server.go
```diff
@@ -119,6 +119,8 @@ func (k MsgServer) PlaceLimitOrder(
 		return &types.MsgPlaceLimitOrderResponse{}, err
 	}
 
+	totalLimitOrders()
```
jcompagni10: Not sure what this metric does either.

jcompagni10 (changes requested):
I'm confused by some of the metrics being reported here and what they are meant for. Is there a spec that describes the intended purposes of each metric?

What happens if a tx fails? Are metrics still reported? This seems undesirable for many of these metrics. I could not find any clear documentation on this.

x/dex/keeper/msg_server.go
jcompagni10: We only want to track active limit orders. I think this can be achieved with the same strategy used for the tick liquidity tracking

x/dex/keeper/keeper.go
```diff
@@ -47,3 +48,77 @@ func (k Keeper) Logger(ctx sdk.Context) log.Logger {
 func (k Keeper) GetAuthority() string {
 	return k.authority
 }
+
```
jcompagni10: would be preferable to have all of these functions in dex/types/events.go for consistency sake

x/dex/types/types.go
```diff
@@ -1 +1,15 @@
 package types
+
+const (
+	// EventTypeNeutronMessage defines the event type used by the Interchain Queries module events.
+	EventTypeNeutronMessage = "neutron"
+
+	AttributeDenom                   = "denom"
+	AttributeWithdrawn               = "total_withdrawn"
+	AttributeGasConsumed             = "gas_consumed"
+	AttributeIncExpiredOrders        = "total_orders_expired_inc"
```
jcompagni10: For clarity, I would call this "total_limit_orders_expiring_[inc/dec]" they are not expired when they are created

x/dex/keeper/keeper.go
jcompagni10: Since we are now just using events and there is not an issue with overloading cardinality it would be nice to add an additional field for pairId, so that have the option to segregate all of these events by specific pairs.

x/dex/keeper/msg_server.go
```diff
@@ -215,6 +217,9 @@ func (k MsgServer) MultiHopSwap(
 	goCtx context.Context,
 	msg *types.MsgMultiHopSwap,
 ) (*types.MsgMultiHopSwapResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
```
jcompagni10: Sorry, I know the naming is a bit confusing. But we want to track gas anytime we are doing a swap (LimitOrders + multihop swap_. The correct place to do that is here https://github.com/neutron-org/neutron/blob/9d2df9d1ba8d34591b89c1eaeecc426d4acd4f97/x/dex/keeper/liquidity.go#L11

x/dex/types/types.go
```diff
@@ -1 +1,15 @@
...
+	AttributeDecExpiredOrders        = "total_orders_expired_dec"
+	AttributeTotalLimitOrders        = "total_orders_limit"
```
jcompagni10: "total_limit_orders"

x/dex/keeper/keeper.go
```diff
@@ -47,3 +48,77 @@ func (k Keeper) Logger(ctx sdk.Context) log.Logger {
...
+func getEventsIncTotalOrders() sdk.Events {
+	return sdk.Events{
+		sdk.NewEvent(
+			types.EventTypeNeutronMessage,
+			sdk.NewAttribute(sdk.AttributeKeyModule, types.ModuleName),
+			sdk.NewAttribute(sdk.AttributeKeyAction, types.AttributeTotalLimitOrders),
+		),
+	}
+}
+
```
jcompagni10: Also need DecTotalOrders

x/dex/keeper/limit_order_tranche.go
```diff
@@ -57,6 +57,7 @@ func (k Keeper) SaveTranche(ctx sdk.Context, tranche *types.LimitOrderTranche) {
 	} else {
 		k.SetInactiveLimitOrderTranche(ctx, tranche)
 		k.RemoveLimitOrderTranche(ctx, tranche.Key)
+		ctx.EventManager().EmitEvents(getEventsDecTotalTickLiquidities())
```
jcompagni10: Either need to unify TickLiquidities, TotalLimitOrders TotalExpiringLimitOrders or have dec events for LimitOrders and ExpiringLimitOrders here as well

jcompagni10 (changes requested):

x/dex/keeper/limit_order_expiration.go
```diff
@@ -154,5 +154,10 @@ func (k Keeper) PurgeExpiredLimitOrders(ctx sdk.Context, curTime time.Time) {
 		}
 
 		k.RemoveLimitOrderExpirationByKey(ctx, iterator.Key())
+		// TODO: either delete tradePairID from this event or provide pair ID somehow (seems it's unreachable here)
+		ctx.EventManager().EmitEvents(types.GetEventsDecExpiringOrders(&types.TradePairID{
+			MakerDenom: "",
```
jcompagni10: Tranche is accessible on line 147, you can get PairId there.

x/dex/keeper/pool.go
```diff
@@ -136,19 +136,20 @@ func (k Keeper) GetPoolIDByParams(
...
 	ctx.EventManager().EmitEvent(types.CreateTickUpdatePoolReserves(*pool.LowerTick0))
 	ctx.EventManager().EmitEvent(types.CreateTickUpdatePoolReserves(*pool.UpperTick1))
 }
 
-func (k Keeper) updatePoolReserves(ctx sdk.Context, reserves *types.PoolReserves) {
+func (k Keeper) updatePoolReserves(ctx sdk.Context, reserves *types.PoolReserves, poolID uint64) {
 	if reserves.HasToken() {
 		k.SetPoolReserves(ctx, reserves)
 	} else {
+		ctx.EventManager().EmitEvents(types.GetEventsDecTotalPoolReserves(poolID))
```
jcompagni10: I would use pairID not poolID for consistency

jcompagni10 (changes requested):

x/dex/keeper/pool.go
jcompagni10: PairId does not need to be passed all the way through. It can be accessed via `reserves.Key.TradePairId.MustPairID()`

x/dex/keeper/core.go
```diff
@@ -605,6 +607,7 @@ func (k Keeper) WithdrawFilledLimitOrderCore(
 		coinTakerDenomOut := sdk.NewCoin(tradePairID.TakerDenom, amountOutTokenOut)
 		coinMakerDenomRefund := sdk.NewCoin(tradePairID.MakerDenom, remainingTokenIn)
 		coins := sdk.NewCoins(coinTakerDenomOut, coinMakerDenomRefund)
+		ctx.EventManager().EmitEvents(types.GetEventsWithdrawnAmount(coins))
```
jcompagni10: For consistency, I would just track CoinTakerDenomOut. coinMakerDenomRefund is the same coin they would get if they cancel the LO, which we don't track.

x/dex/keeper/msg_server.go
```diff
@@ -152,6 +152,11 @@ func (k MsgServer) PlaceLimitOrder(
 		return &types.MsgPlaceLimitOrderResponse{}, err
 	}
 
+	ctx.EventManager().EmitEvents(types.GetEventsIncTotalOrders(&types.TradePairID{
```
jcompagni10: This should be removed. You are already incrementing in the correct place in keeper.limit_order_tranche.go

jcompagni10 (changes requested):

pr0n00gler (dismissed):

x/dex/keeper/limit_order_tranche.go
```diff
@@ -57,6 +57,11 @@ func (k Keeper) SaveTranche(ctx sdk.Context, tranche *types.LimitOrderTranche) {
 	} else {
 		k.SetInactiveLimitOrderTranche(ctx, tranche)
 		k.RemoveLimitOrderTranche(ctx, tranche.Key)
+		// TODO: either delete tradePairID from this event or provide pair ID somehow (seems it's unreachable here)
+		ctx.EventManager().EmitEvents(types.GetEventsDecTotalOrders(&types.TradePairID{
+			MakerDenom: "",
+			TakerDenom: "",
+		}))
```
jcompagni10:
```suggestion
ctx.EventManager().EmitEvents(types.GetEventsDecTotalOrders(tranche.Key.TradePairId))
```

jcompagni10 (approved):

jcompagni10 (dismissed):

jcompagni10 (approved):

pr0n00gler (approved):

PR #543, opened 2024-05-29
feat: fix estimateXX dex queries #NTRN-318

jcompagni10:
Task: https://hadronlabs.atlassian.net/browse/NTRN-318

Related PRs:
https://github.com/neutron-org/neutron-sdk/pull/149
https://github.com/neutron-org/neutron-dev-contracts/pull/55
https://github.com/neutron-org/neutron-integration-tests/pull/315

List of changes:
1. Don't enforce balance checks when running estimateXX queries


proto/neutron/dex/query.proto
```diff
@@ -260,25 +260,22 @@ message QueryGetPoolReservesResponse {
 }
 
 message QueryEstimateMultiHopSwapRequest {
-  string creator = 1;
-  string receiver = 2;
-  repeated MultiHopRoute routes = 3;
-  string amount_in = 4 [
+  repeated MultiHopRoute routes = 1;
```
mmikev: I don't think we can just change field ids.
It will break backwards compatibility

pr0n00gler (changes requested):

x/dex/keeper/grpc_query_estimate_multi_hop_swap.go
```diff
@@ -8,14 +8,13 @@ import (
...
-// TODO: This doesn't run ValidateBasic() checks.
 func (k Keeper) EstimateMultiHopSwap(
 	goCtx context.Context,
 	req *types.QueryEstimateMultiHopSwapRequest,
 ) (*types.QueryEstimateMultiHopSwapResponse, error) {
 	msg := types.MsgMultiHopSwap{
-		Creator:        req.Creator,
-		Receiver:       req.Receiver,
+		Creator:        "neutron1dft8nwxzr0u27wvr2cknpermjkreqvp9fdy0uz",
+		Receiver:       "neutron1dft8nwxzr0u27wvr2cknpermjkreqvp9fdy0uz",
```
mmikev, line 18: What is this address?

proto/neutron/dex/query.proto
```diff
@@ -302,13 +297,12 @@ message QueryEstimatePlaceLimitOrderRequest {
     (gogoproto.jsontag) = "amount_in"
   ];
   LimitOrderType order_type = 7;
-
   // expirationTime is only valid iff orderType == GOOD_TIL_TIME.
-  google.protobuf.Timestamp expiration_time = 8 [
+  google.protobuf.Timestamp expiration_time = 9 [
```
mmikev: Wire numbers have been changed, let's do not do that

pr0n00gler (changes requested):

x/dex/keeper/grpc_query_estimate_multi_hop_swap.go
jcompagni10: Just a random valid address. Will add a comment. But let me know if there's a more preferable way to do this. I agree it's a bit odd.

jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/9319924715

x/dex/keeper/grpc_query_estimate_multi_hop_swap.go
NeverHappened: Probably need to create another message `QueryMultiHopSwap` with it's own validate and reusing validate logic somehow if possible.
Otherwise kinda ugly

pr0n00gler (changes requested):
Integration tests didn't pass

joldie777:
Test run: https://github.com/neutron-org/neutron-tests/actions/runs/9597098058

x/dex/keeper/grpc_query_estimate_multi_hop_swap.go
```diff
@@ -28,23 +28,23 @@ func (k Keeper) EstimateMultiHopSwap(
 	ctx := sdk.UnwrapSDKContext(goCtx)
 	cacheCtx, _ := ctx.CacheContext()
 
-	callerAddr := sdk.MustAccAddressFromBech32(req.Creator)
-	receiverAddr := sdk.MustAccAddressFromBech32(req.Receiver)
-
+	oldBk := k.bankKeeper
+	k.bankKeeper = NewSimulationBankKeeper(k.bankKeeper)
```
sotnikov-s: do we do that bank keeper substitution only because of [these module<>bankKeeper interactions](https://github.com/neutron-org/neutron/blob/e57b3d56eb86faacb86f687da57a1fa5d7c7e6e9/x/dex/keeper/core.go#L292-L312)? also I can see that we use a cached context to not update the dex keeper's state. if yes, I think we should rethink the way we implemented the code.

for example, speaking of the [MultiHopSwapCore](https://github.com/neutron-org/neutron/blob/e57b3d56eb86faacb86f687da57a1fa5d7c7e6e9/x/dex/keeper/core.go#L238) method: the logic for a simulation and for a real swap is mostly the same:
**1.** calc the best route **(common logic for both cases)**
**2.** write route context (mutate dex storage as far as I understand)
**3.** send coins between receiver, caller and module account
**4.** emit events

why don't we create a separate method that does only the **1.** step and is used in both simulation and real swap? in simulation we only call this method and return the result, but in the real swap we follow it up with [the remaining necessary steps](https://github.com/neutron-org/neutron/blob/e57b3d56eb86faacb86f687da57a1fa5d7c7e6e9/x/dex/keeper/core.go#L291-L323)?

sotnikov-s: more or less the same for [PlaceLimitOrderCore](https://github.com/neutron-org/neutron/blob/e57b3d56eb86faacb86f687da57a1fa5d7c7e6e9/x/dex/keeper/core.go#L330)

x/dex/keeper/grpc_query_estimate_place_limit_order.go
```diff
@@ -53,14 +47,15 @@ func (k Keeper) EstimatePlaceLimitOrder(
...
+		[]byte("caller"),
+		[]byte("receiver"),
 	)
 	if err != nil {
 		return nil, err
 	}
 
-	// NB: We're only using a cache context so we don't expect any writes to happen.
+	//nolint:staticcheck // Should be unnecessary but out of an abundance of caution we restore the old bankkeeper
+	k.bankKeeper = oldBk
```
swelf19: What happens if we do an early exit at the line 54 and do not put the real keeper back?
Even if you do a `defer` to retern the keeper back, since a query is not transactional, you can brake the dex keeper with 2 concurrent queries


| q1                                     | q2                                                            |
| -------------------------------------- | ------------------------------------------------------------- |
| oldBk := k.bankKeeper                  |                                                               |
| k.bankKeeper = NewSimulationBankKeeper |                                                               |
|                                        | oldBk := k.bankKeeper (actually it's NewSimulationBankKeeper) |
|                                        | k.bankKeeper = NewSimulationBankKeeper                        |
| k.bankKeeper = oldBk                   |                                                               |
|                                        | k.bankKeeper = oldBk (NewSimulationBankKeeper)                                          |

and now we have dex keeper configured with SimulationBankKeeper, dont we?

if i got it right, my point is - it's pretty dangerous to reconfigure keeper during a query

x/dex/keeper/grpc_query_estimate_multi_hop_swap.go
swelf19: Agree, i prefer a pure function that does estimation.

swelf19 (changes requested):

jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/10062625296

jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/10063040043

x/dex/keeper/grpc_estimate_place_limit_order_test.go
```diff
@@ -0,0 +1,105 @@
+package keeper_test
```
swelf19, line 1: lets rename the file to grpc_**query**_estimate_place_limit_order_test to get exact before grpc_query_estimate_place_limit_order.go in listing

x/dex/keeper/grpc_estimate_place_limit_order_test.go
```diff
@@ -0,0 +1,105 @@
...
+	s.SetupMultiplePools(
+		NewPoolSetup("TokenA", "TokenB", 0, 1, 0, 1),
+		NewPoolSetup("TokenA", "TokenB", 0, 1, 1, 1),
+	)
+
+	// WHEN estimate GTC Limit order selling "TokenA"
+	resp, err := s.App.DexKeeper.EstimatePlaceLimitOrder(s.Ctx, &types.QueryEstimatePlaceLimitOrderRequest{
+		TokenIn:          "TokenA",
+		TokenOut:         "TokenB",
+		TickIndexInToOut: 4,
```
swelf19, line 21: it's out of scope of the pr, but `EstimatePlaceLimitOrder` looks outdated, TickIndexInToOut is not marked as deprecated and `limit_sell_price` is missed.
What do you think about to unify these quries with something like
```proto
message QueryEstimatePlaceLimitOrderRequest {
    MsgPlaceLimitOrder msg = 1;
}
```
to not copypaste whole bunch of fields. The only downside i can see you can get rid of `creator` and `receiver` in a query

swelf19 (commented):
.

PR #547, opened 2024-06-01
DUAL-10 Fix: store tick prices and lookup instead of doing log calculation

jcompagni10:
Previously we used float64 log to calculate a tick from price. There is no strong guarantee that float64 math will perform the same on all issues and poses a small risk of being consensus-breaking. To fix this we switch to using a pre-calculated array of prices which we binary search to find the correct tick for a given price. This is also slightly more efficient and gives us the option to stop doing all price calculations in the future. 

Benchmarks:
BenchmarkCalcTickIndexFromPriceV2   	  301816	      5249 ns/op
BenchmarkCalcTickIndexFromPrice      	  206276	      6662 ns/op 

The price data structure is loaded into memory from a .gob file at startup. The full data structure is calculated to use about 8mb of space, so should have a negligible impact on overall performance


jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/9333093601

x/dex/utils/math.go
```diff
@@ -68,6 +68,19 @@ func MustSafeUint64ToInt64(in uint64) (out int64) {
 	return safeInt64
 }
 
+// func Logv2(val, base math_utils.PrecDec) (math_utils.PrecDec, error) {
```
mmikev: Let's remove commented code?

x/dex/types/price.go
```diff
@@ -91,3 +148,34 @@ func ValidateFairOutput(amountIn math.Int, price math_utils.PrecDec) error {
 	}
 	return nil
 }
+
+// Used for generating the precomputedPrice.gob file
```
mmikev, line 151: We have `neutrond debug ...` cli commands, maybe it makes sense create an additional CLI debug command in neutron with the logic below in case we need to regenerate `.gob` file?

joldie777 (approved):

x/dex/types/price.go
jcompagni10: Happy to add it. But seems a bit strange. Neutrond debug seems targeted to end users. There's never a time an end user would need to regenerate the .gob file. The only case would be if we change the maxTickIndex or the tick price logic. I agree it's a bit funny to leave the code commented out. But in my opinion it makes sense as documentation for what the precomputed_prices.gob is and an easy way to change it if we make internal changes. 
Let me know what you think. Happy to change in whatever way makes most sense

mmikev: Yeah, agree. Let's leave it as it as

pr0n00gler (approved):

x/dex/types/price.go
```diff
@@ -12,11 +17,43 @@ const (
 	// NOTE: 559_680 is the highest possible tick at which price can be calculated with a < 1% error
 	// when using 27 digit decimal precision (via prec_dec).
 	// The error rate for very negative ticks approaches zero, so there is no concern there
-	MaxTickExp uint64 = 559_680
-	MinPrice   string = "0.000000000000000000000000495"
-	MaxPrice   string = "2020125331305056766452345.127500016657360222036663651"
+	MaxTickExp            uint64 = 559_680
+	MinPrice              string = "0.000000000000000000000000495"
+	MaxPrice              string = "2020125331305056766452345.127500016657360222036663651"
+	PrecomputedPricesFile        = "precomputed_prices.gob"
```
mmikev: This constant is unused

jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/9333093601

jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/9333093601

NicholasDotSol (approved):

PR #587, opened 2024-06-18
Feat: Whitelist tokenfactory hooks

jcompagni10:
Add explicit whitelist of codeID & denomCreator pairs to tokenfactory params. Only allow beforeSend hooks to be called for denoms where the denomCreator and codeID of hook are whitelisted.

This also adds a migration for the tokefactory params that includes hooks that we want to whitelist (NFA and astroport)

**related PRs:**
https://github.com/neutron-org/neutron-integration-tests/pull/316 --test fixes for this PR to pass

neutron dao changes: 
https://github.com/neutron-org/neutron-dao/pull/107
https://github.com/neutron-org/neutron-sdk/pull/150
https://github.com/neutron-org/neutronjsplus/pull/43
https://github.com/neutron-org/neutron-integration-tests/pull/314


proto/osmosis/tokenfactory/v1beta1/genesis.proto
```diff
@@ -29,4 +29,5 @@ message GenesisDenom {
     (gogoproto.moretags) = "yaml:\"authority_metadata\"",
     (gogoproto.nullable) = false
   ];
+  bool whitelisted = 3 [(gogoproto.moretags) = "yaml:\"whitelisted\""];
```
mmikev: Can't we just store whitelisted denoms in module's params literally as a list of tokens?

It's more convenient for maintenance

pr0n00gler (changes requested):

jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/9576807194

proto/osmosis/tokenfactory/params.proto
```diff
@@ -0,0 +1,39 @@
+syntax = "proto3";
+package osmosis.tokenfactory;
+
+import "cosmos/base/v1beta1/coin.proto";
+import "cosmos_proto/cosmos.proto";
+import "gogoproto/gogo.proto";
+
+option go_package = "github.com/neutron-org/neutron/v4/x/tokenfactory/types";
+
+message HookWhitelist {
```
sotnikov-s: maybe WhitelistedHook? HookWhitelist sounds like it's the list itself, not an entry in it

proto/osmosis/tokenfactory/params.proto
```diff
@@ -0,0 +1,39 @@
...
+  // See: https://github.com/CosmWasm/token-factory/issues/11
+  uint64 denom_creation_gas_consume = 2 [
+    (gogoproto.moretags) = "yaml:\"denom_creation_gas_consume\"",
+    (gogoproto.nullable) = true
+  ];
+
+  // FeeCollectorAddress is the address where fees collected from denom creation
+  // are sent to
+  string fee_collector_address = 3;
+  repeated HookWhitelist whitelisted_hooks = 4;
```
sotnikov-s: please add a descriptive comment to the field

proto/osmosis/tokenfactory/params.proto
sotnikov-s: please add a descriptive comment to the type

x/tokenfactory/keeper/before_send.go
```diff
@@ -99,6 +99,11 @@ func (k Keeper) callBeforeSendListener(ctx context.Context, from, to sdk.AccAddr
 				return err
 			}
 
+			// Do not invoke hook if denom is not whitelisted
+			if err := k.AssertIsHookWhitelisted(c, coin.Denom, cwAddr); err != nil {
+				return nil
```
zavgorodnii, line 104: why did we choose not to return an error here?

zavgorodnii, line 104: why `return nil`? new hooks can't be set if not whietlisted, would make sense to return an error here imo

x/tokenfactory/module.go
```diff
@@ -135,6 +135,10 @@ func (AppModule) QuerierRoute() string { return types.QuerierRoute }
 func (am AppModule) RegisterServices(cfg module.Configurator) {
 	types.RegisterQueryServer(cfg.QueryServer(), am.keeper)
 	types.RegisterMsgServer(cfg.MsgServer(), keeper.NewMsgServerImpl(am.keeper))
+	m := keeper.NewMigrator(am.keeper)
+	if err := cfg.RegisterMigration(types.ModuleName, 1, m.Migrate1to2); err != nil {
+		panic(fmt.Sprintf("failed to migrate x/tokenfactory from version 1 to 1: %v", err))
```
sotnikov-s:
```suggestion
		panic(fmt.Sprintf("failed to migrate x/tokenfactory from version 1 to 2: %v", err))
```

zavgorodnii: `version 1 to 1` doesn't look right

x/tokenfactory/keeper/before_send.go
```diff
@@ -99,6 +99,11 @@ func (k Keeper) callBeforeSendListener(ctx context.Context, from, to sdk.AccAddr
...
+			}
+
```
sotnikov-s: once have been discussed and based on what we decide will be the best behaviour here, please update the `callBeforeSendListener` to reflect the change

x/tokenfactory/types/params.go
```diff
@@ -105,3 +113,24 @@ func validateFeeCollectorAddress(i interface{}) error {
...
+	hooks, ok := i.([]*HookWhitelist)
+	if !ok {
+		return fmt.Errorf("invalid parameter type: %T", i)
+	}
+
+	seenHooks := map[string]bool{}
+	for _, hook := range hooks {
+		hookStr := hook.String()
+		if seenHooks[hookStr] {
+			return fmt.Errorf("duplicate whitelisted hook: %#v", hook)
```
sotnikov-s: what does `%#v` do? why not `%s` with `hookStr` as value?

x/tokenfactory/types/tx.go
```diff
@@ -32,9 +32,11 @@ func (msg *MsgUpdateParams) Validate() error {
 		return errorsmod.Wrap(err, "authority is invalid")
 	}
 
+	// TODO: This is inconsistent. Per Params.Validate() an empty creator address is valid as long as
+	// DenomCreationFee is nil. But This check fails if FeeCollectorAddress is unset.
 	if _, err := sdk.AccAddressFromBech32(msg.Params.FeeCollectorAddress); err != nil {
 		return errorsmod.Wrap(err, "fee_collector_address is invalid")
 	}
```
sotnikov-s, line 39: why do we even need this piece of code? why there is a part of `msg.Params` validation placed here, not in the `msg.Params.Validate`? we literally have the same validation inside the `msg.Params.Validate` method:
image

x/tokenfactory/types/params.go
jcompagni10: It is supposed to pretty-print it. But actually comparing the output I agree, hookStr is much clearer

x/tokenfactory/keeper/before_send.go
jcompagni10: Just confirming we are good with this behavior? @zavgorodnii @sotnikov-s

x/tokenfactory/types/tx.go
jcompagni10: ValidateFeeCollectorAddress allows an empty address. But params.Validate() does not. I'm fairly certain we should remove the validation of FeeCollectorAddress in params.Validate(). But doing so breaks some tests. So in the interest of keeping the commit as clear and atomic as possible, I did not want to change anything that wasn't essential.

jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/9576807194

x/tokenfactory/migrations/v2/store.go
```diff
@@ -0,0 +1,112 @@
...
+				return errors.New("cannot parse hook contract address")
+			}
+
+			err = keeper.AssertIsHookWhitelisted(ctx, denom, contractAddr)
+			if err != nil {
+				// If hook is not whitelisted delete it
+				store.Delete(iterator.Key())
+			}
+		}
+	}
```
sotnikov-s: a couple of concerns about this piece of code:
1. there are different kinds of errors returned from the `AssertIsHookWhitelisted`, but here they all are treated as the "hook is not whitelisted" case. there are also possible errors about invalid denoms and not found contracts. this is an imperfection of having the only return value that is used for both error values and true/false value. I think it would be cleaner to refactor `AssertIsHookWhitelisted` so it returns `(bool, error)`
2. the `keyParts[2]` and `keyParts[1]` — we never check whether the array is long enough beforehand which can cause to panic. it's a good practice to check the length before referring to array elements using index

I understand that it's a migration code where we kind of know the values we'll operate with and we can be sure about the array index operations and the only possible error returned from `AssertIsHookWhitelisted` (because we're sure the denom and contract address are fine), but I still find the code unclean and rough here. so I'd appreciate to see the code refactored in case you share my concerns and we have some time for doing that.
besides, if we're so sure about the values, why we handle the `AccAddressFromBech32` error? that's inconsistent

x/tokenfactory/keeper/params.go
```diff
@@ -29,3 +29,24 @@ func (k Keeper) SetParams(ctx sdk.Context, params types.Params) error {
 	store.Set(types.ParamsKey, bz)
 	return nil
 }
+
+func (k Keeper) AssertIsHookWhitelisted(ctx sdk.Context, denom string, contractAddress sdk.AccAddress) error {
```
sotnikov-s: please move the method to `keeper.go` or `before_send.go`, or to any other file you find is the best place for this method, because `params.go` is definitely not the one

x/tokenfactory/keeper/before_send.go
sotnikov-s: even though we remove all nonwhitelisted hooks during migration, I still think it's more appropriate to return an error here. we have two options here in case of an unauthorised hook: 1) we completely ignore all the hooks code (return nil here breaking the method execution) or 2) we return error here explicitly making all bank operations with unauthorised hooks fail. I prefer the second one

sotnikov-s: if we still stick to the first option described above, we should replace `return nil` with `continue` because the `return nil` happens in a loop of handling a bank operation that possibly contains multiple assets. if just return, we break the loop and ignore the rest of the assets

jcompagni10: Spoke with Andrei about this. This line is really only here as an emergency override in case something is wrong with a hook and we want to use the sub-dao to turn it off immediately. This is an extreme case and just a stop-gap before we would do an emergency upgrade to remove the problematic hook. In this scenario we do not want the offending hook to be able to block sends. 
Agree on switching to `continue` though.

docs/static/swagger.yaml
```diff
@@ -18774,6 +18795,19 @@ definitions:
...
+            items:
+              properties:
+                code_id:
+                  format: uint64
+                  type: string
+                denom_creator:
+                  type: string
+              type: object
+            title: >-
+              HookWhitelist is the list of hooks which are allowed to be added
```
zavgorodnii: might be missing something, but why do we have `whitelisted_hooks` as the key here and `HookWhitelist` as the value?

x/tokenfactory/keeper/before_send.go
sotnikov-s: please also add an error logging here

docs/static/swagger.yaml
jcompagni10: ah need to regen swagger

x/tokenfactory/migrations/v2/store.go
jcompagni10: Reposting from slack for others context:

So thinking about it more I'm not sure I totally agree with this. After the migration when we use this check in MsgServer it doesn't really matter why the hook isn't whitelisted (invalid contract, invalid denom, no whitelist value) we are always going to return an error. So we might as well use AssertIsHookWhitelisted to generate the errors so they are as informative as possible. If it returns false, nil for the case where there are no "errors" but the hook is not whitelisted then the message server  code has to be
```
whitelisted, err := server.Keeper.AssertIsHookWhitelisted(ctx, msg.Denom, cwAddr)
		if err != nil {
			return nil, err
		}
		if !whitelisted {
			return nil, types.ErrBeforeSendHookNotWhitelisted
		}
```
Which feels messy. Also, we no longer have the context to return an informative error message like we do here:  https://github.com/neutron-org/neutron/blob/a6eae66545df6e128da2c7cac81ff4c1c02deb7b/x/tokenfactory/keeper/params.go#L51
In the case of the migration, I agree that it's not quite perfect. but given that we are migrating a small set of specific hooks I think we can fairly assume that those are correct. If there are other hooks that throw an error as part of the check we do not want to stop the migration. For example, if someone added a hook before the upgrade the pointed to a non-existent contract we wouldn't want the migration to fail.
Let me know what you think?


I will deal with the array case better.

jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/9606704584

jcompagni10:
testrun with all PRs: https://github.com/neutron-org/neutron-tests/actions/runs/9606960942

jcompagni10:
Testrun with just neutron changes: https://github.com/neutron-org/neutron-tests/actions/runs/9607413347

x/tokenfactory/migrations/v2/store.go
sotnikov-s: okay, I see. what do you think about the following approach then?
1. make the errors returned from `AssertIsHookWhitelisted` more specific in terms of types:
image


2. remove hooks only in case of the `types.ErrBeforeSendHookNotWhitelisted error`, otherwise fail:
image

I think it's a quite good enough alternative here

x/tokenfactory/migrations/v2/store.go
```diff
@@ -0,0 +1,128 @@
...
+	ctx.Logger().Info("Migrating tokenfactory hooks...")
+
+	// get hook store
+	store := prefix.NewStore(ctx.KVStore(storeKey), []byte(types.DenomsPrefixKey))
+	iterator := storetypes.KVStorePrefixIterator(store, []byte{})
+	hooksToRemove := make([][]byte, 0)
+	for ; iterator.Valid(); iterator.Next() {
+		keyParts := strings.Split(string(iterator.Key()), types.KeySeparator)
+		if len(keyParts) != 3 {
+			return errors.New("cannot parse BeforeSendHook data")
```
sotnikov-s: please add the keyparts print here so the error message is comprehensive

x/tokenfactory/keeper/before_send.go
```diff
@@ -99,6 +120,20 @@ func (k Keeper) callBeforeSendListener(ctx context.Context, from, to sdk.AccAddr
...
 
+			// Do not invoke hook if denom is not whitelisted
+			// NOTE: hooks must already be whitelisted before they can be added, so under normal operation this check should never fail.
+			// It is here as an emergency override if we want to shutoff a hook. We do not return the error because once it is removed from the whitelist
+			// a hook should not be able to block a send.
+			if err := k.AssertIsHookWhitelisted(c, coin.Denom, cwAddr); err != nil {
+				c.Logger().Error(
+					"Skipped hook execution due to missing whitelist",
+					"err", err,
+					"denom", coin.Amount,
```
sotnikov-s:
```suggestion
					"denom", coin.Denom,
```

x/tokenfactory/migrations/v2/store.go
```diff
@@ -0,0 +1,128 @@
...
+		}
+
+		// Hooks and authorityMetadata are in the same store, we only care about the hooks
+		if keyParts[2] == types.BeforeSendHookAddressPrefixKey {
+			denom := keyParts[1]
+			contractAddr, err := sdk.AccAddressFromBech32(string(iterator.Value()))
+			if err != nil {
+				return errors.New("cannot parse hook contract address")
```
sotnikov-s: please add the contractAddr print here so the error message is comprehensive

x/tokenfactory/types/params.go
```diff
@@ -54,6 +57,10 @@ func (p Params) Validate() error {
 		return fmt.Errorf("failed to validate params: %w", err)
 	}
 
+	if err := validateWhitelistedHooks(p.WhitelistedHooks); err != nil {
+		return fmt.Errorf("failed to validate params: %w", err)
```
sotnikov-s:
```suggestion
		return fmt.Errorf("failed to validate whitelisted hooks: %w", err)
```

x/tokenfactory/migrations/v2/store.go
jcompagni10: The solution you describe above would fail for any hooks where the current contract is not valid. I believe there is at least one such contract right now.

jcompagni10:
Testrun with just neutron changes:https://github.com/neutron-org/neutron-tests/actions/runs/9608462857

sotnikov-s (approved):

pr0n00gler (approved):

zavgorodnii (approved):

PR #590, opened 2024-06-20
Fix: import/export tokenfactory hooks

pr0n00gler:
https://hadronlabs.atlassian.net/browse/NTRN-331


jcompagni10 (commented):
Probably need validation here: 
https://github.com/neutron-org/neutron/blob/fix%2Fimport-export-tokenfactory-hooks/x/tokenfactory/types/genesis.go#L21

x/tokenfactory/keeper/genesis.go
```diff
@@ -24,14 +24,26 @@ func (k Keeper) InitGenesis(ctx sdk.Context, genState types.GenesisState) {
...
 		if err != nil {
 			panic(err)
 		}
+
 		err = k.setAuthorityMetadata(ctx, genDenom.GetDenom(), genDenom.GetAuthorityMetadata())
 		if err != nil {
 			panic(err)
 		}
+
+		if _, err := sdk.AccAddressFromBech32(genDenom.HookContractAddress); genDenom.HookContractAddress != "" && err != nil {
```
swelf19, line 38: i think it's a redundant check, cause we have the address validation in `setBeforeSendHook` method

swelf19 (approved):

x/tokenfactory/keeper/genesis.go
mmikev: I don't think we need any invalid addresses in the state and this method prevents such case

PR #597, opened 2024-06-24
feat: wasmbinding schema #ntrn-327

NeverHappened:
Generates json schema for wasmbinding types

https://hadronlabs.atlassian.net/browse/NTRN-327


NeverHappened:
not needed

PR #611, opened 2024-07-03
Chore: split dex execution from banking operations #NTRN-353

jcompagni10:
**TASK**
https://hadronlabs.atlassian.net/browse/NTRN-353

This PR creates an execute method for each dex operation which performs all the core logic except event emission and banking operations. 
This makes performing simulations (which do not want banking operations) much simpler.  
It is a prerequisite for this PR: https://github.com/neutron-org/neutron/pull/543/files
It also begins a larger process of breaking down all of the core.go methods into more manageable (and readable) functions. 

This PR also splits each of the dex messages into its own dedicated file for better readability and easier refactoring in the future. 

This PR does not change any logic. It is largely just moving code around with a couple of small changes to event handling so that the ExecuteXX functions do not emit events. 

**Integration Tests**

https://github.com/neutron-org/neutron-tests/actions/runs/9896108148


x/dex/keeper/withdraw.go
```diff
@@ -0,0 +1,137 @@
...
+				sharesToRemove,
+				poolDenom,
+			)
+		}
+
+		outAmount0, outAmount1 := pool.Withdraw(sharesToRemove, totalShares)
+		k.SetPool(ctx, pool)
+
+		if sharesToRemove.IsPositive() {
+			if err := k.BurnShares(ctx, callerAddr, sharesToRemove, poolDenom); err != nil {
```
swelf19: There is still a bank operation. On the other hand there is no such thing like `QueryEstimateWithdraw`, so may be it's ok. But the top level comment is confusing

jcompagni10: fixed

x/dex/keeper/multihop_swap.go
```diff
@@ -17,6 +19,117 @@ type MultihopStep struct {
...
+		bestRoute.route,
+		bestRoute.dust,
+	))
+
+	return bestRoute.coinOut, nil
+}
+
+// CalulateMultiHopSwap handles the core logic for MultiHopSwap -- simulating swap operations across all routes (when applicable)
+// and picking the best route to execute. It uses a cache and does not modify state.
+func (k Keeper) CalulateMultiHopSwap(
```
swelf19, line 85: Any reason, why we can not rename other Execute to Calculate?

jcompagni10: It's not really a "calculate" since it's still making state changes. I would love to find a way to re-write all of them so that are non-mutative, but I haven't found an efficient way to do it.

swelf19: Good point

swelf19 (approved):

jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/9896108148

NicholasDotSol (approved):

jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/9985153021

pr0n00gler:
https://github.com/neutron-org/neutron-tests/actions/runs/9988527820

PR #613, opened 2024-07-04
Chore: update chain manager contract

pr0n00gler:
Chain manager contract must be updated according to this PR: https://github.com/neutron-org/neutron-dao/pull/107


pr0n00gler:
https://github.com/neutron-org/neutron-tests/actions/runs/9794556724

PR #620, opened 2024-07-11
fix: fix add-consumer-section for v4

NeverHappened:
wasn't working after upgrade to v4 - because of the new check on validating genesis


NeverHappened:
solution in rehearsal is cleaner

PR #623, opened 2024-07-11
Fix failed tx search query due to Slinky tx in each block

pr0n00gler:


pr0n00gler:
https://github.com/neutron-org/neutron-tests/actions/runs/9895619343

pr0n00gler:
https://github.com/neutron-org/neutron-tests/actions/runs/9895619343

PR #624, opened 2024-07-11
Chore: Flip Dex Price Logic #NTRN-355

jcompagni10:
**Issue**
https://hadronlabs.atlassian.net/browse/NTRN-355?atlOrigin=eyJpIjoiYTI0ODY0YTRhNzdhNDQ1NWFlODZhNzE0N2E0ZDk2MmYiLCJwIjoiaiJ9

Currently the dex stores prices as 1.0001^-TickIndex
This has been a persistent confusion for builders and integrators and also introduces unnecessary loss of precision. We can safely flip prices the other way without changing functionality


jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/10189392992

jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/10254298350

jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/10256440025

x/dex/migrations/v4/store.go
mmikev, line 1: Hm, could you clarify why do you change a code for the previous migration?

I don't think we are supposed to change old migrations

pr0n00gler (changes requested):

x/dex/migrations/v4/store.go
jcompagni10: `MustCalcPrice` fn has been inverted so to achieve the same migration we must flip the sign. The output of the migration should be the same

jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/10256440025

pr0n00gler:
Unit tests don't work, please fix it

x/dex/keeper/limit_order_tranche.go
```diff
@@ -29,7 +29,7 @@ func NewLimitOrderTranche(
 		TotalMakerDenom:    math.ZeroInt(),
 		TotalTakerDenom:    math.ZeroInt(),
 		ExpirationTime:     goodTil,
-		PriceTakerToMaker:  priceTakerToMaker,
+		MakerPrice:         price,
```
swelf19, line 33: Does it mean `PriceTakerToMaker` is unused now? Lets it remove it then.

proto/neutron/dex/pool_reserves.proto
```diff
@@ -26,10 +26,11 @@ message PoolReserves {
...
   ];
-  string price_opposite_taker_to_maker = 4 [
-    (gogoproto.moretags) = "yaml:\"price_opposite_taker_to_maker\"",
+  // This is the price of the PoolReserves denominated in the opposite token. (ie. 1 TokenA with a maker_price of 10 is worth 10 TokenB )
+  string maker_price = 5 [
+    (gogoproto.moretags) = "yaml:\"maker_price\"",
     (gogoproto.customtype) = "github.com/neutron-org/neutron/v4/utils/math.PrecDec",
     (gogoproto.nullable) = false,
-    (gogoproto.jsontag) = "price_opposite_taker_to_maker"
+    (gogoproto.jsontag) = "maker_price"
```
swelf19, line 45: it's a matter of taste, but i like name `price_maker_to_taker` better as an opposite price

x/dex/migrations/v4/store.go
swelf19: I think the question was - do we get the same migration output in 2 cases
1) Our case. User migrates 3 -> 4 (old one) (neutron v4.2.1) and then 4 -> 5. **3 -> old 4 -> 5**
2) Imported dex module to third parthy project. User migrates 3 vesrion -> 5 version (current one).  **3 -> new 4 -> 5**

I dont think it's safe in terms compatibility to modify already released migrations

mmikev: Yes, that's exactly what i meant

proto/neutron/dex/pool_reserves.proto
jcompagni10: Just to clarify. The plan is to remove the field `price_taker_to_maker` in the following release. I agree that if we were keeping both they should be named similarly. Personally, I have always, found `price_taker_to_maker` confusing. It is ambiguous whether it is referring to the price the maker will pay for the taker denom or the rate at which maker denoms can be exchanged for taker denoms. IMHO `maker_price` is a clearer description of the price at which the maker is selling a token. 
That being said, a primary goal of this PR is to eliminate confusion around naming conventions, so if the consensus is that `price_maker_to_taker` is clearer I am happy to go with that. 

@swelf19 @pr0n00gler @NicholasDotSol  Thoughts here?

x/dex/keeper/limit_order_tranche.go
jcompagni10: The plan is to remove it. But for compatibility, I think its better to deprecate first and then remove in the subsequent release.

x/dex/migrations/v4/store.go
jcompagni10: I don't think it's an issue. You can see that the tests are unchanged. So the migrations still produce the same output. @swelf19 @pr0n00gler

jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/10708235338

jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/10949752474

jcompagni10:
I'm assuming this: 
> I think the question was - do we get the same migration output in 2 cases
Our case. User migrates 3 -> 4 (old one) (neutron v4.2.1) and then 4 -> 5. 3 -> old 4 -> 5
Imported dex module to third parthy project. User migrates 3 vesrion -> 5 version (current one). 3 -> new 4 -> 5

Was resolved by the comment above @swelf19, please correct me if not

jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/11021574416

proto/neutron/dex/pool_reserves.proto
NicholasDotSol: agree, it only sounds more descriptive on surface level, in fact it's ambiguous

x/dex/keeper/liquidity.go
```diff
@@ -56,7 +56,7 @@ func (k Keeper) Swap(
 		// but due to rounding and inaccuracy of fixed decimal math, it is possible
```
NicholasDotSol, line 56: Should update the comment on ln55

x/dex/types/price.go
```diff
@@ -96,7 +96,7 @@ func CalcTickIndexFromPrice(price math_utils.PrecDec) (int64, error) {
 		return 0, ErrPriceOutsideRange
 	}
 
-	if price.GT(math_utils.OnePrecDec()) {
+	if price.LT(math_utils.OnePrecDec()) {
```
swelf19, line 99: As far as i understand you need to update comment at line 100 -> `>=1`

x/dex/migrations/v4/store.go
swelf19: Yes, seems you are right. No changes in output. Dont see any problems with a migration now.

swelf19 (dismissed):
Pls resolve conflicts

jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/11021574416

swelf19 (approved):

PR #625, opened 2024-07-11
chore: Remove old dex migrations

jcompagni10:
Remove old dex migrations.

There are multiple incoming PRs (ie. https://github.com/neutron-org/neutron/pull/624) that conflict with these migrations. Since they are no longer necessary we can safely remove them.

**Integration Test Run** 
https://github.com/neutron-org/neutron-tests/actions/runs/9897237914


pr0n00gler (dismissed):

joldie777 (approved):

quasisamurai (approved):

pr0n00gler (changes requested):
After a couple of considerations we think it's not safe to remove migrations from the module. Module is a separate entity and if i import it as a third-part developer i must get a guarantee that all changes made by developer in the code also reflected in the state by migrations. What do you think?

PR #646, opened 2024-07-31
feat: refactor ICQ module documentation [NTRN-349]

sotnikov-s:
### task
https://hadronlabs.atlassian.net/browse/NTRN-349

### this PR
- covers all `.proto` and `sudo` definitions of the `interchainqueries` module with comments aiming to make use of the descriptions via links to code based documentation (e.g. [QueryRegisteredQueriesRequest](https://pkg.go.dev/github.com/neutron-org/neutron/v4@v4.0.1/x/interchainqueries/types#QueryRegisteredQueriesRequest))

### related PRs
- https://github.com/neutron-org/neutron-docs/pull/219
- https://github.com/neutron-org/neutron-dev-contracts/pull/71
- https://github.com/neutron-org/neutron-integration-tests/pull/367
- https://github.com/neutron-org/neutron-std/pull/7
- https://github.com/neutron-org/neutron-sdk/pull/168


sotnikov-s:
https://github.com/neutron-org/neutron-tests/actions/runs/11836995682

NeverHappened (dismissed):

proto/neutron/interchainqueries/genesis.proto
```diff
@@ -8,57 +8,59 @@ import "neutron/interchainqueries/params.proto";
...
-  // Amount of coins deposited for the query.
+  // Amount of coins paid for the Interchain Query registration. The deposit is paid back to the
+  // remover. The remover can be either the query owner (during the submit timeout) or anybody.
   repeated cosmos.base.v1beta1.Coin deposit = 10 [
     (gogoproto.castrepeated) = "github.com/cosmos/cosmos-sdk/types.Coins",
     (gogoproto.nullable) = false
   ];
-
-  // Timeout before query becomes available for everybody to remove.
+  // Duration in blocks that is required to pass since the query registration/update for the
```
swelf19: What the 'update' word means here. How to keep query alive

proto/neutron/interchainqueries/params.proto
```diff
@@ -6,22 +6,19 @@ import "gogoproto/gogo.proto";
...
 
-// Params defines the parameters for the module.
+// The parameters for the module.
 message Params {
   option (gogoproto.goproto_stringer) = false;
-  // Defines amount of blocks required before query becomes available for
-  // removal by anybody
+  // The amount of blocks required to pass since an Interchain Query registration/update for the
+  // query to become available for removal by anybody.
   uint64 query_submit_timeout = 1;
```
swelf19, line 15: How is this param related to `submit_timeout` of an interchain query

proto/neutron/interchainqueries/query.proto
```diff
@@ -10,65 +10,85 @@ import "neutron/interchainqueries/tx.proto";
...
 
-// QueryParamsRequest is request type for the Query/Params RPC method.
+// Request type for the Query/Params RPC method.
 message QueryParamsRequest {}
 
-// QueryParamsResponse is response type for the Query/Params RPC method.
+// Response type for the Query/Params RPC method.
 message QueryParamsResponse {
-  // params holds all the parameters of this module.
+  // Stores all parameters of the module.
```
swelf19: Is the `stores` word correct for this certain usecase?

proto/neutron/interchainqueries/tx.proto
```diff
@@ -13,147 +13,191 @@ import "tendermint/crypto/proof.proto";
...
+  // The second half of the storage path. The remaining part of the full path to an IAVL storage node.
   bytes key = 2;
-
-  // is the value in IAVL store
+  // A base64-encoded value read from the given storage path.
   bytes value = 3;
-
-  // is the Merkle Proof which proves existence of key-value pair in IAVL
-  // storage
+  // The Merkle Proof which proves existence of key-value pair in IAVL storage. Is used to verify
```
swelf19: Or proves nonexistance if no key-value pair in IAVL storage, not sure is it obvious from the comment.

proto/neutron/interchainqueries/genesis.proto
sotnikov-s: yeah, that's a good catch. especially given the ambiguity behind the 'update' term which in the API means the query parameters update. will fix

proto/neutron/interchainqueries/params.proto
sotnikov-s: will rephrase, too

proto/neutron/interchainqueries/genesis.proto
```diff
@@ -8,57 +8,59 @@ import "neutron/interchainqueries/params.proto";
...
 
+// A path to an IAVL storage node.
 message KVKey {
-  // Path (storage prefix) to the storage where you want to read value by key
-  // (usually name of cosmos-sdk module: 'staking', 'bank', etc.)
+  // The first half of the storage path. It is supposed to be a substore name for the query
+  // (e.g. bank, staking, etc.).
   string path = 1;
-  // Key you want to read from the storage
+  // The second half of the storage path. The remaining part of the full path to an IAVL storage node.
```
mmikev: I personally don't quite like `halves` descriptions here.
I think it's much more easier to understand when we operate _storage names_ (it's literally module names) and _keys_ (what specific item you want to read from the module)

proto/neutron/interchainqueries/query.proto
```diff
@@ -10,65 +10,87 @@ import "neutron/interchainqueries/tx.proto";
...
 }
 
+// Response type for the Query/RegisteredQueries RPC method.
 message QueryRegisteredQueriesResponse {
+  // A list of registered Interchain Queries.
   repeated RegisteredQuery registered_queries = 1 [(gogoproto.nullable) = false];
-
-  // pagination defines the pagination in the response.
+  // Current page information. Use values from previous response in the next request in consecutive
+  // requests with paginated responses.
```
mmikev, line 69:
```suggestion
  // Current page information. Use values from this response in the next request in consecutive
  // requests with paginated responses.
```

> in the next request in consecutive requests

This looks like a copypasted typo

proto/neutron/interchainqueries/tx.proto
```diff
@@ -13,147 +13,194 @@ import "tendermint/crypto/proof.proto";
...
+  // The first half of the storage path. It is supposed to be a substore name for the query
+  // (e.g. bank, staking, etc.).
   string storage_prefix = 1;
-
-  // is the key in IAVL store
+  // The second half of the storage path. The remaining part of the full path to an IAVL storage node.
   bytes key = 2;
-
-  // is the value in IAVL store
+  // A base64-encoded value read from the given storage path.
```
mmikev: This is a bit misleading i think.
The value is not base64-encoded, it's just `bytes` field. Base64 encoding happens when you try to convert this into some human-readable format (JSON, etc). In that way `bytes` fields will be represented as base64 string field.

But your comment implies that the `value` field itself is base64 encoded + it has `bytes` type

proto/neutron/interchainqueries/genesis.proto
sotnikov-s: actually these are not module names, just any higher level keys to ease storage navigation. for example, `auth` module has AccountKeeper which has `acc` as its store key. but I got your point, will rephrase

proto/neutron/interchainqueries/query.proto
sotnikov-s: no, that's alright:

re replacing `previous` to `this`: that's a general description of pagination usage, not for a particular response

re copy-paste typo: no typo. again, that's just the way how users are supposed to treat/use pagination generally

proto/neutron/interchainqueries/genesis.proto
```diff
@@ -8,57 +8,59 @@ import "neutron/interchainqueries/params.proto";
...
 
+// Represents a path to an IAVL storage node.
 message KVKey {
-  // Path (storage prefix) to the storage where you want to read value by key
-  // (usually name of cosmos-sdk module: 'staking', 'bank', etc.)
+  // The substore name used in an Interchain Query. Typically, this corresponds to the keeper's
+  // storeKey, usually the module's name, such as "bank", "staking", etc.
   string path = 1;
-  // Key you want to read from the storage
+  // A hexadecimal-encoded byte array representing the key for specific data in the module's storage.
```
mmikev: The same about _hexadecimal_ encoding: this is correct for only for one particular ABCI query, but here it's just bytes

pr0n00gler (changes requested):

pr0n00gler (dismissed):

x/contractmanager/types/sudo.go
```diff
@@ -5,20 +5,33 @@ import (
...
+		// Data is the body of the transaction.
+		Data []byte `json:"data"`
 	} `json:"tx_query_result"`
 }
 
-// MessageKVQueryResult is passed to a contract's sudo() entrypoint when a result
-// was submitted for a kv-query.
+// MessageKvQueryResult is the model of the `sudo` message sent to a smart contract when a KV
+// Interchain Query result is submitted. If the owner of a KV Interchain Query wants to handle
+// updates, they must implement a `sudo` entry point to process `kv_query_result` messages and
```
NeverHappened: Maybe `handle updates right away` or `handle updates on submitting results` to avoid ambiguity?
Since we can query info later

NeverHappened (dismissed):

pr0n00gler (approved):

NeverHappened (approved):

joldie777 (approved):

PR #669, opened 2024-08-13
feat: async pruning enabled

swelf19:
Updated cosmos-sdk to enable async pruning: https://github.com/neutron-org/cosmos-sdk/pull/39


pr0n00gler:
https://github.com/neutron-org/neutron-tests/actions/runs/10527141661

PR #671, opened 2024-08-14
feat: added execution stage for schedules #NTRN-339

joldie777:
Task: https://hadronlabs.atlassian.net/browse/NTRN-339

Related PRs:
https://github.com/neutron-org/neutron-sdk/pull/161


wasmbinding/bindings/msg.go
```diff
@@ -195,9 +195,10 @@ type ForceTransfer struct {
 
 // AddSchedule adds new schedule to the cron module
 type AddSchedule struct {
-	Name   string               `json:"name"`
-	Period uint64               `json:"period"`
-	Msgs   []MsgExecuteContract `json:"msgs"`
+	Name    string               `json:"name"`
+	Period  uint64               `json:"period"`
+	Msgs    []MsgExecuteContract `json:"msgs"`
+	Blocker uint64               `json:"blocker"`
```
NeverHappened: how about changing type to string? it's gonna enable rust side to understand the meaning of this variable without mapping onto number

x/cron/keeper/grpc_query_schedule_test.go
```diff
@@ -134,7 +134,7 @@ func createNSchedule(t *testing.T, ctx sdk.Context, k *cronkeeper.Keeper, n int3
 		item.Msgs = nil
 		item.LastExecuteHeight = uint64(ctx.BlockHeight())
 
-		err := k.AddSchedule(ctx, item.Name, item.Period, item.Msgs)
+		err := k.AddSchedule(ctx, item.Name, item.Period, item.Msgs, uint64(item.Blocker))
```
NeverHappened: Casting to uint64 looks a bit weird.
Can we just make AddSchedule last arg of `BlockerType` instead of uint64?

proto/neutron/cron/schedule.proto
```diff
@@ -5,6 +5,13 @@ import "gogoproto/gogo.proto";
 
 option go_package = "github.com/neutron-org/neutron/v4/x/cron/types";
 
+// BlockerType defines when messages will be executed in the block
+enum BlockerType {
```
NeverHappened: Need to think of a better name. How about:
```go
enum ExecutionStage {
  BEGIN_BLOCKER = 0;
  END_BLOCKER = 1;
  BOTH_BLOCKERS = 2;
}
```

x/cron/keeper/keeper.go
```diff
@@ -67,19 +67,32 @@ func (k *Keeper) Logger(ctx sdk.Context) log.Logger {
 // ExecuteReadySchedules gets all schedules that are due for execution (with limit that is equal to Params.Limit)
 // and executes messages in each one
 // NOTE that errors in contract calls rollback all already executed messages
-func (k *Keeper) ExecuteReadySchedules(ctx sdk.Context) {
+func (k *Keeper) ExecuteReadySchedules(ctx sdk.Context, isBeginBlocker bool) {
```
NeverHappened: Looks like `isBeginBlocker` arg can be or type `Blocker` (until renamed of course). That way we can directly compare schedule blocker param with the passed arg

NeverHappened: Also there are two checks but can be only one: 
How about refactoring it something like this (NOTE: not tested!)
```go
func (k *Keeper) ExecuteReadySchedules(ctx sdk.Context, currentStage types.BlockerType) {
	telemetry.ModuleMeasureSince(types.ModuleName, time.Now(), LabelExecuteReadySchedules)
	schedules := k.getSchedulesReadyForExecution(ctx)

	for _, schedule := range schedules {
		if IsExecutableAtThisStage(schedule, currentStage) {
			err := k.executeSchedule(ctx, schedule)
			recordExecutedSchedule(err, schedule)
		}
	}
}

func IsExecutableAtThisStage(schedule types.Schedule, currentStage types.BlockerType) bool {
	return schedule.Blocker == types.BlockerType_BOTH || schedule.Blocker == currentStage
}

```

NeverHappened (changes requested):

wasmbinding/bindings/msg.go
joldie777: Done

x/cron/keeper/grpc_query_schedule_test.go
joldie777: Done

proto/neutron/cron/schedule.proto
joldie777: Done

x/cron/keeper/keeper.go
joldie777: Great point! Done

proto/neutron/cron/tx.proto
```diff
@@ -15,10 +16,48 @@ option go_package = "github.com/neutron-org/neutron/v4/x/cron/types";
...
+  // Removes schedule.
+  rpc RemoveSchedule(MsgRemoveSchedule) returns (MsgRemoveScheduleResponse);
+  // Updates the module parameters.
   rpc UpdateParams(MsgUpdateParams) returns (MsgUpdateParamsResponse);
   // this line is used by starport scaffolding # proto/tx/rpc
 }
 
+// MsgAddSchedule is the MsgAddSchedule request type.
+message MsgAddSchedule {
+  option (amino.name) = "cron/MsgAddSchedule";
```
NeverHappened, line 30: what's this for? I mean I don't know :)

NeverHappened (dismissed):

proto/neutron/cron/tx.proto
joldie777: Me neither :D
I saw this option being used everywhere and just copy-pasted it.

proto/neutron/cron/tx.proto
```diff
@@ -15,10 +16,48 @@ option go_package = "github.com/neutron-org/neutron/v4/x/cron/types";
...
+  option (cosmos.msg.v1.signer) = "authority";
+
+  // Authority is the address of the governance account.
+  string authority = 1 [(cosmos_proto.scalar) = "cosmos.AddressString"];
+
+  string name = 2;
+  uint64 period = 3;
+  repeated MsgExecuteContract msgs = 4 [(gogoproto.nullable) = false];
+  ExecutionStage execution_stage = 5;
```
sotnikov-s: please cover everything possible in proto files with comprehensive comments. that is important for understanding for both developers and users

proto/neutron/cron/schedule.proto
```diff
@@ -5,6 +5,13 @@ import "gogoproto/gogo.proto";
 
 option go_package = "github.com/neutron-org/neutron/v4/x/cron/types";
 
+// ExecutionStage defines when messages will be executed in the block
+enum ExecutionStage {
+  BEGIN_BLOCKER = 0;
+  END_BLOCKER = 1;
+  BOTH_BLOCKERS = 2;
+}
+
```
sotnikov-s: according to proto definition best practices, it's recommended to prefix each value of a enum with the enum name, e.g. EXEC_STAGE_BEGIN_BLOCKER, EXEC_STAGE_END_BLOCKER, etc. for example, here's an ibc-go piece of code: https://github.com/cosmos/ibc-go/blob/69f98e507203fff431248264edf97fc553f52eb6/proto/ibc/core/channel/v1/channel.proto#L57-L78

sotnikov-s: the `BOTH_BLOCKERS` value strictly limits us in terms of possible cron module's functionality extension. what if we decide to add another execution point? what would the `both` mean in this case? I think we can consider
1. making the `message Schedule`'s `execution_stage` field an array of execution stages, e.g. `[end_blocker]`, `[begin_blocker]`, `[begin_blocker, end_blocker]`
2. removing the `both_blockers` value from the enum and use a combination of the other two: `[begin_blocker, end_blocker]`

one more thing is please think about how we can maintain a good performance when working with the `execution_stage` arrays. if just keep it is an array in the state, it would lead to O(n) calculation done twice for each schedule each block. and if we add more execution points in the future, the number of O(n) will increase

proto/neutron/cron/v1/schedule.proto
```diff
@@ -0,0 +1,29 @@
...
+
+option go_package = "github.com/neutron-org/neutron/v4/x/cron/types/v1";
+
+message Schedule {
+  // Name of schedule
+  string name = 1;
+  // Period in blocks
+  uint64 period = 2;
+  // Msgs that will be executed every period amount of time
+  repeated MsgExecuteContract msgs = 3 [(gogoproto.nullable) = false];
```
sotnikov-s: > every period amount of time

that's confusing. could you please clarify/rephrase that it's about blocks gap between sequential calls? here and in the current `schedule.proto`

wasmbinding/message_plugin.go
```diff
@@ -985,7 +985,15 @@ func (m *CustomMessenger) addSchedule(ctx sdk.Context, contractAddr sdk.AccAddre
...
 	}
 
-	err := m.CronKeeper.AddSchedule(ctx, addSchedule.Name, addSchedule.Period, msgs)
+	var executionStage crontypes.ExecutionStage
+
+	if v, ok := crontypes.ExecutionStage_value[addSchedule.ExecutionStage]; !ok {
+		executionStage = crontypes.ExecutionStage_END_BLOCKER
+	} else {
+		executionStage = crontypes.ExecutionStage(v)
+	}
```
sotnikov-s: do we even need that? why don't enforce stage being set? is that for backwards compatibility? if yes, with what? also, as far as I'm concerned, you can simply make the `ExecutionStage_END_BLOCKER` the first value in the enum (i.e. set it the value 0) and it will be the default value in case of explicit parameter absence

x/cron/keeper/keeper.go
```diff
@@ -67,19 +67,27 @@ func (k *Keeper) Logger(ctx sdk.Context) log.Logger {
 // ExecuteReadySchedules gets all schedules that are due for execution (with limit that is equal to Params.Limit)
 // and executes messages in each one
 // NOTE that errors in contract calls rollback all already executed messages
```
sotnikov-s: do we even need this comment line? that's a pretty customary behaviour when dealing with smart contracts, whereas here that huge NOTE looks like it's some peculiarity. it would be better if it was REMEMBER, not NOTE, but still I think it's better to remove the line for clarity and brevity

x/cron/keeper/keeper.go
```diff
@@ -67,19 +67,27 @@ func (k *Keeper) Logger(ctx sdk.Context) log.Logger {
...
-		err := k.executeSchedule(ctx, schedule)
-		recordExecutedSchedule(err, schedule)
+		if isExecutableAtTheStage(schedule, executionStage) {
+			err := k.executeSchedule(ctx, schedule)
+			recordExecutedSchedule(err, schedule)
+		}
 	}
 }
 
 // AddSchedule adds new schedule to execution for every block `period`.
```
sotnikov-s: > for every block `period`.

that's also confusing. could you please rephrase it similarly to the other rephrasing I've asked you about? :)

sotnikov-s (commented):
to be continued...

x/cron/keeper/msg_server.go
```diff
@@ -22,8 +22,44 @@ func NewMsgServerImpl(keeper Keeper) types.MsgServer {
...
+	}
+
+	authority := k.GetAuthority()
+	if authority != req.Authority {
+		return nil, errors.Wrapf(sdkerrors.ErrInvalidRequest, "invalid authority; expected %s, got %s", authority, req.Authority)
+	}
+
+	ctx := sdk.UnwrapSDKContext(goCtx)
+	if err := k.Keeper.AddSchedule(ctx, req.Name, req.Period, req.Msgs, req.ExecutionStage); err != nil {
+		return nil, err
```
sotnikov-s: please wrap the error

proto/neutron/cron/tx.proto
joldie777: Done for proto files in Cron module

proto/neutron/cron/schedule.proto
joldie777: Done

joldie777: Done, but not sure if adding `schedulesExecutionStages` map is the right way to maintain a good performance. This needs to be discussed.

proto/neutron/cron/v1/schedule.proto
joldie777: Please check if the new phrase is acceptable.

wasmbinding/message_plugin.go
joldie777: Remade

x/cron/keeper/keeper.go
joldie777: Removed

x/cron/keeper/keeper.go
joldie777: I've tried to rephrase

joldie777: Done

x/cron/keeper/keeper.go
```diff
@@ -27,6 +27,8 @@ var (
 
 	MetricLabelSuccess      = "success"
 	MetricLabelScheduleName = "schedule_name"
+
+	schedulesExecutionStages map[string]map[types.ExecutionStage]struct{}
```
sotnikov-s: please refactor to Keeper.Store usage based on our discussion

sotnikov-s (changes requested):

x/cron/keeper/keeper.go
joldie777: We've discussed this with Mike and decided to forego the option of executing schedule in several stages.

x/cron/keeper/msg_server.go
joldie777: Done

wasmbinding/message_plugin.go
```diff
@@ -985,7 +985,7 @@ func (m *CustomMessenger) addSchedule(ctx sdk.Context, contractAddr sdk.AccAddre
 		})
 	}
 
-	err := m.CronKeeper.AddSchedule(ctx, addSchedule.Name, addSchedule.Period, msgs)
+	err := m.CronKeeper.AddSchedule(ctx, addSchedule.Name, addSchedule.Period, msgs, crontypes.ExecutionStage(crontypes.ExecutionStage_value[addSchedule.ExecutionStage]))
```
sotnikov-s: bindings handler shouldn't access Keepers directly, the interaction should be done via module's message server. otherwise we skip the validation step which is done on the message server's layer. that's not your fault but ab old bug. could you please replace cron keeper direct usage with its message server for all occasions in this file?

x/cron/keeper/keeper.go
```diff
@@ -89,7 +94,13 @@ func (k *Keeper) AddSchedule(ctx sdk.Context, name string, period uint64, msgs [
 		Period:            period,
 		Msgs:              msgs,
 		LastExecuteHeight: uint64(ctx.BlockHeight()), // let's execute newly added schedule on `now + period` block
+		ExecutionStage:    executionStage,
 	}
+
+	if _, ok := types.ExecutionStage_name[int32(executionStage)]; !ok {
+		schedule.ExecutionStage = types.ExecutionStage_EXECUTION_STAGE_END_BLOCKER
+	}
```
sotnikov-s: discussed with @pr0n00gler — we don't need a default value for exec stage, and we better explicitly require the passed value to be in the list, otherwise return an error. Please add this kind of validation to the Validate method

x/cron/migrations/v2/store_test.go
```diff
@@ -0,0 +1,53 @@
...
+	store := prefix.NewStore(ctx.KVStore(storeKey), types.ScheduleKey)
+	bz := cdc.MustMarshal(&schedule)
+	store.Set(types.GetScheduleKey(schedule.Name), bz)
+
+	// Run migration
+	suite.NoError(v2.MigrateStore(ctx, cdc, storeKey))
+
+	// Check Schedule has correct ExecutionStage
+	newSchedule, _ := app.CronKeeper.GetSchedule(ctx, schedule.Name)
+	suite.Equal(newSchedule.ExecutionStage, types.ExecutionStage_EXECUTION_STAGE_END_BLOCKER)
```
sotnikov-s, line 59: please add a sanity check that other fields are left intact

sotnikov-s (changes requested):

wasmbinding/message_plugin.go
joldie777: Done

x/cron/keeper/keeper.go
joldie777: Done

x/cron/migrations/v2/store_test.go
joldie777: Done

wasmbinding/message_plugin.go
```diff
@@ -88,7 +89,8 @@ type CustomMessenger struct {
 	Adminserver           admintypes.MsgServer
 	Bank                  *bankkeeper.BaseKeeper
 	TokenFactory          *tokenfactorykeeper.Keeper
-	CronKeeper            *cronkeeper.Keeper
+	Cronmsgserver         crontypes.MsgServer
+	Cronqueryserver       crontypes.QueryServer
```
sotnikov-s: could you please rename these fields to camel case? CronMsgServer and CronQueryServer

wasmbinding/message_plugin.go
```diff
@@ -1000,12 +1010,29 @@ func (m *CustomMessenger) addSchedule(ctx sdk.Context, contractAddr sdk.AccAddre
...
+	authority := authtypes.NewModuleAddress(admintypes.ModuleName)
+
+	_, err = m.Cronmsgserver.RemoveSchedule(ctx, &crontypes.MsgRemoveSchedule{
+		Authority: authority.String(),
+		Name:      removeSchedule.Name,
+	})
+	if err != nil {
+		ctx.Logger().Error("failed to removeSchedule",
+			"from_address", contractAddr.String(),
+			"error", err,
```
sotnikov-s: would be good to have the schedule's `Name` as well here and in the returned error

x/cron/keeper/msg_server_test.go
```diff
@@ -6,12 +6,171 @@ import (
...
+			"authority is invalid",
+		},
+		{
+			"invalid authority",
+			types.MsgRemoveSchedule{
+				Authority: "",
+				Name:      "name",
+			},
+			"authority is invalid",
+		},
```
sotnikov-s: this test case is just the same as the previous one. I think you mean to provide
```go
Authority: "invalid authority",
```
here just like you did in the `TestMsgAddScheduleValidate`?

sotnikov-s (changes requested):

proto/neutron/cron/tx.proto
NeverHappened: okay this is just for amino encoding, doesn't matter much, let it stay:
https://github.com/cosmos/cosmos-sdk/blob/e8f28bf5db18b8d6b7e0d94b542ce4cf48fed9d6/proto/cosmos/bank/v1beta1/tx.proto#L41

x/cron/keeper/msg_server_test.go
joldie777: Correct! Thx!

wasmbinding/message_plugin.go
joldie777: Done

wasmbinding/message_plugin.go
joldie777: Done

sotnikov-s (dismissed):

NeverHappened (dismissed):

pr0n00gler (approved):

PR #680, opened 2024-08-19
fix: add admin signer

NeverHappened:
Bumps admin module in which:
added signer for the MsgAddAdmin and MsgDeleteAdmin.
This allows for sending these messages using stargate.


pr0n00gler:
https://github.com/neutron-org/neutron-tests/actions/runs/10456280537

pr0n00gler (approved):

PR #685, opened 2024-08-20
Autoswap accounting fixes #NTRN-385

jcompagni10:
This PR fixes 4 issues to related to how autoswap shares are calculated. For broader context see the autoswap spec: https://www.notion.so/hadron/1f17de6d999f4ace9f322262dde11a93?v=5bca99093f134baf9688c9bf12f1aa1c&p=7aca1c1dd1c54ae38202792b79cfe54e&pm=s

Issue 1: 
The price (`price1To0Center`) used for the calculation was incorrect. 
Issue 2: 
Autoswap shares were being issued at full value instead of being discounted by the normal share value calculation (existingShares / poolValue) 
Issue 2.5: 
The share value calculation must include the autoswap fee as part of the share value. 
Issue 3: 
Autoswap calculation does not take into account prices and handles 1 sides pools incorrectly, forcing autoswap amount to be too high. See: https://www.notion.so/Autoswap-Spec-ca5f35a4cd5b4dbf9ae27e0454ddd445?pvs=4#12032ea59b0e802c925efae10c3ca85f

Because of these issues, it is possible that an incorrect number of shares were issued for pool deposits. As of this moment, there are no pools that are affected. But out of caution, there is an upgrade that will withdraw any potentially affected pools.


pr0n00gler:
Is it ready for review? Cause the description says otherwise

jcompagni10:
> Is it ready for review? Cause the description says otherwise

Yes, ready for review

jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/10688522598

jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/10708528286

x/dex/keeper/pool.go
```diff
@@ -177,3 +177,20 @@ func (k Keeper) SetPoolCount(ctx sdk.Context, count uint64) {
...
 }
+
+func (k Keeper) GetAllPoolShareholders(ctx sdk.Context) map[uint64][]types.PoolShareholder {
+	result := make(map[uint64][]types.PoolShareholder)
+	balances := k.bankKeeper.GetAccountsBalances(ctx)
+	for _, balance := range balances {
+		for _, coin := range balance.Coins {
+			poolID, err := types.ParsePoolIDFromDenom(coin.Denom)
+			if err != nil {
+				continue
```
mmikev, line 178: Shouldn't we have a log or something to track errors?

x/dex/migrations/v5/store.go
mmikev, line 1: I wouldn't say is a migration per se.

I think this logic must be located in the upgrade handler for v5 upgrade

pr0n00gler (changes requested):

x/dex/keeper/pool.go
jcompagni10: Not really an error. We use ParsePoolIDFromDenom to check if the denom is a PoolShare or just another random coin. err in this case just means not a PoolShareDenom.

jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/10744249913

jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/11040046087

jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/10744249913

pr0n00gler (dismissed):

jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/11341156034

pr0n00gler (approved):

PR #659, opened 2024-08-08
FIX: LO cancellation accounting

jcompagni10:
Fix issue with how TotalTakerDenom is calculated after withdraw

PR #690, opened 2024-08-22
Fix: Limit order cancel accounting

jcompagni10:
Port over fix to limit order cancel account 

+ add more tests


jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/10516277754

pr0n00gler:
A copy of https://github.com/neutron-org/neutron/pull/659

PR #696, opened 2024-08-27
Feat: Add SimulateXX queries

jcompagni10:
Add queries to simulate all dex operations. Allows for easier CW interaction with the dex. 

Also added a couple extra fields to the normal Dex message responses 

This is a more comprehensive PR replacing the earlier version: https://github.com/neutron-org/neutron/pull/543 

**Related PRs**
https://github.com/neutron-org/neutron-sdk/pull/162
https://github.com/neutron-org/neutron-dev-contracts/pull/63
https://github.com/neutron-org/neutron-integration-tests/pull/344


jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/10677779510

x/dex/keeper/cancel_limit_order.go
```diff
@@ -30,23 +30,25 @@ func (k Keeper) CancelLimitOrderCore(
...
-		return err
+		return sdk.Coin{}, sdk.Coin{}, err
 	}
 
-	// This will never panic since TradePairID has already been successfully constructed by ExecuteCancelLimitOrder
-	pairID := tradePairID.MustPairID()
+	makerDenom := makerCoinOut.Denom
+	takerDenom := takerCoinOut.Denom
+	// This will never panic since TradePairID has already been successfully constructed by ExecuteWithdrawFilledLimitOrder
+	pairID := types.MustNewPairID(makerDenom, takerDenom)
```
swelf19: Outdated comment, there is no TradePairID anymore. And as we discussed, need to fix pair creation.

x/dex/types/keys.go
```diff
@@ -207,3 +207,6 @@ func JITGoodTilTime() time.Time {
 const (
 	ExpiringLimitOrderGas = 10_000
 )
+
+// Dummy Address used for simulate queries
+const DummyAddress = "neutron1pq7j6za5zjcl3um9t5gfyleues336tv04tyq0k"
```
swelf19, line 212: We do not simulate bank operation during estimate queries, do we still need a dummy address?

jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/10677779510

zavgorodnii:
please discuss with @pr0n00gler the wasmbindings situation

x/dex/types/keys.go
jcompagni10: Still required for message validation for MHS, deposit and place limit order. 
We could have a separate validation function for the simulate queries. But I prefer to re-use as much logic as possible.

jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/10677779510

x/dex/types/keys.go
swelf19: Validation only? Then let me rephrase the question, why do we want to fill the fields on behalf of the user? It's a simulation query and i assume user take care about address fields.

jcompagni10: The dummy address is only used for Deposit, PlaceLimitOrder, and MHS. For these cases, the address specified has no effect on the simulation. So it's just a convenience that it's one less field they need to specify.

swelf19 (dismissed):

pr0n00gler (approved):

x/dex/keeper/grpc_query_simulate_cancel_limit_order.go
```diff
@@ -0,0 +1,40 @@
...
+
+	"github.com/neutron-org/neutron/v4/x/dex/types"
+)
+
+func (k Keeper) SimulateCancelLimitOrder(
+	goCtx context.Context,
+	req *types.QuerySimulateCancelLimitOrderRequest,
+) (*types.QuerySimulateCancelLimitOrderResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+	cacheCtx, _ := ctx.CacheContext()
```
mmikev, line 16: An interesting question: does this uncommited cached context consume gas during execution in contracts?

In other words, is this simulation free for contracts during their execution?

jcompagni10: My understanding is that cacheCtx still consumes gas. So, yes, contract will pay the query gas cost

jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/10677779510

PR #698, opened 2024-08-29
Ensure that empty tranches are deleted 

jcompagni10:
There are some cases when a tranche is canceled where is not deleted and instead is saved as an inactive tranche.
This doesn't pose any immediate problem, but in the long term would result in unnecessary state bloat. 

this PR changes the k.saveTranche logic to ensure that empty tranches are fully deleted.


jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/10609901210

x/dex/keeper/limit_order_tranche.go
```diff
@@ -55,7 +55,7 @@ func (k Keeper) SaveTranche(ctx sdk.Context, tranche *types.LimitOrderTranche) {
 	if tranche.HasTokenIn() {
 		k.SetLimitOrderTranche(ctx, tranche)
 	} else {
-		k.SetInactiveLimitOrderTranche(ctx, tranche)
+		k.SaveInactiveTranche(ctx, tranche)
```
mmikev, line 58: It's counterintuitive that PR is about removing of old tranches, but the method name implies to save them instead.

Is there any way to change it?

pr0n00gler (changes requested):

jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/10609901210

pr0n00gler:
Please attach a successful testrun 🙃

jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/10708540677

x/dex/keeper/cancel_limit_order.go
```diff
@@ -88,8 +88,8 @@ func (k Keeper) ExecuteCancelLimitOrder(
 		return sdk.Coin{}, sdk.Coin{}, nil, sdkerrors.Wrapf(types.ErrCancelEmptyLimitOrder, "%s", tranche.Key.TrancheKey)
 	}
 
-	k.SaveTrancheUser(ctx, trancheUser)
-	k.SaveTranche(ctx, tranche)
+	k.SaveOrRemoveTrancheUser(ctx, trancheUser)
```
mmikev: I mean, this is much more confusing for me now (it's not obvious at all when the method removes a tranche and when it saves a tranche)

Can't it be two separate methods?

pr0n00gler (changes requested):

x/dex/keeper/cancel_limit_order.go
mmikev: Like, i understand there is a pretty simple condition inside the method, but i think the code would be much easier to read if the condition would be located outside of the method

What do you think?

sotnikov-s: I agree. I'm used to combined methods and functions like `GetOrCreate` and `CreateOrUpdate`, but `SaveOrRemove` are totally opposite operations, and the meaning is unclear for me. I'm still quite bad at the dex module, could you please explain the meaning to me? Even better: could you please add a comprehensive explanation to the comment of the method?

x/dex/keeper/limit_order_tranche.go
```diff
@@ -51,11 +51,11 @@ func (k Keeper) FindLimitOrderTranche(
...
-func (k Keeper) SaveTranche(ctx sdk.Context, tranche *types.LimitOrderTranche) {
+func (k Keeper) SaveOrMakeInactiveTranche(ctx sdk.Context, tranche *types.LimitOrderTranche) {
 	if tranche.HasTokenIn() {
 		k.SetLimitOrderTranche(ctx, tranche)
 	} else {
-		k.SetInactiveLimitOrderTranche(ctx, tranche)
+		k.SaveOrRemoveInactiveTranche(ctx, tranche)
 		k.RemoveLimitOrderTranche(ctx, tranche.Key)
 		ctx.EventManager().EmitEvents(types.GetEventsDecTotalOrders(tranche.Key.TradePairId))
 	}
```
sotnikov-s: the method is called `SaveOrMakeInactiveTranche`, but inside we either
- `SetLimitOrderTranche` which sound like the `Save`, or
- `SaveOrRemoveInactiveTranche` which doesn't sound like `MakeInactive`

This is confusing. Could you please add comprehensive comments here and probably give these better names?

x/dex/keeper/place_limit_order.go
```diff
@@ -174,13 +174,13 @@ func (k Keeper) ExecutePlaceLimitOrder(
...
 
-		k.SaveTranche(ctx, placeTranche)
+		k.SaveOrMakeInactiveTranche(ctx, placeTranche)
 
 		totalIn = totalIn.Add(amountLeft)
 		sharesIssued = amountLeft
 	}
 
-	k.SaveTrancheUser(ctx, trancheUser)
+	k.SaveOrRemoveTrancheUser(ctx, trancheUser)
```
sotnikov-s: we're placing a limit order here. can it be the case that we go to the first branch of the `SaveOrRemoveTrancheUser` that does `RemoveLimitOrderTrancheUser` in case `if trancheUser.IsEmpty()`?

x/dex/keeper/withdraw_filled_limit_order.go
```diff
@@ -98,7 +98,7 @@ func (k Keeper) ExecuteWithdrawFilledLimitOrder(
 
 	}
 
-	k.SaveTrancheUser(ctx, trancheUser)
+	k.SaveOrRemoveTrancheUser(ctx, trancheUser)
```
sotnikov-s: we're withdrawing a filled limit order here. can it be the case that we go to the second branch of the `SaveOrRemoveTrancheUser` that does `SetLimitOrderTrancheUser` in case if `!trancheUser.IsEmpty()`?

sotnikov-s: if answers for this and [this](https://github.com/neutron-org/neutron/pull/698#discussion_r1750439050) are "no", then we definitely can follow [this](https://github.com/neutron-org/neutron/pull/698#discussion_r1747403603) Mike's suggestion and move the logic from `SaveOrRemoveTrancheUser` outside and use pure `Save` and `Remove` methods instead

sotnikov-s: also could you please have a look at all the combined methods and

1) check whether they are applied correctly or there can only be one of the conditions met;
2) if combination is applied correctly in several places, which would imply that we need that method to avoid code duplication, cover the method and its usage with comments so it's clear for the reader how it works and why its called there

x/dex/keeper/cancel_limit_order.go
```diff
@@ -88,8 +88,8 @@ func (k Keeper) ExecuteCancelLimitOrder(
...
+	k.SaveOrMakeInactiveTranche(ctx, tranche)
```
sotnikov-s, line 92: according to the [docs](https://docs.neutron.org/neutron/modules/dex/overview/concepts/tick-liquidity), we set limit order tranches inactive in begin blockers based on their expiration time. why there is a chance we set to inactive here? is this documentation part up to date?

image

pr0n00gler (approved):

jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/11023124360

NicholasDotSol (approved):

jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/11023124360

PR #706, opened 2024-09-03
ibc rate limit module NTRN-393

quasisamurai:
This PR implements a neutron's adaptation of Osmosis' [ibc-rate-limit](https://github.com/osmosis-labs/osmosis/tree/main/x/ibc-rate-limit).
It includes:
- Module itself
- Slightly reworked app.go's modules & keepers init
- Rewritten & adapted tests
- [PR](https://github.com/neutron-org/rate-limiter-contract/pull/1) w updated contract by range team
- [Reworked](https://github.com/neutron-org/neutron/pull/718)  params model
- [Int tests](https://github.com/neutron-org/neutron-integration-tests/pull/347)

[TASK](https://hadronlabs.atlassian.net/jira/software/projects/NTRN/boards/3?selectedIssue=NTRN-393)


x/ibc-rate-limit/contracts/rate-limiter/src/state/storage.rs
```diff
@@ -0,0 +1,53 @@
...
+    rate_limit::RateLimit,
+    rbac::{QueuedMessage, Roles},
+};
+
+/// Only this address can manage the contract. This will likely be the
+/// governance module, but could be set to something else if needed
+pub const GOVMODULE: Item<Addr> = Item::new("gov_module");
+/// Only this address can execute transfers. This will likely be the
+/// IBC transfer module, but could be set to something else if needed
+pub const IBCMODULE: Item<Addr> = Item::new("ibc_module");
```
sotnikov-s: I don't see any usage of this storage item but in tests. seems unnecessary, or am I missing something?

x/ibc-rate-limit/contracts/rate-limiter/src/message_queue.rs
```diff
@@ -0,0 +1,413 @@
...
+///
+/// Returns the id of the queued message
+pub fn queue_message(
+    deps: &mut DepsMut,
+    env: Env,
+    msg: ExecuteMsg,
+    info: MessageInfo,
+) -> Result<String, ContractError> {
+    let timelock_delay = TIMELOCK_DELAY.load(deps.storage, info.sender.to_string())?;
+    let message_id = format!("{}_{}", env.block.height, env.transaction.unwrap().index);
```
sotnikov-s: it's worth replacing the unwrap()

x/ibc-rate-limit/contracts/rate-limiter/src/helpers.rs
```diff
@@ -0,0 +1,61 @@
...
+
+use cosmwasm_std::{to_binary, Addr, CosmosMsg, StdResult, WasmMsg};
+
+use crate::msg::ExecuteMsg;
+use crate::msg::SudoMsg;
+
+/// CwTemplateContract is a wrapper around Addr that provides a lot of helpers
+/// for working with this.
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq, JsonSchema)]
+pub struct RateLimitingContract(pub Addr);
```
sotnikov-s: since this struct and its methods are only used in `integration_tests.rs`, I suggest moving these to that file for better structure and understanding that these are for tests only. that would also justify the unwrap() usage below

x/ibc-rate-limit/contracts/rate-limiter/src/query.rs
```diff
@@ -0,0 +1,187 @@
...
+
+/// Returns the id's of all queued messages
+pub fn get_message_ids(storage: &dyn Storage) -> StdResult<Binary> {
+    to_binary(
+        &MESSAGE_QUEUE
+            .iter(storage)?
+            .filter_map(|message| Some(message.ok()?.message_id))
+            .collect::<Vec<_>>(),
+    )
+}
```
sotnikov-s: this handler seems to need pagination

x/ibc-rate-limit/contracts/rate-limiter/src/contract.rs
```diff
@@ -0,0 +1,203 @@
...
+        QueryMsg::GetMessageIds => query::get_message_ids(deps.storage),
+        QueryMsg::GetMessage { id } => query::get_queued_message(deps.storage, id),
+        QueryMsg::GetDenomRestrictions { denom } => {
+            query::get_denom_restrictions(deps.storage, denom)
+        }
+    }
+}
+
+#[cfg_attr(not(feature = "library"), entry_point)]
+pub fn migrate(deps: DepsMut, _env: Env, _msg: MigrateMsg) -> Result<Response, ContractError> {
```
sotnikov-s: why don't we empty the migrate func for simplicity and brevity? it doesn't seem to be applicable for us

x/ibc-rate-limit/contracts/rate-limiter/src/sudo.rs
```diff
@@ -0,0 +1,199 @@
...
+            Ok(add_rate_limit_attributes(acc, result))
+        });
+
+    results.iter().try_fold(response?, |acc, result| {
+        Ok(add_rate_limit_attributes(acc, result))
+    })
+}
+
+// #[cfg(any(feature = "verbose_responses", test))]
+fn add_rate_limit_attributes(response: Response, result: &RateLimit) -> Response {
```
sotnikov-s: it's not a noop since `#[cfg(any(feature = "verbose_responses", test))]` is commented out for the `add_rate_limit_attributes` fn and not applied for calls, right?

app/app.go
```diff
@@ -588,41 +597,45 @@ func New(
...
+		authtypes.NewModuleAddress(adminmoduletypes.ModuleName).String(),
+	)
+
+	app.WireICS20PreWasmKeeper(appCodec)
+	app.PFMModule = packetforward.NewAppModule(app.PFMKeeper, app.GetSubspace(pfmtypes.ModuleName))
+
 	app.ICAControllerKeeper = icacontrollerkeeper.NewKeeper(
 		appCodec, keys[icacontrollertypes.StoreKey], app.GetSubspace(icacontrollertypes.SubModuleName),
-		app.IBCKeeper.ChannelKeeper, // may be replaced with middleware such as ics29 feerefunder
+		app.RateLimitingICS4Wrapper, // определяется в wireisc20
```
mmikev: Wrong language

app/app.go
```diff
@@ -1615,6 +1581,7 @@ func initParamsKeeper(appCodec codec.BinaryCodec, legacyAmino *codec.LegacyAmino
 	paramsKeeper.Subspace(tokenfactorytypes.StoreKey).WithKeyTable(tokenfactorytypes.ParamKeyTable())
 	paramsKeeper.Subspace(interchainqueriesmoduletypes.StoreKey).WithKeyTable(interchainqueriesmoduletypes.ParamKeyTable())
 	paramsKeeper.Subspace(interchaintxstypes.StoreKey).WithKeyTable(interchaintxstypes.ParamKeyTable())
+	// paramsKeeper.Subspace(ibcratelimittypes.ModuleName)
```
mmikev: Commented code, remove it, please

app/app.go
```diff
@@ -1673,3 +1640,78 @@ func overrideWasmVariables() {
...
+		app.PFMKeeper,
+		&wasmHooks,
+	)
+
+	ibcratelimitKeeper := ibcratelimitkeeper.NewKeeper(appCodec, app.keys[ibcratelimittypes.ModuleName], authtypes.NewModuleAddress(adminmoduletypes.ModuleName).String())
+	// ChannelKeeper wrapper for rate limiting SendPacket(). The wasmKeeper needs to be added after it's created
+	rateLimitingICS4Wrapper := ibcratelimit.NewICS4Middleware(
+		app.HooksICS4Wrapper,
+		&app.AccountKeeper,
+		// wasm keeper we set later.
```
mmikev: Please leave a comment where exactly we set it

proto/osmosis/ibcratelimit/v1beta1/genesis.proto
mmikev, line 1: Let's move everything related to ibcratelimit into `proto/neutron` folder

x/ibc-rate-limit/client/cli/query.go
mmikev, line 1: Should be uncommented, i suppose?

x/ibc-rate-limit/ibc_module.go
```diff
@@ -0,0 +1,257 @@
...
+	Receiver string `protobuf:"bytes,4,opt,name=receiver,proto3" json:"receiver,omitempty"`
+}
+
+func ValidateReceiverAddress(packet exported.PacketI) error {
+	var receiverObj receiverParser
+
+	if err := json.Unmarshal(packet.GetData(), &receiverObj); err != nil {
+		return err
+	}
+	if len(receiverObj.Receiver) >= 4096 {
```
mmikev: Let's make it a constant

x/ibc-rate-limit/ics4_wrapper.go
```diff
@@ -0,0 +1,108 @@
...
+		return i.channel.SendPacket(ctx, chanCap, sourcePort, sourceChannel, timeoutHeight, timeoutTimestamp, data)
+	}
+
+	// setting 0 as a default so it can be properly parsed by cosmwasm
+	fullPacket := channeltypes.Packet{
+		Sequence:           0,
+		SourcePort:         sourcePort,
+		SourceChannel:      sourceChannel,
+		DestinationPort:    "omitted",
+		DestinationChannel: "omitted",
```
mmikev, line 79: Why do we need to omit destination port and channel?

x/ibc-rate-limit/ics4_wrapper.go
```diff
@@ -0,0 +1,108 @@
...
+	if packetdata.Denom == "" || packetdata.Amount == "" {
+		return i.channel.SendPacket(ctx, chanCap, sourcePort, sourceChannel, timeoutHeight, timeoutTimestamp, data)
+	}
+	contract := i.GetContractAddress(ctx)
+	if contract == "" {
+		// The contract has not been configured. Continue as usual
+		return i.channel.SendPacket(ctx, chanCap, sourcePort, sourceChannel, timeoutHeight, timeoutTimestamp, data)
+	}
+
+	// setting 0 as a default so it can be properly parsed by cosmwasm
```
mmikev, line 73: What does it mean exactly?

x/ibc-rate-limit/keeper/keeper.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/neutron-org/neutron/v4/x/ibc-rate-limit/types"
+)
+
+// Keeper of the globalfee store
+type Keeper struct {
+	cdc      codec.BinaryCodec
+	storeKey storetypes.StoreKey
+
+	// the address capable of executing a MsgUpdateParams message. Typically, this
+	// should be the x/gov module account.
```
mmikev:
```suggestion
	// should be the x/adminmodule module account.
```

x/ibc-rate-limit/types/codec.go
```diff
@@ -0,0 +1,25 @@
...
+
+import (
+	"github.com/cosmos/cosmos-sdk/codec"
+	cdctypes "github.com/cosmos/cosmos-sdk/codec/types"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/cosmos-sdk/types/msgservice"
+)
+
+func RegisterCodec(cdc *codec.LegacyAmino) {
+	cdc.RegisterConcrete(&MsgUpdateParams{}, "osmosis/ibc-rate-limit/update-params", nil)
```
mmikev:
```suggestion
	cdc.RegisterConcrete(&MsgUpdateParams{}, "neutron/ibc-rate-limit/update-params", nil)
```

Please, change it everywhere

app/app.go
```diff
@@ -893,23 +859,18 @@ func New(
 	)
 	interchainTxsModule := interchaintxs.NewAppModule(appCodec, app.InterchainTxsKeeper, app.AccountKeeper, app.BankKeeper)
 	contractManagerModule := contractmanager.NewAppModule(appCodec, app.ContractManagerKeeper)
+	ibcRateLimitmodule := ibcratelimitmodule.NewAppModule(appCodec, *app.RateLimitingICS4Wrapper.IbcratelimitKeeper, *app.RateLimitingICS4Wrapper)
```
mmikev: Let's maybe use pointers here?
Cause we know structs can bring a lot of problems in the future

pr0n00gler (changes requested):

x/ibc-rate-limit/ics4_wrapper.go
quasisamurai: It is needed to be packet parsed properly, so **0** here is just a placeholder which is unused later, but not return error at the same time (and not messing with seq numbers at the same time)

x/ibc-rate-limit/ics4_wrapper.go
quasisamurai: Same as below, just a placeholder

x/ibc-rate-limit/ibcratelimitmodule/module.go
```diff
@@ -0,0 +1,149 @@
+package ibcratelimitmodule
```
sotnikov-s: why do we have such an unusual folders structure? we usually place the `module.go` and the `module_simulation.go` files directly in the module's root

proto/neutron/ibcratelimit/v1beta1/query.yml
```diff
@@ -0,0 +1,10 @@
+keeper:
+  path: "github.com/neutron-org/neutron/v4/x/ibc-rate-limit"
+  struct: "Keeper"
+client_path: "github.com/neutron-org/neutron/v4/x/ibc-rate-limit/client"
+queries:
+  Params:
+    proto_wrapper:
+      query_func: "k.GetParams"
+    cli:
+      cmd: "GetParams"
```
sotnikov-s: what's the point of this file?

sotnikov-s: ah, I see, it's involved in `query_proto_wrap.go` and `grpc_query.go` files generation. what's the point of these files? I can see they contain some kind of grpc clients for the module. why doesn't the proto generated `QueryClient` from `query.pb.go` suffice?

x/ibc-rate-limit/ibc_module.go
```diff
@@ -0,0 +1,255 @@
...
+			ctx.EventManager().EmitEvent(
+				sdk.NewEvent(
+					types.EventBadRevert,
+					sdk.NewAttribute(sdk.AttributeKeyModule, types.ModuleName),
+					sdk.NewAttribute(types.AttributeKeyFailureType, "acknowledgment"),
+					sdk.NewAttribute(types.AttributeKeyPacket, string(packet.GetData())),
+					sdk.NewAttribute(types.AttributeKeyAck, string(acknowledgement)),
+				),
+			)
+		}
```
sotnikov-s, line 183: why don't we emit the `RevertSentPacket` error message as an event attribute? here and in the `OnTimeoutPacket`. for example, here we do the error emission: https://github.com/neutron-org/neutron/blob/9cb8580f2cde7c07d1e6f074680fc589f3700a4c/x/ibc-rate-limit/ibc_module.go#L128-L136

x/ibc-rate-limit/ics4_wrapper.go
sotnikov-s: the contract doesn't need half of the fields of the Packet struct. we could minify the sudo msg payload by just including the required fields

x/ibc-rate-limit/ibc_module.go
```diff
@@ -0,0 +1,255 @@
...
+		return utils.NewEmitErrorAcknowledgement(ctx, types.ErrBadMessage, err.Error())
+	}
+
+	contract := im.ics4Middleware.GetContractAddress(ctx)
+	if contract == "" {
+		// The contract has not been configured. Continue as usual
+		return im.app.OnRecvPacket(ctx, packet, relayer)
+	}
+
+	err := CheckAndUpdateRateLimits(ctx, im.ics4Middleware.ContractKeeper, "recv_packet", contract, packet)
```
sotnikov-s: we have `msgSend` and `msgRecv` defined in `x/ibc-rate-limit/rate_limit.go` which can be reused here as the third argument. same for the `OnRecvPacket` call in the `SendPacket` method

proto/neutron/ibcratelimit/v1beta1/query.yml
quasisamurai: We do, this file needs to be removed thx

x/ibc-rate-limit/ibcratelimitmodule/module.go
quasisamurai: Because the original module has. If no one is against our default structure, i'm ready to rework this part

.gitignore
```diff
@@ -7,3 +7,7 @@ data
 .mutagen.yml
 vendor/
 .testchains
+
+x/ibc-rate-limit/contracts/rate-limiter/target/
```
zavgorodnii: what are these paths and what generates [files in] them?

app/app.go
```diff
@@ -588,41 +597,45 @@ func New(
...
+		authtypes.NewModuleAddress(adminmoduletypes.ModuleName).String(),
+	)
+
+	app.WireICS20PreWasmKeeper(appCodec)
+	app.PFMModule = packetforward.NewAppModule(app.PFMKeeper, app.GetSubspace(pfmtypes.ModuleName))
+
 	app.ICAControllerKeeper = icacontrollerkeeper.NewKeeper(
 		appCodec, keys[icacontrollertypes.StoreKey], app.GetSubspace(icacontrollertypes.SubModuleName),
-		app.IBCKeeper.ChannelKeeper, // may be replaced with middleware such as ics29 feerefunder
+		app.RateLimitingICS4Wrapper, // defiened in wireisc20
```
zavgorodnii:
```suggestion
		app.RateLimitingICS4Wrapper, // defined in wireisc20
```

zavgorodnii:
this PR adds a new module. are there any module docs?

testutil/test_helpers.go
```diff
@@ -317,6 +364,10 @@ func NewProviderConsumerCoordinator(t *testing.T) *ibctesting.Coordinator {
 	coordinator.Chains[chainID] = ibctesting.NewTestChainWithValSet(t, coordinator,
 		chainID, providerChain.Vals, providerChain.Signers)
 
+	chainID = ibctesting.GetChainID(4)
```
zavgorodnii: please move all magic numbers to constants

x/ibc-rate-limit/.cargo/config.toml
```diff
@@ -0,0 +1,3 @@
+[alias]
```
zavgorodnii: why is there a cargo file? also i can see that both `x/ibc-rate-limit/bytecode` and `x/ibc-rate-limit/artifacts` contain `rate_limiter.wasm`, is that intentional?

x/ibc-rate-limit/ibc_module.go
zavgorodnii: please move the string literal to a constant

x/ibc-rate-limit/ibc_module.go
```diff
@@ -0,0 +1,255 @@
...
+	if err != nil {
+		if strings.Contains(err.Error(), "rate limit exceeded") {
```
zavgorodnii: same here, string literal to constant

x/ibc-rate-limit/README.md
```diff
@@ -0,0 +1,307 @@
+# IBC Rate Limit
+
+The IBC Rate Limit module is responsible for adding a governance-configurable rate limit to IBC transfers.
+This is a safety control, intended to protect assets on osmosis in event of:
+
+* a bug/hack on osmosis
```
zavgorodnii: neutron

zavgorodnii: and there are a lot of such entries, please replace them all

x/ibc-rate-limit/ibc_module.go
```diff
@@ -0,0 +1,255 @@
...
+	ctx sdk.Context,
+	packet exported.PacketI,
+) error {
+	contract := im.ics4Middleware.GetContractAddress(ctx)
+	if contract == "" {
+		// The contract has not been configured. Continue as usual
+		return nil
+	}
+
+	if err := UndoSendRateLimit(
```
zavgorodnii: replace with a simple return?

x/ibc-rate-limit/ibcratelimitmodule/module.go
zavgorodnii: let's keep things uniform yes

x/ibc-rate-limit/ics4_wrapper.go
```diff
@@ -0,0 +1,108 @@
...
+		SourcePort:         sourcePort,
+		SourceChannel:      sourceChannel,
+		DestinationPort:    "omitted",
+		DestinationChannel: "omitted",
+		Data:               data,
+		TimeoutTimestamp:   timeoutTimestamp,
+		TimeoutHeight:      timeoutHeight,
+	}
+
+	err := CheckAndUpdateRateLimits(ctx, i.ContractKeeper, "send_packet", contract, fullPacket)
```
zavgorodnii: please move to constant

x/ibc-rate-limit/rate_limit.go
```diff
@@ -0,0 +1,153 @@
...
+	if err != nil {
+		return err
+	}
+
+	sendPacketMsg, err := BuildWasmExecMsg(
+		msgType,
+		packet,
+	)
+	if err != nil {
+		return err
```
zavgorodnii: here and everywhere in this file, please wrap the errors with additional information about what failed (e.g. `fmt.Errorf(err, "failed to BuildWasmExecMsg")`)

x/ibc-rate-limit/.cargo/config.toml
quasisamurai: The original contract wasn't optimised so i optimised it myself leaving 2 versions in the module until further discussions. i'll leave only optimised version for now

x/ibc-rate-limit/ibc_module.go
quasisamurai: why?

testutil/test_helpers.go
quasisamurai: It is not a magic constant, but the common way we init new chains using testutil. i've switched all of such entries to constants but not quite sure if  it has improved readability

.gitignore
quasisamurai: contract code was stored and built right in a module,  this is a rudiment

quasisamurai:
https://github.com/neutron-org/neutron-tests/actions/runs/11123879488/job/30908498345

pr0n00gler (dismissed):

x/ibc-rate-limit/ibc_middleware_test.go
```diff
@@ -0,0 +1,763 @@
...
+
+	// Setup contract
+	testOwner := sdk.MustAccAddressFromBech32(testutil.TestOwnerAddress)
+	suite.StoreTestCode(suite.ChainA.GetContext(), testOwner, "./bytecode/rate_limiter.wasm")
+	quotas := suite.BuildChannelQuota("weekly", channelTest, denom, 604800, 5, 5)
+	addr := suite.InstantiateRLContract(quotas)
+	suite.RegisterRateLimitingContract(addr)
+
+	// send 2.5% (quota is 5%)
+	fmt.Printf("Sending %s from A to B. Represented in chain A as wrapped? %v\n", denom, !native)
```
mmikev: Remove printlns, please

x/ibc-rate-limit/ibc_middleware_test.go
```diff
@@ -0,0 +1,763 @@
...
+	_, err := suite.AssertSend(true, suite.MessageFromAToB(denom, sendAmount))
+	suite.Require().NoError(err)
+
+	// send 2.5% (quota is 5%)
+	fmt.Println("trying to send ", sendAmount)
+	r, _ := suite.AssertSend(true, suite.MessageFromAToB(denom, sendAmount))
```
mmikev, line 308: errors are not handles, fix it everywhere, please

quasisamurai: there is no need for that as soon as `AssertSend` is designed to handle (or expect) the error. behaviour depends on the first parameter

x/ibc-rate-limit/ibc_middleware_test.go
quasisamurai: discussed in dms, we decided to do so

x/ibc-rate-limit/ibc_middleware_test.go
```diff
@@ -0,0 +1,747 @@
...
+          {"channel_id": "%s", "denom": "%s", "quotas": [{"name":"%s", "duration": %d, "send_recv":[%d, %d]},{"name":"%s", "duration": %d, "send_recv":[%d, %d]}] }
+    `, channel, denom, name, duration1, sendPercentage1, recvPercentage1, name2, duration2, sendPercentage2, recvPercentage2)
+}
+
+// Tests
+
+// Test that Sending IBC messages works when the middleware isn't configured
+func (suite *MiddlewareTestSuite) TestSendTransferNoContract() {
+	suite.ConfigureTransferChannel()
+	one := sdkmath.NewInt(1)
```
sotnikov-s: it's `NewInt(2)` used in the [fullSendTest](https://github.com/neutron-org/neutron/blob/c4358887889516185c6ee7077a41f18c2c24f0e2/x/ibc-rate-limit/ibc_middleware_test.go#L318-L319) to avoid some kind of rounding issues. not sure what kind of issues are avoided by that, but I think we better use `NewInt(2)` instead of `NewInt(1)` just in case. here and in several other places

x/ibc-rate-limit/ibc_middleware_test.go
```diff
@@ -0,0 +1,747 @@
...
+	r, err := suite.AssertSend(true, suite.MessageFromAToB(denom, sendAmount))
+	suite.Require().NoError(err)
+
+	// Calculate remaining allowance in the quota
+	attrs := suite.ExtractAttributes(suite.FindEvent(r.GetEvents(), "wasm"))
+
+	used, ok := sdkmath.NewIntFromString(attrs["daily_used_out"])
+	suite.Require().True(ok)
+
+	suite.Require().Equal(used, sendAmount)
```
sotnikov-s, line 400: would be very nice to see such checks after each Send try and after quotas reset to be assured that daily and weekly quotas are as we expect them to be. could you please add such assertions? e.g. check that daily and weekly used out quotas equal to `sendAmount` after the first send, then, once daily is reset, check that the daile quota is reset indeed, and weekly is still at `sendAmount`, and so on for all milestones when we expect the quota usages to change

sotnikov-s (approved):
briefly looked at the code, more thoroughly looked at the tests — LGTM

pr0n00gler (approved):

PR #713, opened 2024-09-06
Fix: Add case to handle negative withdraw amounts

jcompagni10:
Because sharesWithdrawn is rounded up, it is possible to end up in a case where sharesWithdrawn > maxAllowedToWithdraw. We fix this case by checking that the withdraw amount is positive.


jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/10744272200

jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/11022390493

NicholasDotSol (approved):

swelf19 (approved):

pr0n00gler (approved):

PR #715, opened 2024-09-09
feat: dex state testing all tests

swelf19:


jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/11845366477

PR #716, opened 2024-09-09
IBC Rate Limit Changes For Neutron

bonedaddy:
# Overview

Adds changes needed to make the osmosis ibc rate limit contract compatible with neutron

# Changes Made

* Dependency updates
* Migrations needed due to dependency updates
* Replace usage of deprecated functions as a result of dependency changes
* Unit test updates to ensure original functoin is still intact
* Replace osmosis addresses in unit tests with neutron addresses

# Notes

Some comments in the code in the code mention "osmosis" however I wasn't sure whether or not to change this, so please let me know if I should do this.


x/ibc-rate-limit/Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [workspace]
-
+resolver = "2"
```
sotnikov-s: [it is said in the docs](https://doc.rust-lang.org/cargo/reference/resolver.html#resolver-versions) that the resolver is set to version 2 by default if the edition is 2021 or newer. the target branch [defines the edition as 2021](https://github.com/neutron-org/neutron/blob/051a12f0a395be62a8f5c8b6b693f4614a1098b6/x/ibc-rate-limit/contracts/rate-limiter/Cargo.toml#L5), so this line is not necessary. please correct me if I'm wrong

x/ibc-rate-limit/contracts/rate-limiter/src/execute.rs
```diff
@@ -156,23 +156,23 @@ mod tests {
...
-    const GOV_ADDR: &str = "osmo1tzz5zf2u68t00un2j4lrrnkt2ztd46kfzfp58r";
+    const IBC_ADDR: &str = "neutron1cdlz8scnf3mmxdnf4njmtp7vz4gps7fswphrqn";
+    const GOV_ADDR: &str = "neutron1w02khza7ux68ccwmz2hln97mkjspjxes8y2k9v";
 
     #[test] // Tests AddPath and RemovePath messages
     fn management_add_and_remove_path() {
         let mut deps = mock_dependencies();
         IBCMODULE
-            .save(deps.as_mut().storage, &Addr::unchecked(IBC_ADDR))
+            .save(deps.as_mut().storage, &MockApi::default().addr_make(IBC_ADDR))
```
sotnikov-s, line 166: why do you prefer `&MockApi::default().addr_make` over `Addr::unchecked`?

x/ibc-rate-limit/contracts/rate-limiter/src/packet.rs
```diff
@@ -113,11 +103,15 @@ impl Packet {
...
-        .query(&deps.querier)?;
-        Uint256::from_str(&res.amount.unwrap_or_default().amount)
+        let response: QuerySupplyOfResponse = deps.querier.query(&cosmwasm_std::QueryRequest::Grpc({
+            GrpcQuery {
+                path: "/cosmos.bank.v1beta1.Query/SupplyOf".to_string(),
+                data: Binary::from_base64(
+                    &to_base64(self.local_denom(direction))
+                )?
+            }
+        }))?;
```
sotnikov-s: I think we can reuse the grpc query helpers here just as it is done here https://github.com/neutron-org/neutron-dev-contracts/blob/2589bbdda4496608626d457902d548e36134945a/contracts/grpc_querier/src/contract.rs#L68

sotnikov-s (commented):
overall LGTM. just a couple of questions and preferences that may be fixed after the merge when working with the base branch

x/ibc-rate-limit/contracts/rate-limiter/src/execute.rs
bonedaddy: After upgrading to `2.x` of `cosmwasm-std` unit tests were triggering the following error

```
called `Result::unwrap()` on an `Err` value: Std(GenericErr { msg: "Error decoding bech32", backtrace: <disabled> })
```

When reading through the [migration](https://github.com/CosmWasm/cosmwasm/blob/main/MIGRATING.md) docs for upgrading to `2.x` this is mentioned

> Replace addresses in unit tests with valid bech32 addresses. This has to be done for all addresses that are validated or canonicalized during the test or within the contract. The easiest way to do this is by using MockApi::addr_make. It generates a bech32 address from any string:

So I just ended up replacing all instances of `Addr::unchecked` with `MockApi::default().addr_make`

x/ibc-rate-limit/Cargo.toml
bonedaddy: You are most definitely correct, I have pushed a commit which changes this.

x/ibc-rate-limit/contracts/rate-limiter/src/packet.rs
bonedaddy: Makes sense and that would simplify the function a lot, should I remove and change to 

```rust
    pub fn channel_value(&self, deps: Deps, direction: &FlowType) -> Result<Uint256, StdError> {
        Ok(Uint256::from_uint128(deps.querier.query_supply(self.local_denom(direction))?.amount))
    }
```

sotnikov-s: yeah, that should work

sotnikov-s (approved):

pr0n00gler (approved):

PR #737, opened 2024-09-23
Chore: Add ErrNoLiquidity

jcompagni10:
Add an additional error state for MHS and LOs so that the end user can disambiguate between the limit price not being satisfied and insufficient liquidity. This is helpful for simulations and contracts because more information so that limit order request can be changed to match current dex conditions.


jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/11016860500

NicholasDotSol (approved):

pr0n00gler (approved):

PR #707, opened 2024-09-03
Chore: more dex event attributes

jcompagni10:
Add additional dex event attributes for better indexing (p2p) and frontend integration. 

*PlaceLimitOrder:*
`SwapAmountIn`
`SwapAmountOut`
*WithdrawFilledLimitOrder:*
`TokenInAmountOut`
`TokenOutAmountOut`
(matches CancelLimitOrder)

Also adds new TrancheUserUpdateEvent

PR #741, opened 2024-09-24
Chore: better update fn naming

jcompagni10:
Clearer naming for update fns. This eliminates confusion about what the various functions used for updating KV store records do. 
Note: because of PRs like [this](https://github.com/neutron-org/neutron/pull/707), that add additional event handlers to the save functions it is preferable to use a single update function, so that any future logic/event hooks only need to happen in one place. 

`UpdateLimitOrderTranche`
`UpdateLimitOrderTrancheUser`
`UpdatePool`
`UpdatePoolReserves`

Also, lots of extra comments and small refactoring so that each function is easier to understand.


jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/11016553281

NicholasDotSol (dismissed):

swelf19 (dismissed):

pr0n00gler (approved):

PR #740, opened 2024-09-24
Chore: ICS V6

pr0n00gler:
Related PRs:
* https://github.com/neutron-org/neutron/pull/740
* https://github.com/neutron-org/neutron-std/pull/1
* https://github.com/neutron-org/neutron-integration-tests/pull/359
* https://github.com/neutron-org/neutronjsplus/pull/67
* https://github.com/neutron-org/neutronjs/pull/9
* https://github.com/neutron-org/neutron-dev-contracts/pull/69

PR #745, opened 2024-09-27
CV fixSignerListener: Listening for new connection log spam

chillyvee:
This patch resolves a log spam problem:
```SignerListener: Listening for new connection```


pr0n00gler:
Thanks, CometBFT will be bumped here: https://github.com/neutron-org/neutron/pull/740

PR #761, opened 2024-11-08
fix: really removes tmp dir after cli run

swelf19:
Currently we get a new tmp dir every `/tmp/neutron*` after every cli call (any query or tx), and the dir never removed.
The PR fixes that.


pr0n00gler (approved):

swelf19:
https://github.com/neutron-org/neutron-tests/actions/runs/11778438084

PR #766, opened 2024-11-18
Feat: state verifier

pr0n00gler:
This PR implements a module that allows to verify that some `storage values` were indeed present on a particular `block height` in the chain.

The idea is the similar how Neutron's KV ICQ works: each `StorageValue` in Cosmos SDK is stored in [KV-IAVL storage](https://github.com/cosmos/iavl).
And to be more precise it's stored in a structure called [`MerkleTree`](https://github.com/cosmos/cosmos-sdk/blob/ae77f0080a724b159233bd9b289b2e91c0de21b5/docs/interfaces/lite/specification.md).
The tree allows to compose `Proof` for `key` and `value` pairs that can prove two things using `RootHash` of the tree:
* `key` and `value` are present in the tree;
* `key` is not present in the tree.

On each block the module's `BeginBlocker` is being called, and it saves `ConsensusState` of the current block height in the storage to use it for verification of storage values later:

```go
consensusState := tendermint.ConsensusState{
    Timestamp:          ctx.BlockTime(), // current block time
    Root:               ibccommitmenttypes.NewMerkleRoot(headerInfo.AppHash), // .AppHash for the previous block
    NextValidatorsHash: cometInfo.GetValidatorsHash(), // hash of the validator set for the next block
}
```

For verification only `.Root` (`header.AppHash`) is used, but it's good to save all the values just in case and do not leave them empty.

The main query of the module that accepts slice of `[]StorageValue` structures and `blockHeight` on which those `StorageValues` are present.
The module verifies the values and returns an error if values cannot be verified `{valid: true}` structure if values are valid.


app/app.go
```diff
@@ -11,6 +11,9 @@ import (
 	"time"
 
 	dynamicfeestypes "github.com/neutron-org/neutron/v5/x/dynamicfees/types"
+	state_verifier "github.com/neutron-org/neutron/v5/x/state-verifier"
```
NeverHappened: snake_case?

third_party/proto/ibc/lightclients/tendermint/v1/tendermint.proto
```diff
@@ -0,0 +1,101 @@
+syntax = "proto3";
+
+package ibc.lightclients.tendermint.v1;
+
+option go_package = "github.com/cosmos/ibc-go/v8/modules/light-clients/07-tendermint;tendermint";
```
NeverHappened, line 5: typo here?

proto/neutron/state_verifier/v1/query.proto
```diff
@@ -0,0 +1,23 @@
...
+}
+
+// QueryVerifyStateValuesRequest describes a structure to verify storage values from Neutron state from a particular height in the past
+message QueryVerifyStateValuesRequest {
+  uint64 height = 1;
+  repeated neutron.interchainqueries.StorageValue storage_values = 2;
+}
+
+message QueryVerifyStateValuesResponse {
+  bool valid = 1;
```
NeverHappened, line 27: what if we didn't have consensus state on that height? would it make sense to return specific value in this case instead of error? just speculating

x/state-verifier/keeper/query.go
```diff
@@ -0,0 +1,18 @@
...
+	"github.com/neutron-org/neutron/v5/x/state-verifier/types"
+)
+
+// VerifyStateValues implements `VerifyStateValues` gRPC query to verify storage values
+func (k *Keeper) VerifyStateValues(ctx context.Context, request *types.QueryVerifyStateValuesRequest) (*types.QueryVerifyStateValuesResponse, error) {
+	if err := k.Verify(sdk.UnwrapSDKContext(ctx), int64(request.Height), request.StorageValues); err != nil {
+		return nil, err
+	}
+
+	return &types.QueryVerifyStateValuesResponse{Valid: true}, nil
```
NeverHappened: Looks incorrect that we never return `Valid: false` actually

x/state-verifier/keeper/keeper.go
```diff
@@ -0,0 +1,158 @@
...
+	if csBz == nil {
+		return errors.Wrap(sdkerrors.ErrKeyNotFound, fmt.Sprintf("consensus state for block %d not found", blockHeight))
+	}
+
+	var cs tendermint.ConsensusState
+	if err := k.cdc.Unmarshal(csBz, &cs); err != nil {
+		return errors.Wrap(sdkerrors.ErrJSONUnmarshal, err.Error())
+	}
+
+	for _, result := range values {
```
NeverHappened: why name it `result`?

x/state-verifier/types/keys.go
```diff
@@ -0,0 +1,25 @@
...
+	prefixConsensusStateKey = iota + 1
+)
+
+var ConsensusStateKey = []byte{prefixConsensusStateKey}
+
+func GetConsensusStateKey(height int64) []byte {
+	return append(ConsensusStateKey, []byte(strconv.FormatInt(height, 10))...)
+}
+
+func ExtractHeightFromConsensusStateKey(key []byte) (int64, error) {
```
NeverHappened: unused? if needed would be good to have few tests on this function

NeverHappened (changes requested):

third_party/proto/ibc/lightclients/tendermint/v1/tendermint.proto
mmikev: Don't think so, it's copypasted from ibc-go itself: https://github.com/cosmos/ibc-go/blob/6b84068b10d15f208ae1839ee5479f59fbf1c3df/proto/ibc/lightclients/tendermint/v1/tendermint.proto#L5

proto/neutron/state_verifier/v1/query.proto
mmikev: I don't get your question, elaborate, please

x/state-verifier/keeper/query.go
mmikev: Yeah, but i don't have any others good ways in mind:
* returning just an empty struct seems not right as well;
* returning false (not an err) in case values are not proovable doesn't seem correct as well (we need return an err to tell what's wrong exactly);

Do you have any suggestions?

x/state-verifier/keeper/keeper.go
mmikev: fixed

x/state-verifier/types/keys.go
mmikev: removed

app/app.go
mmikev: fixed

x/state-verifier/keeper/query.go
NeverHappened: Agreed, seems to be no other way

proto/neutron/state_verifier/v1/query.proto
NeverHappened: Ah, I see you return errors in all cases

NeverHappened (dismissed):

x/state-verifier/types/keys.go
```diff
@@ -0,0 +1,21 @@
...
+)
+
+const (
+	prefixConsensusStateKey = iota + 1
+)
+
+var ConsensusStateKey = []byte{prefixConsensusStateKey}
+
+func GetConsensusStateKey(height int64) []byte {
+	return append(ConsensusStateKey, []byte(strconv.FormatInt(height, 10))...)
```
swelf19: Not sure it's a good way to build a key from a number, simple formatting like this gives asc order like '10' '100' '20' during iteration, when you really expect '10' '20' '100'. I can see 'sdk.Uint64ToBigEndian' widly used to convert number into sortable (in integer way) string

x/state-verifier/keeper/keeper.go
```diff
@@ -0,0 +1,158 @@
...
+		return errors.Wrap(sdkerrors.ErrJSONUnmarshal, err.Error())
+	}
+
+	for _, value := range values {
+		proof, err := ibccommitmenttypes.ConvertProofs(value.Proof)
+		if err != nil {
+			return errors.Wrapf(sdkerrors.ErrInvalidType, "failed to convert crypto.ProofOps to MerkleProof: %v", err)
+		}
+
+		path := ibccommitmenttypes.NewMerklePath(value.StoragePrefix, string(value.Key))
```
swelf19: The same code block is used in icq module https://github.com/neutron-org/neutron/blob/main/x/interchainqueries/keeper/msg_server.go#L254
How do you think to extract it into common method and reuse it, instead of just copying

x/state-verifier/types/keys.go
mmikev: Fixed

x/state-verifier/keeper/keeper.go
mmikev: Done, please, check it

swelf19 (dismissed):

proto/neutron/state_verifier/v1/genesis.proto
```diff
@@ -0,0 +1,16 @@
+syntax = "proto3";
+package neutron.state_verifier.v1;
+
+import "ibc/lightclients/tendermint/v1/tendermint.proto";
+
+option go_package = "github.com/neutron-org/neutron/v5/x/state-verifier/types";
+
+// ConsensusState describes a "light" consensus state of Neutron at a particular height
```
sotnikov-s: please read the documentation policy and apply the guidelines https://www.notion.so/hadron/WIP-Neutron-team-documentation-policy-14485d6b9b10808e92b8e309fdf70d91

sotnikov-s (changes requested):

utils/storageverification/errors.go
```diff
@@ -0,0 +1,11 @@
+package storageverification
+
+import "cosmossdk.io/errors"
+
+const StateVerificationCodespace = "state_verification"
```
sotnikov-s: it's sometimes storage verification, sometimes state verification. could you please choose between these terms and stick with it (inspired by [ubiquitous language](https://martinfowler.com/bliki/UbiquitousLanguage.html))?

utils/storageverification/verify.go
```diff
@@ -0,0 +1,50 @@
...
+		path := ibccommitmenttypes.NewMerklePath(value.StoragePrefix, string(value.Key))
+		// identify what kind proofs (non-existence proof always has *ics23.CommitmentProof_Nonexist as the first item) we got
+		// and call corresponding method to verify it
+		switch proof.GetProofs()[0].GetProof().(type) {
+		// we can get non-existence proof if someone queried some key which is not exists in the storage on remote chain
+		case *ics23.CommitmentProof_Nonexist:
+			if err := proof.VerifyNonMembership(proofSpecs, root, path); err != nil {
+				return errors.Wrapf(ErrInvalidProof, "failed to verify proof: %v", err)
+			}
+			value.Value = nil
```
sotnikov-s: why do we overwrite values during a verification operation?

utils/storageverification/verify.go
```diff
@@ -0,0 +1,50 @@
...
+func VerifyStorageValues(stValues []*types.StorageValue, root exported.Root, proofSpecs []*ics23.ProofSpec, verifyCallback VerifyCallback) error {
+	for index, value := range stValues {
+		proof, err := ibccommitmenttypes.ConvertProofs(value.Proof)
+		if err != nil {
+			return errors.Wrapf(ErrInvalidType, "failed to convert crypto.ProofOps to MerkleProof: %v", err)
+		}
+
+		if verifyCallback != nil {
+			if err := verifyCallback(index); err != nil {
+				return errors.Wrapf(ErrInvalidStorageValue, "failed to verify: %v", err)
```
sotnikov-s: would be more clear what exactly failed
```suggestion
				return errors.Wrapf(ErrInvalidStorageValue, "verify callback failed at index %d: %v", index, err)
```

utils/storageverification/verify.go
```diff
@@ -0,0 +1,50 @@
...
+import (
+	"cosmossdk.io/errors"
+	ibccommitmenttypes "github.com/cosmos/ibc-go/v8/modules/core/23-commitment/types"
+	"github.com/cosmos/ibc-go/v8/modules/core/exported"
+	ics23 "github.com/cosmos/ics23/go"
+
+	"github.com/neutron-org/neutron/v6/x/interchainqueries/types"
+)
+
+type VerifyCallback func(index int) error
```
sotnikov-s: I think it would be more clear if we called it e.g. `type ExtraVerifier func...` (or something on your taste that emphasises additional verification capability). VeryfiCallback sounds more like something what is called after a successful verification for me, not like something that supplements basic verification

and please add some comments on the type declaration itself

x/state-verifier/README.md
```diff
@@ -0,0 +1,33 @@
+# Overview
+
+The State Verifier module allows to verify that some `storage values` were indeed present on a particular `block height` in the chain.
```
sotnikov-s:
```suggestion
The State Verifier module allows to verify that some `storage values` were indeed present or absent on a particular `block height` in the chain.
```

x/state-verifier/README.md
```diff
@@ -0,0 +1,33 @@
...
+    Root:               ibccommitmenttypes.NewMerkleRoot(headerInfo.AppHash), // .AppHash for the previous block
+    NextValidatorsHash: cometInfo.GetValidatorsHash(), // hash of the validator set for the next block
+}
+```
+
+For verification only `.Root` (`header.AppHash`) is used, but it's good to save all the values just in case and do not leave them empty.
+
+### VerifyStateValues query
+The main query of the module that accepts slice of `[]StorageValue` structures and `blockHeight` on which those `StorageValues` are present.
+The module verifies the values and returns an error if values cannot be verified `{valid: true}` structure if values are valid.
```
sotnikov-s: punctuation is incorrect, is it?

x/state-verifier/README.md
```diff
@@ -0,0 +1,33 @@
...
+    Timestamp:          ctx.BlockTime(), // current block time
+    Root:               ibccommitmenttypes.NewMerkleRoot(headerInfo.AppHash), // .AppHash for the previous block
+    NextValidatorsHash: cometInfo.GetValidatorsHash(), // hash of the validator set for the next block
+}
+```
+
+For verification only `.Root` (`header.AppHash`) is used, but it's good to save all the values just in case and do not leave them empty.
+
+### VerifyStateValues query
+The main query of the module that accepts slice of `[]StorageValue` structures and `blockHeight` on which those `StorageValues` are present.
```
sotnikov-s:
```suggestion
The main query of the module that accepts slice of `[]StorageValue` structures and `blockHeight` against which those `StorageValues` are verified.
```

x/state-verifier/genesis.go
```diff
@@ -0,0 +1,32 @@
...
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/neutron-org/neutron/v6/x/state-verifier/keeper"
+	"github.com/neutron-org/neutron/v6/x/state-verifier/types"
+)
+
+// InitGenesis initializes the module's state from a provided genesis state.
+func InitGenesis(ctx sdk.Context, k *keeper.Keeper, genState types.GenesisState) {
+	// this line is used by starport scaffolding # genesis/module/init
```
sotnikov-s: do we need this comment?

x/state-verifier/keeper/keeper.go
```diff
@@ -0,0 +1,131 @@
...
+	icqtypes "github.com/neutron-org/neutron/v6/x/interchainqueries/types"
+	"github.com/neutron-org/neutron/v6/x/state-verifier/types"
+)
+
+type (
+	Keeper struct {
+		cdc        codec.BinaryCodec
+		storeKey   storetypes.StoreKey
+		cometInfo  comet.BlockInfoService
+		headerInfo header.Service
```
sotnikov-s: don't you think it would be clearer to name them somehow with `Service` suffix, not with `Info` suffix?

sotnikov-s (changes requested):

proto/neutron/state_verifier/v1/genesis.proto
mmikev: I think it's fixed, please check

utils/storageverification/errors.go
mmikev: Renamed

utils/storageverification/verify.go
mmikev: I removed this particular line but please check the code here: https://github.com/neutron-org/neutron/pull/766/commits/400c7a21deebb674116c67076244b96fd4506cfa#diff-e173e7b6204e2bc1c65d454af5203d12a4644644ad3706a4cf4a8e39c17fed69R253

utils/storageverification/verify.go
mmikev: Done

x/state-verifier/README.md
mmikev: Done

x/state-verifier/genesis.go
mmikev: Done

x/state-verifier/keeper/keeper.go
mmikev: Done

utils/stateverification/verify.go
```diff
@@ -0,0 +1,50 @@
...
+	"cosmossdk.io/errors"
+	ibccommitmenttypes "github.com/cosmos/ibc-go/v8/modules/core/23-commitment/types"
+	"github.com/cosmos/ibc-go/v8/modules/core/exported"
+	ics23 "github.com/cosmos/ics23/go"
+
+	"github.com/neutron-org/neutron/v6/x/interchainqueries/types"
+)
+
+// ExtraVerifier is a type
+type ExtraVerifier func(index int) error
```
sotnikov-s: > ExtraVerifier is a type

is it? :) please complete the description

proto/neutron/state_verifier/v1/query.proto
```diff
@@ -0,0 +1,27 @@
...
+message QueryVerifyStateValuesRequest {
+  // height refers to the block height to which the storage values belong.
+  uint64 height = 1;
+  // storage_values that are present on the specified height and must be verified against
+  repeated neutron.interchainqueries.StorageValue storage_values = 2;
+}
+
+// QueryVerifyStateValuesResponse is a response structure for `VerifyStateValues` query
+message QueryVerifyStateValuesResponse {
+  bool valid = 1;
```
sotnikov-s, line 27: please add a description for the field. I'd suggest to emphasise that this is validity of all the storage values passed to the request at a specific height

proto/neutron/state_verifier/v1/genesis.proto
```diff
@@ -0,0 +1,20 @@
...
+message ConsensusState {
+  // height describes a block height for which the consensus height is saved
+  int64 height = 1;
+
+  // ConsensusState defines the consensus state from Tendermint
+  ibc.lightclients.tendermint.v1.ConsensusState cs = 2;
```
sotnikov-s, line 14: don't miss this policy point: https://www.notion.so/hadron/WIP-Neutron-team-documentation-policy-14485d6b9b10808e92b8e309fdf70d91?pvs=4#14585d6b9b1080c4a4cef735072d8981 let me know if you disagree or something

proto/neutron/state_verifier/v1/query.proto
```diff
@@ -0,0 +1,27 @@
...
+option go_package = "github.com/neutron-org/neutron/v5/x/state-verifier/types";
+
+service Query {
+  // VerifyStateValues verifies the values and returns an error if values cannot be verified. The query returns `QueryVerifyStateValuesResponse` response.
+  rpc VerifyStateValues(QueryVerifyStateValuesRequest) returns (QueryVerifyStateValuesResponse) {
+    option (google.api.http).get = "/neutron/state-verifier/verify_state_values";
+  }
+}
+
+// QueryVerifyStateValuesRequest describes a structure to verify storage values from Neutron state from a particular height in the past
```
sotnikov-s: > ... storage values from **Neutron** state ...

I think we should consider not mentioning Neutron in our modules' code and docs but use some generic term like network/chain/something else. here's why:
1. other teams developing chains based on cosmos sdk might import our modules and there will be an inappropriate Neutron mention;
2. this is kind of out of scope for modules to know which exactly network they are used/ran in. it would be more correct to imply that there is some abstract network or chain, but not neutron or some other concrete entity

what do you think?

x/state-verifier/README.md
```diff
@@ -30,4 +30,5 @@ For verification only `.Root` (`header.AppHash`) is used, but it's good to save
 
 ### VerifyStateValues query
 The main query of the module that accepts slice of `[]StorageValue` structures and `blockHeight` against which those `StorageValues` are verified.
-The module verifies the values and returns an error if values cannot be verified `{valid: true}` structure if values are valid.
\ No newline at end of file
+The module verifies the values and returns an error if values cannot be verified. The query returns `{valid: true}` structure if values are valid.
```
sotnikov-s, line 33: when does the module return `{valid: false}`? reading that, the reader might think that it's either true if values are valid or error if they're not

x/interchainqueries/keeper/msg_server.go
```diff
@@ -235,43 +236,29 @@ func (m msgServer) SubmitQueryResult(goCtx context.Context, msg *types.MsgSubmit
 			return nil, err
 		}
 
-		for index, result := range msg.Result.KvResults {
-			proof, err := ibccommitmenttypes.ConvertProofs(result.Proof)
-			if err != nil {
-				ctx.Logger().Debug("SubmitQueryResult: failed to ConvertProofs",
-					"error", err, "query", query, "message", msg)
-				return nil, errors.Wrapf(types.ErrInvalidType, "failed to convert crypto.ProofOps to MerkleProof: %v", err)
+		if err := stateverification.VerifyStorageValues(msg.Result.KvResults, consensusState.GetRoot(), clientState.ProofSpecs, func(index int) error {
```
sotnikov-s: what do you think about moving the inline-defined extra verification func to a variable with detailed explanation of its implementation and pecularities? that could improve readability of both the extra verification func and the SubmitQueryResult method

x/state-verifier/keeper/keeper.go
sotnikov-s: don't see any changes in this regard

x/state-verifier/keeper/query.go
```diff
@@ -0,0 +1,18 @@
...
+)
+
+// VerifyStateValues implements `VerifyStateValues` gRPC query to verify storage values
+func (k *Keeper) VerifyStateValues(ctx context.Context, request *types.QueryVerifyStateValuesRequest) (*types.QueryVerifyStateValuesResponse, error) {
+	if err := k.Verify(sdk.UnwrapSDKContext(ctx), int64(request.Height), request.StorageValues); err != nil { //nolint:gosec
+		return nil, err
+	}
+
+	return &types.QueryVerifyStateValuesResponse{Valid: true}, nil
+}
```
sotnikov-s: do you remember that we usually decoupled Keeper from query and msg server implementations eg for [query](https://github.com/neutron-org/neutron/blob/main/x/revenue/keeper/grpc_query_server.go) and [msg](https://github.com/neutron-org/neutron/blob/main/x/revenue/keeper/msg_server.go)? could you please stick with this approach?

sotnikov-s (changes requested):

utils/stateverification/verify.go
```diff
@@ -0,0 +1,50 @@
...
+			if err := proof.VerifyNonMembership(proofSpecs, root, path); err != nil {
+				return errors.Wrapf(ErrInvalidProof, "failed to verify proof: %v", err)
+			}
+		case *ics23.CommitmentProof_Exist:
+			if err := proof.VerifyMembership(proofSpecs, root, path, value.Value); err != nil {
+				return errors.Wrapf(ErrInvalidProof, "failed to verify proof: %v", err)
+			}
+		default:
+			return errors.Wrapf(ErrInvalidProof, "unknown proof type %T", proof.GetProofs()[0].GetProof())
+		}
```
sotnikov-s: I think it would make sense to add more details on returned errors here. for example like that:

```go
case *ics23.CommitmentProof_Nonexist:
	if err := proof.VerifyNonMembership(proofSpecs, root, path); err != nil {
		return errors.Wrapf(ErrInvalidProof, "non-membership verification failed for value at index %d with path %s: %v", index, path, err)
	}
case *ics23.CommitmentProof_Exist:
	if err := proof.VerifyMembership(proofSpecs, root, path, value.Value); err != nil {
		return errors.Wrapf(ErrInvalidProof, "membership verification failed for value at index %d with path %s: %v", index, path, err)
	}
```

utils/stateverification/verify.go
mmikev: Done

proto/neutron/state_verifier/v1/query.proto
mmikev: Done

proto/neutron/state_verifier/v1/genesis.proto
mmikev: Done

proto/neutron/state_verifier/v1/query.proto
mmikev: Done

x/interchainqueries/keeper/msg_server.go
mmikev: Done

x/state-verifier/keeper/query.go
mmikev: Done

utils/stateverification/verify.go
mmikev: Done

x/state-verifier/keeper/keeper.go
mmikev: Done

x/state-verifier/README.md
mmikev: > reading that, the reader might think that it's either true if values are valid or error if they're not

Exactly, this is what i suggest to do

proto/neutron/state_verifier/v1/query.proto
```diff
@@ -7,21 +7,22 @@ import "neutron/interchainqueries/tx.proto";
...
 }
 
-// QueryVerifyStateValuesRequest describes a structure to verify storage values from Neutron state from a particular height in the past
+// Describes a structure to verify storage values from the chain state from a particular height in the past
 message QueryVerifyStateValuesRequest {
-  // height refers to the block height to which the storage values belong.
+  // Refers to the block height to which the storage values belong.
   uint64 height = 1;
-  // storage_values that are present on the specified height and must be verified against
+  // A slice of []StorageValue that are present on the specified height and must be verified against
```
sotnikov-s: 1. `A slice of []StorageValue` makes it a slice of slices which is incorrect and might be confusing
2. `that are present on the specified height` is also incorrect because there might be non-existence proofs provided
3. the `StorageValue` type is imported from the icq module which doesn't seem to be correct: it doesn't have to do with the icq module. do you have ideas on how to do it the right way?

proto/neutron/state_verifier/v1/query.proto
sotnikov-s: still can see a Neutron mention in the `genesis.proto` file

sotnikov-s (changes requested):

proto/neutron/state_verifier/v1/query.proto
mmikev: 1. Fixed
2. Fixed
3. You are implying that either we need to move this type into state-verifier module, or maybe even into a separate utils package (where we already have values verification code). It make sense but:
    1. Our current project structure does not allow us to generate proto code in some custom folder/package (e.g. /utils). It's possible to do that only into x/<MODULE>. We can consider adding this functionality but it's out of scope.
    2.  I don't like the idea of moving this type somewhere else from the ICQ module because it's not kinda backwards compatible. I'll try to think about what can be broken exactly but let's not do this right now anyway.

proto/neutron/state_verifier/v1/query.proto
mmikev: Fixed

sotnikov-s (approved):

PR #792, opened 2024-12-20
chore: keep dex message events [NTRN-439] 

jcompagni10:
see: https://github.com/neutron-org/wasmd/pull/277


jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/12915215317

PR #805, opened 2025-01-23
Feat: Swap on Deposit #NTRN-429

jcompagni10:
If a deposit is behind enemy lines (BEL) swap through the offending liquidity with the supplied liquidity. Either the deposit amount will be used up or the offending liquidity will be removed. This allows for easier deposits at tight spreads without first having to place a limit order to clear the liquidity.  

**RELATED** 
https://github.com/neutron-org/neutron-std/pull/11
**TODO**:
-Quint model update
-neutron-std update


jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/13793677689

jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/13850458813

x/dex/keeper/deposit.go
```diff
@@ -167,5 +184,101 @@ func (k Keeper) ExecuteDeposit(
...
+			depositAmount0 = inAmount0.Sub(swapToken0In.Amount)
+			inAmount1 = amount1
+			depositAmount1 = amount1.Add(swapToken1Out.Amount)
+			swappedToken0 = true
+		}
+
+	}
+
+	if amount1.IsPositive() {
+		// Use amount1 to swap any Token0 ticks < -depositTick1
```
swelf19: either remove or update the comment, like the one above (about -1)

x/dex/keeper/deposit.go
```diff
@@ -167,5 +184,101 @@ func (k Keeper) ExecuteDeposit(
...
+	tickIndex int64,
+	fee uint64,
+	amount0, amount1 math.Int,
+	slopToleranceBPs uint64,
+) (inAmount0, inAmount1, depositAmount0, depositAmount1 math.Int, err error) {
+	feeInt64 := dexutils.MustSafeUint64ToInt64(fee)
+	inAmount0, inAmount1 = amount0, amount1
+	depositAmount0, depositAmount1 = inAmount0, inAmount1
+	swappedToken0 := false
+	if amount0.IsPositive() {
```
swelf19, line 215: How do you think is it possible to extract the code under the `if` condition into a function to reuse it in the second `if`

x/dex/keeper/integration_deposit_doublesided_test.go
```diff
@@ -205,3 +205,110 @@ func (s *DexTestSuite) TestDepositValueAccural() {
...
+	s.assertAliceBalances(0, 0)
+
+	s.assertLiquidityAtTickInt(math.NewInt(40000000), math.NewInt(25440286), 1998, 1)
+}
+
+func (s *DexTestSuite) TestDepositToken1BELWithSwapAll() {
+	s.fundAliceBalances(10, 20)
+	s.fundBobBalances(30, 0)
+
+	// GIVEN TokenA liquidity at tick 10,000
```
swelf19: 10,002?

x/dex/keeper/integration_deposit_doublesided_test.go
```diff
@@ -205,3 +205,110 @@ func (s *DexTestSuite) TestDepositValueAccural() {
...
+	// WHEN alice deposits TokenB at tick -10,003 (BEL)
+	resp := s.aliceDeposits(
+		NewDepositWithOptions(10, 20, -10004, 1,
+			types.DepositOptions{FailTxOnBel: true, SwapOnDeposit: true, SwapOnDepositSlopToleranceBps: 10},
+		),
+	)
+
+	// THEN all of alice's tokenB is swapped
+	// and she deposits ~17.3TokenA and 0TokenB
+	// A = 10 + 20 / 1.0001^10000 = 17.3
```
swelf19: The same, 10002. TickA=-1*tickIndex+fee, correct?
and the ouput correlates with `20 / 1.0001^10002` not `20 / 1.0001^10000`

x/dex/keeper/deposit.go
```diff
@@ -167,5 +184,101 @@ func (k Keeper) ExecuteDeposit(
...
+		// Use Amount0 to swap any Token1 ticks < (-depositTick0 -1 )
+		depositTickToken0 := -tickIndex + feeInt64
+		// subtract 1 from limit price because we can have double-sided liquidity at the deposit tick
+		// we don't want to swap through this liquidity
```
swelf19, line 219: What is the motivation behind? Increase maker market?

x/dex/keeper/integration_deposit_doublesided_test.go
```diff
@@ -205,3 +205,110 @@ func (s *DexTestSuite) TestDepositValueAccural() {
 	s.carolWithdraws(NewWithdrawalInt(math.NewInt(90484150), 0, 10))
 	s.assertCarolBalances(100, 1)
 }
+
+func (s *DexTestSuite) TestDepositToken0BELWithSwapPartial() {
+	s.fundAliceBalances(50, 10)
+	s.fundBobBalances(0, 30)
+
+	// GIVEN TokenA liquidity at tick 2002-2004
```
swelf19: isn't bob depositing B?

x/dex/keeper/deposit.go
```diff
@@ -167,5 +184,101 @@ func (k Keeper) ExecuteDeposit(
...
+		limitPrice0 := types.MustCalcPrice(-depositTickToken0 - 1)
+		tradePairID := types.MustNewTradePairID(pairID.Token0, pairID.Token1)
+
+		swapToken0In, swapToken1Out, orderFilled, err := k.Swap(ctx, tradePairID, amount0, nil, &limitPrice0)
+		if err != nil {
+			return math.ZeroInt(), math.ZeroInt(), math.ZeroInt(), math.ZeroInt(), err
+		}
+
+		if err := CheckSwapOnDepositSlopTolerance(swapToken0In.Amount, swapToken1Out.Amount, limitPrice0, slopToleranceBPs); err != nil {
```
swelf19: I think it's not necessary call the tollerance check if swapToken0In is zero. We can move it under the `swapToken0In.IsPositive()` condition, and maybe even remove `in CheckSwapOnDepositSlopTolerance` if zero check L273

jcompagni10: With new refactor I think it's pretty clear/clean. Let me know what you think

x/dex/keeper/deposit.go
jcompagni10: The invariant here is that liquidity cannot be placed behind enemy lines. Placing liquidity at the same price is fine. The goal of this PR is just to uphold that invariant. From the perspective of super vaults doing one more swap would probably be desirable. But for other use cases it may make sense to limit the number of swaps (and gas cost.)

jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/13850458813

PR #807, opened 2025-01-23
chore: update to `slinky/v1.2.0`

aljo242:
This PR updates neutron to the latest version of `slinky` which includes wasm safe queries.
- updates build rules to use go1.23 (this was auto bumped by updating slinky)
- update lints that are picked up by new go version
- disable `gosec` lint on tests (creating lots of noise that is not needed)


pr0n00gler:
Thanks! But already bumped here: https://github.com/neutron-org/neutron/pull/784

PR #808, opened 2025-01-24
Feat: Add swapMetadata to TickUpdates #NTRN-440

jcompagni10:
Add SwapAmountIn and SwapAmountOut attributes to tick updates for easier indexing of swap events.

### DIFF
--- a/x/dex/keeper/liquidity_test.go
+++ b/x/dex/keeper/liquidity_test.go
@@ -547,6 +547,22 @@ func (s *DexTestSuite) TestSwapExhaustsLOAndLP() {
 	// There should be total of 6 tick updates
 	// (limitOrder, 2x deposit,  2x swap LP, swap LO)
 	s.AssertNEventValuesEmitted(types.TickUpdateEventKey, 6)
+
+	tickUpdates := s.GetAllMatchingEvents(types.TickUpdateEventKey)
+
+	// LimitOrder TickUpdate has correct SwapMetadatrra
+	loTickUpdate := tickUpdates[3]
+	loSwapIn, _ := loTickUpdate.GetAttribute(types.AttributeSwapAmountIn)
+	loSwapOut, _ := loTickUpdate.GetAttribute(types.AttributeSwapAmountOut)
+	s.Equal("10000000", loSwapIn.Value)
+	s.Equal("10000000", loSwapOut.Value)
+
+	// LP TickUpdate has correct SwapMetadatrra
+	lpTickUpdate := tickUpdates[5]
+	lpSwapIn, _ := lpTickUpdate.GetAttribute(types.AttributeSwapAmountIn)
+	lpSwapOut, _ := lpTickUpdate.GetAttribute(types.AttributeSwapAmountOut)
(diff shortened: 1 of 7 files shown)


jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/12944784314

x/dex/keeper/liquidity_test.go
```diff
@@ -547,6 +547,22 @@ func (s *DexTestSuite) TestSwapExhaustsLOAndLP() {
 	// There should be total of 6 tick updates
 	// (limitOrder, 2x deposit,  2x swap LP, swap LO)
 	s.AssertNEventValuesEmitted(types.TickUpdateEventKey, 6)
+
+	tickUpdates := s.GetAllMatchingEvents(types.TickUpdateEventKey)
+
+	// LimitOrder TickUpdate has correct SwapMetadatrra
```
dib542, line 553:
```suggestion
	// LimitOrder TickUpdate has correct SwapMetadata
```
Mistype 😅, is also on other added comment

x/dex/keeper/liquidity_test.go
```diff
@@ -547,6 +547,22 @@ func (s *DexTestSuite) TestSwapExhaustsLOAndLP() {
...
+	loSwapOut, _ := loTickUpdate.GetAttribute(types.AttributeSwapAmountOut)
+	s.Equal("10000000", loSwapIn.Value)
+	s.Equal("10000000", loSwapOut.Value)
+
+	// LP TickUpdate has correct SwapMetadatrra
+	lpTickUpdate := tickUpdates[5]
+	lpSwapIn, _ := lpTickUpdate.GetAttribute(types.AttributeSwapAmountIn)
+	lpSwapOut, _ := lpTickUpdate.GetAttribute(types.AttributeSwapAmountOut)
+	s.Equal("9000000", lpSwapIn.Value)
+	s.Equal("8999100", lpSwapOut.Value)
```
dib542: Maybe add a negative test that the LP swap opposite tick does not contain swap attributes (to prevent double counting)

It's nice to have a "Swap on Deposit" test here 👍

x/dex/keeper/liquidity_test.go
```diff
@@ -547,6 +547,22 @@ func (s *DexTestSuite) TestSwapExhaustsLOAndLP() {
...
 }
```
dib542, line 573: Maybe add a negative test, that the regular deposits does not include any swap attributes

dib542 (dismissed):
This looks good 👍 

This is all we need to enable very easy analysis and indexing of
- pair price
- swap volume 
- LP fees
for timeseries or latest value / 24hr aggregation values.

Note: I previously thought a AmountIn/AmountOut or DepositIn/DepositOut like attribute would also be needed, but such an attribute would only be helpful to compute change in Dex liqudity volume easily, the aggregate of which can be fairly easily calculated (TVL diff over time) and the specific price changes of liquidity over time is not of current interest (and can also still be derived from current data).

dib542 (approved):

pr0n00gler (approved):

PR #824, opened 2025-02-04
fix: off-by-one error in cron periods #ntrn-451

NeverHappened:
Fixes problem when period 1 would not execute schedule each block, only once per two schedules.

https://hadronlabs.atlassian.net/browse/NTRN-451


pr0n00gler (approved):

pr0n00gler:
https://github.com/neutron-org/neutron-tests/actions/runs/13367228420

swelf19 (approved):

PR #847, opened 2025-03-03
Feat: add max out lo event attr

jcompagni10:


jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/13641405576

pr0n00gler (approved):

PR #855, opened 2025-03-09
refactor: replace golang.org/x/exp with stdlib

wangjingcun:
Since Go 1.21, the methods in x/exp used here can already be replaced by methods from the standard library


jcompagni10 (approved):

wangjingcun:
Thanks for your review.

pr0n00gler (approved):

PR #860, opened 2025-03-14
Chore: add request amount in to LO event

jcompagni10:


jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/13850432258

pr0n00gler (approved):

PR #861, opened 2025-03-15
Feat: better telemetry for cron execs

jcompagni10:
fix how timing is reported for cron execution


x/cron/keeper/keeper.go
```diff
@@ -67,13 +69,14 @@ func (k *Keeper) Logger(ctx sdk.Context) log.Logger {
...
 func (k *Keeper) ExecuteReadySchedules(ctx sdk.Context, executionStage types.ExecutionStage) {
-	telemetry.ModuleMeasureSince(types.ModuleName, time.Now(), LabelExecuteReadySchedules)
+	startTime := time.Now()
 	schedules := k.getSchedulesReadyForExecution(ctx, executionStage)
 
 	for _, schedule := range schedules {
 		err := k.executeSchedule(ctx, schedule)
 		recordExecutedSchedule(err, schedule)
 	}
+	telemetry.ModuleMeasureSince(types.ModuleName, startTime, LabelExecuteReadySchedules)
```
mmikev: I think just adding `defer` resolves the problem and it's more cleaner from my perspective (Cosmos SDK [uses](https://github.com/neutron-org/cosmos-sdk/blob/c3892cfa7e2e3dc2295209862b58f1e203a18d07/x/slashing/abci.go#L17) `defer` as well):
https://go.dev/play/p/Eof5hpEyxQX

pr0n00gler (dismissed):

x/cron/keeper/keeper.go
```diff
@@ -205,8 +210,9 @@ func (k *Keeper) executeSchedule(ctx sdk.Context, schedule types.Schedule) error
 			)
 			return err
```
swelf19, line 211: in case of error, you are missing both `telemetry.ModuleMeasureSince` calls, it's better  to use `defer`

x/cron/keeper/keeper.go
```diff
@@ -67,13 +69,14 @@ func (k *Keeper) Logger(ctx sdk.Context) log.Logger {
 // ExecuteReadySchedules gets all schedules that are due for execution (with limit that is equal to Params.Limit)
 // and executes messages in each one
 func (k *Keeper) ExecuteReadySchedules(ctx sdk.Context, executionStage types.ExecutionStage) {
-	telemetry.ModuleMeasureSince(types.ModuleName, time.Now(), LabelExecuteReadySchedules)
+	startTime := time.Now()
```
mmikev: I still don't understand why do you need a separate variable `startTime` here.

Just calling an original method `telemetry.ModuleMeasureSince(types.ModuleName, time.Now(), LabelExecuteReadySchedules)` via `defer` resolves the initial problem

pr0n00gler (changes requested):

x/cron/keeper/keeper.go
jcompagni10: Gotcha. Changed

jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/14497637374

x/cron/keeper/keeper.go
```diff
@@ -205,6 +209,7 @@ func (k *Keeper) executeSchedule(ctx sdk.Context, schedule types.Schedule) error
 			)
 			return err
 		}
+		defer telemetry.ModuleMeasureSince(types.ModuleName, startTimeContract, LabelExecuteCronContract, schedule.Name, msg.Contract)
```
mmikev: Hm, suppose we have two schedules:A and B. A gets called first, B is second.
Doesn't this `defer` call mean, that a measurement for A call contains a time consumption of the B call?
Because the `ModuleMeasureSince` is called only in the end of the whole execution.

jcompagni10: Resloved

swelf19 (approved):

pr0n00gler (approved):

PR #876, opened 2025-03-30
feat: allow indexers to know pool_id without a chain query

dib542:
Currently there is no way to determine a dex pool's properties (`TickIndex` and `Fee`) from every possible dex deposit and withdrawal without performing a chain query (eg. `neutrond q dex show-pool-metadata 1`).

This is very inconvenient for indexers that are attempting to stream real-time data about user<>dex positions and user<>vault<>dex as they are being made, as the requirement to query the chain separately both complicates and slows down the process to stream the data. 

The easiest fix is to add the pool ID to the deposit and withdrawal events. An alternative approach would be a separate "new pool" event that would need to be before the deposit or withdrawal, however this is less convenient. It is helpful that the pool ID is attached to the related deposits and withdrawals as it provides an easy way to calculate the running total shares of each pool (sum of `SharesMinted - SharesRemoved`).

### DIFF
--- a/x/dex/keeper/deposit.go
+++ b/x/dex/keeper/deposit.go
@@ -176,6 +176,7 @@ func (k Keeper) ExecuteDeposit(
 			depositAmount1,
 			inAmount0,
 			inAmount1,
+			pool.Id,
 			outShares.Amount,
 		)
 		events = append(events, depositEvent)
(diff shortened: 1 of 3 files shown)


jcompagni10 (approved):
lgtm

jcompagni10 (changes requested):
pls fix linter errors

jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/14243795020

dib542:
@jcompagni10 fixed linter errors 🙏

jcompagni10 (approved):

dib542:
Closed in favor of #912

PR #892, opened 2025-04-28
Chore: remove ccv related stuff

pr0n00gler:
This PR remove unnecessary keepers, test helpers, unused methods, etc. mostly related to interchain security (but also it removes outdated Auction keepers that is not used anymore)


pr0n00gler:
https://github.com/neutron-org/neutron-tests/actions/runs/14707873099

app/app.go
```diff
@@ -985,7 +908,6 @@ func New(
 		dexModule,
 		marketmapModule,
 		oracleModule,
-		auction.NewAppModule(appCodec, app.AuctionKeeper),
 		consensus.NewAppModule(appCodec, app.ConsensusParamsKeeper),
 		// always be last to make sure that it checks for all invariants and not only part of them
 		crisis.NewAppModule(&app.CrisisKeeper, skipGenesisInvariants, app.GetSubspace(crisistypes.ModuleName)),
```
swelf19, line 914: out of scope, but, is there a reason, why `stateverifier` is the last module. If not, lets change the order, set crisis last here

swelf19 (commented):
We removed the auction and consumer modules, but it seems that both stores are still present. There's probably no point in keeping them in the state - maybe it makes sense to add them to the removal list for the next upgrade?

app/app.go
mmikev: Fixed

pr0n00gler:
> We removed the auction and consumer modules, but it seems that both stores are still present. There's probably no point in keeping them in the state - maybe it makes sense to add them to the removal list for the next upgrade?

I'll do it in a separate PR related to v7 release

swelf19 (approved):

PR #899, opened 2025-04-29
Fix rounding issue in place limit order taker test

GarmashAlex:
Changed precision tolerance in swapAmountOut test from 0 to 1 to account for
minor rounding differences that occur during price calculations and token
conversions.


pr0n00gler:
Thanks for fix, but it's outdated

PR #902, opened 2025-04-30
fix: iavl-disable-fastnode flag

swelf19:
Without this PR, the node ignores the iavl-disable-fastnode option in the configuration file and the command-line argument.


pr0n00gler:
https://github.com/neutron-org/neutron-tests/actions/runs/15468608897

pr0n00gler (approved):

PR #903, opened 2025-04-30
feat: add optimistic block production flag and config variable #NTRN-516

joldie777:
Task: https://hadronlabs.atlassian.net/browse/NTRN-516

Related PRs:
https://github.com/neutron-org/cosmos-sdk/pull/101


joldie777:
Test run: https://github.com/neutron-org/neutron-tests/actions/runs/14764951973

Optimistic execution for test run is enabled in the app.toml by [this](https://github.com/neutron-org/neutron-integration-tests/blob/main/setup/scripts/4_update-config.sh#L25) code

pr0n00gler (dismissed):

cmd/neutrond/root.go
```diff
@@ -307,6 +302,22 @@ func (ac appCreator) newApp(
 		baseapp.SetIndexEvents(cast.ToStringSlice(appOpts.Get(server.FlagIndexEvents))),
 		baseapp.SetSnapshot(snapshotStore, snapshottypes.SnapshotOptions{Interval: cast.ToUint64(appOpts.Get(server.FlagStateSyncSnapshotInterval)), KeepRecent: cast.ToUint32(appOpts.Get(server.FlagStateSyncSnapshotKeepRecent))}),
 		baseapp.SetChainID(chainID),
+	}
+
+	if isEnabled, err := cast.ToBoolE(appOpts.Get(server.FlagOptimisticExecutionEnabled)); err == nil && isEnabled {
```
sotnikov-s: don't think it's reasonable to use `cast.ToBoolE` here without handling the error: in the error cases the returned bool is always false. in all other cases above we use the simpler `cast.ToBool` version and use the value as is. I don't think we should handle error here and we can just get rid of the `cast.ToBoolE` and use the `cast.ToBool` func. what do you think?

sotnikov-s (dismissed):

NeverHappened (dismissed):

cmd/neutrond/root.go
joldie777: Done

sotnikov-s (approved):

pr0n00gler (approved):

PR #905, opened 2025-04-30
Feat: set ntrn metadata in genesis

pr0n00gler:
Sets NTRN metadata in the bank module in the genesis


pr0n00gler:
https://github.com/neutron-org/neutron-tests/actions/runs/14755123817

PR #908, opened 2025-05-07
fix: set name, symbol and display for created denoms metadata #NTRN-515

joldie777:
Task: https://hadronlabs.atlassian.net/browse/NTRN-515


x/tokenfactory/migrations/v3/store.go
mmikev, line 1: Since the storage layout itself (and state structures) is not changing, i think it should be an upgrade handler that just updates metadata structure not the migration handler. What do you think?

joldie777: Moved

NeverHappened (dismissed):
LGTM!

app/upgrades/v7.0.0/upgrades.go
```diff
@@ -38,7 +40,31 @@ func CreateUpgradeHandler(
...
 	}
 }
+
+func UpgradeDenomsMetadata(ctx sdk.Context, tk tokenfactorykeeper.Keeper, bk bankkeeper.Keeper) error {
+	iterator := tk.GetAllDenomsIterator(ctx)
+	defer iterator.Close()
+	for ; iterator.Valid(); iterator.Next() {
+		denom := string(iterator.Value())
+
+		metadata, _ := bk.GetDenomMetaData(ctx, denom)
```
mmikev: Unhandled error?

joldie777: Missed that. Added check

app/upgrades/v7.0.0/upgrades.go
```diff
@@ -38,7 +40,34 @@ func CreateUpgradeHandler(
...
+
+func UpgradeDenomsMetadata(ctx sdk.Context, tk tokenfactorykeeper.Keeper, bk bankkeeper.Keeper) error {
+	iterator := tk.GetAllDenomsIterator(ctx)
+	defer iterator.Close()
+	for ; iterator.Valid(); iterator.Next() {
+		denom := string(iterator.Value())
+
+		metadata, found := bk.GetDenomMetaData(ctx, denom)
+		if !found {
+			continue
```
NeverHappened: i don't wanna be too intrusive here, but i think we either return err here or at least log that something is not found since it's not a good behavior

joldie777: Added logging

mmikev: Should we continue here at all?
In what cases the metadata cannot be found?

NeverHappened: Agreed, just return err

sotnikov-s: why don't we just use [GetAllDenomMetaData](https://github.com/neutron-org/cosmos-sdk/blob/96d7aa456bfeeed456f98f3de6010e0840d1e5c9/x/bank/keeper/keeper.go#L243-L252) here and iterate over the result?

joldie777: Metadata cannot be found when denom is transferred from another chain via IBC. In that case no metadata for that denom will be created in the store.
But thanks to @sotnikov-s there is no need to care about it anymore!
Refactored.

sotnikov-s (dismissed):

pr0n00gler (changes requested):
Fix linter, please

sotnikov-s (dismissed):

app/upgrades/v7.0.0/upgrades.go
```diff
@@ -38,7 +39,29 @@ func CreateUpgradeHandler(
...
+	allDenomMetadata := bk.GetAllDenomMetaData(ctx)
+
+	for _, metadata := range allDenomMetadata {
+		denom := metadata.Base
+
+		metadata.Name = denom
+		metadata.Symbol = denom
+		metadata.Display = denom
+
+		bk.SetDenomMetaData(ctx, metadata)
```
NeverHappened: not sure, but aren't we overriding the metadata of all denoms this way? not only tokenfactory, but other denoms as well. Old way of doing seemed to be more correct, since we fetched tokenfactory denoms

NeverHappened (changes requested):

app/upgrades/v7.0.0/upgrades.go
joldie777: Added check for emptiness of these fields

NeverHappened (approved):

pr0n00gler (approved):

PR #917, opened 2025-05-16
fix: err and outofgas recovery #ntrn-521

NeverHappened:
Intended behavior
 - gas always consumed in block and track (when err or no err)
 - contract state is recovered - that means that we use child context and write only if no errors and gas problems
 - on error block cancels tx, track ignores it
 - on inner out of gas we return err and DO NOT propagate outOfGas
 - on outer out of gas we DO propagate outOfGas as normal

https://hadronlabs.atlassian.net/browse/NTRN-521


x/tokenfactory/keeper/before_send_test.go
```diff
@@ -253,36 +253,69 @@ func (suite *KeeperTestSuite) TestNonWhitelistedHooksNotCalled() {
...
+			useHookedDenom:  true,
+			gasLimit:        30_000_000,
+			expectedError:   false,
+			expectedPanic:   false,
 		},
 		{
-			name:            "sending tokenfactory denom from account to account with infinite contract should panic",
-			wasmFile:        "./testdata/infinite_track_beforesend.wasm",
-			useFactoryDenom: true,
+			name:            "sending tokenfactory denom with infinite contract should not return error on blockBeforeSend",
```
swelf19: Actually we are expecting the error in this particular test case

x/tokenfactory/keeper/before_send_test.go
```diff
@@ -328,20 +362,44 @@ func (suite *KeeperTestSuite) TestInfiniteTrackBeforeSend() {
 			_, err = suite.msgServer.SetBeforeSendHook(suite.ChainA.GetContext(), types.NewMsgSetBeforeSendHook(suite.ChainA.SenderAccounts[0].SenderAccount.GetAddress().String(), factoryDenom, cosmwasmAddress.String()))
 			suite.Require().NoError(err, "test: %v", tc.name)
 
-			ctx := suite.ChainA.GetContext().WithGasMeter(storetypes.NewGasMeter(30_000_000))
+			ctx := suite.ChainA.GetContext().WithGasMeter(storetypes.NewGasMeter(tc.gasLimit))
 
 			if tc.blockBeforeSend {
```
swelf19, line 381: Why, in the case where tc.blockBeforeSend == true, are we testing transfers between regular accounts, while in the case where tc.blockBeforeSend == false, we’re testing transfers between module accounts?

x/tokenfactory/keeper/before_send_test.go
```diff
@@ -328,20 +362,44 @@ func (suite *KeeperTestSuite) TestInfiniteTrackBeforeSend() {
...
-				suite.Require().NoError(err)
+				if tc.expectedPanic {
+					suite.Require().Panics(func() {
+						err = suite.GetNeutronZoneApp(suite.ChainA).BankKeeper.SendCoinsFromModuleToModule(ctx, icqtypes.ModuleName, dextypes.ModuleName, tokenToSend)
+					})
+				} else {
+					err = suite.GetNeutronZoneApp(suite.ChainA).BankKeeper.SendCoinsFromModuleToModule(ctx, icqtypes.ModuleName, dextypes.ModuleName, tokenToSend)
+				}
+
+				if tc.expectedError {
```
swelf19: Is it even possible to get an error when tc.blockBeforeSend == false?
It seems like this could be a bit misleading.
It looks like you can just leave assert suite.Require().NoError(err) instead of the whole if condition.

x/tokenfactory/keeper/before_send.go
```diff
@@ -134,46 +128,87 @@ func (k Keeper) callBeforeSendListener(ctx context.Context, from, to sdk.AccAddr
...
-						Amount: CWCoinFromSDKCoin(coin),
-					},
-				}
-				msgBz, err = json.Marshal(msg)
-			}
+			gasConsumed, err := k.callBeforeSendForCoin(sdkCtx, blockBeforeSend, from, to, coin, cwAddr)
+
+			// consume gas used for calling contract to the parent ctx
+			// note that we consume gas even in case of error
+			sdkCtx.GasMeter().ConsumeGas(gasConsumed, "track before send gas")
```
sotnikov-s: not `track` anymore

```suggestion
			sdkCtx.GasMeter().ConsumeGas(gasConsumed, "before send gas")
```

x/tokenfactory/keeper/before_send_test.go
```diff
@@ -310,10 +343,11 @@ func (suite *KeeperTestSuite) TestInfiniteTrackBeforeSend() {
...
-			if tc.useFactoryDenom {
+			if tc.useHookedDenom {
 				tokenToSend = sdk.NewCoins(sdk.NewInt64Coin(factoryDenom, 100))
 			} else {
-				tokenToSend = tc.tokenToSend
+				// denom without hook attached
+				tokenToSend = sdk.NewCoins(sdk.NewInt64Coin("foo", 1000000))
 			}
 
 			// send the mint module tokenToSend
```
sotnikov-s: please fix/update the comment

x/tokenfactory/keeper/before_send_test.go
```diff
@@ -328,20 +362,44 @@ func (suite *KeeperTestSuite) TestInfiniteTrackBeforeSend() {
...
 
-			ctx := suite.ChainA.GetContext().WithGasMeter(storetypes.NewGasMeter(30_000_000))
+			ctx := suite.ChainA.GetContext().WithGasMeter(storetypes.NewGasMeter(tc.gasLimit))
 
 			if tc.blockBeforeSend {
-				err = suite.GetNeutronZoneApp(suite.ChainA).BankKeeper.SendCoins(ctx, suite.ChainA.SenderAccounts[0].SenderAccount.GetAddress(), suite.ChainA.SenderAccounts[1].SenderAccount.GetAddress(), tokenToSend)
-				suite.Require().Error(err)
+				if tc.expectedPanic {
+					suite.Require().Panics(func() {
+						err = suite.GetNeutronZoneApp(suite.ChainA).BankKeeper.SendCoins(ctx, suite.ChainA.SenderAccounts[0].SenderAccount.GetAddress(), suite.ChainA.SenderAccounts[1].SenderAccount.GetAddress(), tokenToSend)
```
sotnikov-s: this is soooo hard to read code lines like that. could you please add line breaks by func arguments e.g. like follows here and in all cases the lines are long and arguments are complex?
image

x/tokenfactory/keeper/before_send_test.go
```diff
@@ -328,20 +362,44 @@ func (suite *KeeperTestSuite) TestInfiniteTrackBeforeSend() {
...
+				}
 
 				// send should happen regardless of trackBeforeSend results
-				distributionModuleAddress := suite.GetNeutronZoneApp(suite.ChainA).AccountKeeper.GetModuleAddress(dextypes.ModuleName)
-				distributionModuleBalances := suite.GetNeutronZoneApp(suite.ChainA).BankKeeper.GetAllBalances(suite.ChainA.GetContext(), distributionModuleAddress)
-				suite.Require().True(distributionModuleBalances.Equal(tokenToSend))
+				// except if panics
+				if !tc.expectedPanic {
+					distributionModuleAddress := suite.GetNeutronZoneApp(suite.ChainA).AccountKeeper.GetModuleAddress(dextypes.ModuleName)
+					distributionModuleBalances := suite.GetNeutronZoneApp(suite.ChainA).BankKeeper.GetAllBalances(suite.ChainA.GetContext(), distributionModuleAddress)
```
sotnikov-s: it's a dex module balance not distribution one

x/tokenfactory/keeper/before_send_test.go
```diff
@@ -253,36 +253,69 @@ func (suite *KeeperTestSuite) TestNonWhitelistedHooksNotCalled() {
...
 		name            string
 		wasmFile        string
-		tokenToSend     sdk.Coins
-		useFactoryDenom bool
+		useHookedDenom  bool
 		blockBeforeSend bool
+		gasLimit        uint64
 		expectedError   bool
+		expectedPanic   bool
 	}{
```
sotnikov-s: could you please write some descriptions for these parameters with side effects mention? this is really hard to understand the whole test flow and peculiarities by just the field names as they are. e.g.
```go
// blockBeforeSend defines whether to use a block before send or track before send. In
// case of track before send, the transfer is done between module accounts because blah-blah...
blockBeforeSend bool
```

x/tokenfactory/keeper/before_send_test.go
```diff
@@ -253,36 +253,69 @@ func (suite *KeeperTestSuite) TestNonWhitelistedHooksNotCalled() {
...
 		{
-			name:            "sending tokenfactory denom from module to module with infinite contract should panic",
+			name:            "sending tokenfactory denom from module to module with infinite contract should not return error on trackBeforeSend",
 			wasmFile:        "./testdata/infinite_track_beforesend.wasm", // https://github.com/neutron-org/neutron-dev-contracts/tree/chore/additional-tf-test-contracts/contracts/infinite-track-beforesend
```
sotnikov-s: please move this value to a const with the long comment attached to the const description for brevity and cleanness

x/tokenfactory/keeper/before_send_test.go
```diff
@@ -253,36 +253,69 @@ func (suite *KeeperTestSuite) TestNonWhitelistedHooksNotCalled() {
...
 			blockBeforeSend: true,
+			useHookedDenom:  true,
+			gasLimit:        30_000_000,
+			expectedError:   true,
+			expectedPanic:   false,
+		},
+		{
+			name:            "block_before_send: sending tokenfactory denom from module to module with infinite contract should panic when outer layer gas is breached on trackBeforeSend",
+			wasmFile:        "./testdata/infinite_track_beforesend.wasm", // https://github.com/neutron-org/neutron-dev-contracts/tree/chore/additional-tf-test-contracts/contracts/infinite-track-beforesend
+			blockBeforeSend: false,
```
sotnikov-s: this is said in the name that it is `block_before_send` whereas `blockBeforeSend: false`. please fix it and make sure other test case names are correct

x/tokenfactory/keeper/before_send_test.go
```diff
@@ -253,36 +253,69 @@ func (suite *KeeperTestSuite) TestNonWhitelistedHooksNotCalled() {
...
 		{
-			name:            "sending non-tokenfactory denom from module to module with infinite contract should not panic",
-			wasmFile:        "./testdata/infinite_track_beforesend.wasm",
-			tokenToSend:     sdk.NewCoins(sdk.NewInt64Coin("foo", 1000000)),
-			useFactoryDenom: false,
+			name:            "block_before_send: sending tokenfactory denom with infinite contract should panic when outer layer gas is breached on blockBeforeSend",
+			wasmFile:        "./testdata/infinite_track_beforesend.wasm", // https://github.com/neutron-org/neutron-dev-contracts/tree/chore/additional-tf-test-contracts/contracts/infinite-track-beforesend
+			blockBeforeSend: true,
+			useHookedDenom:  true,
+			gasLimit:        300_000, // lower than 500_000 inner constant, so it should trigger outer context outOfGas panic
```
sotnikov-s: don't really understand the reasoning of the panic: `300_000` is lower than `500_000` so it should panic. could you please elaborate?

x/tokenfactory/keeper/before_send_test.go
```diff
@@ -253,36 +253,69 @@ func (suite *KeeperTestSuite) TestNonWhitelistedHooksNotCalled() {
...
+			name:            "sending non subscribed denom from module to module with infinite contract should not panic or return error",
+			wasmFile:        "./testdata/infinite_track_beforesend.wasm", // https://github.com/neutron-org/neutron-dev-contracts/tree/chore/additional-tf-test-contracts/contracts/infinite-track-beforesend
+			useHookedDenom:  false,
+			blockBeforeSend: false,
+			gasLimit:        30_000_000,
+			expectedError:   false,
+			expectedPanic:   false,
+		},
+		{
+			name:            "Try using no 100 contract - sends that more then 100 gets denied",
```
sotnikov-s: don't like the description: it's a bit incorrect in grammar and doesn't align with the style of other test case names. here's an idea from my side. if you don't like it I'd like to know your thoughts
```suggestion
			name:            "sending tokenfactory denom with expected error from contract",
```

sotnikov-s (changes requested):

x/tokenfactory/keeper/before_send_test.go
NeverHappened: It was like this before as well. I think the reason is because module to module send does not call block_before_send at all

x/tokenfactory/keeper/before_send_test.go
NeverHappened: there is an inner gas limit
`var BeforeSendHookGasLimit = uint64(500_000)`
it's there to avoid infinite loops in case where contract is being called outside of transactions (like in begin blocker or smth)
inner loop context cannot panic, it catches inner out of gas.
so this situation simulates case where inner loop consumes 500_000 gas, exits with err because it caught out of gas err, and then in the outer loop consumes gas, and because of 300_000 limit it should panic

NeverHappened:
https://github.com/neutron-org/neutron-tests/actions/runs/15538078678

PR #932, opened 2025-06-19
feat: disable ibc fees #ntrn-503

NeverHappened:
Add ability to disable ibc fees using feerefunder.Params.FeeEnabled param.

https://hadronlabs.atlassian.net/browse/NTRN-503

PR's:
- [neutron](https://github.com/neutron-org/neutron/pull/932) - add EnableFee param
- [integration-tests](https://github.com/neutron-org/neutron-integration-tests/pull/389) - add tests when fee is disabled
- [neutronjsplus](https://github.com/neutron-org/neutronjsplus/pull/86) - add param to update params proposal
- [neutronjs](https://github.com/neutron-org/neutronjs/pull/18) - add param to neutronjs
- [neutron-std](https://github.com/neutron-org/neutron-std/pull/14) - regen neutron feerefunder types
- [neutron-sdk](https://github.com/neutron-org/neutron-sdk/pull/175) - use updated neutron std to modify submit_tx to use Option for fees (None if no fees)
- [neutron-dev-contracts](https://github.com/neutron-org/neutron-dev-contracts/pull/77) - add ability to set empty fees


NeverHappened:
https://github.com/neutron-org/neutron-tests/actions/runs/15829547797

app/upgrades/v8.0.0/upgrades_test.go
```diff
@@ -0,0 +1,51 @@
...
+
+func TestKeeperTestSuite(t *testing.T) {
+	suite.Run(t, new(UpgradeTestSuite))
+}
+
+func (suite *UpgradeTestSuite) SetupTest() {
+	suite.IBCConnectionTestSuite.SetupTest()
+}
+
+func (suite *UpgradeTestSuite) TestUpgradeDenomsMetadata() {
```
joldie777: `TestUpgradeFeeKeeperParams` seems like a more appropriate name

x/feerefunder/keeper/keeper.go
```diff
@@ -225,6 +231,14 @@ func (k Keeper) removeFeeInfo(ctx sdk.Context, packetID types.PacketID) {
 func (k Keeper) checkFees(ctx sdk.Context, fees types.Fee) error {
 	params := k.GetParams(ctx)
 
+	if err := fees.Validate(); err != nil {
+		return errors.Wrapf(err, "fees do not pass validation")
+	}
+
+	if fees.AckFee.IsZero() || fees.TimeoutFee.IsZero() {
```
joldie777, line 246: The check seems redundant since it is already done [here](https://github.com/neutron-org/neutron/blob/feat/feerefunder-validation-zero/x/feerefunder/types/fee.go#L56).

x/feerefunder/types/fee.go
```diff
@@ -33,9 +33,9 @@ func (m Fee) Total() sdk.Coins {
 
 // Validate asserts that each Fee is valid:
 // * RecvFee must be zero;
-// * AckFee and TimeoutFee must be non-zero
```
joldie777, line 36: Why removed?

x/feerefunder/types/params.go
```diff
@@ -14,27 +14,31 @@ import (
...
+	return Params{MinFee: minFee, FeeEnabled: feeEnabled}
 }
 
 // DefaultParams returns a default set of parameters
 func DefaultParams() Params {
-	return NewParams(DefaultFees)
+	return NewParams(DefaultFees, DefaultFeeEnabled)
 }
 
 // ParamSetPairs get the params.ParamSet
```
joldie777, line 44: Shouldn't `paramtypes.NewParamSetPair(KeyFeeEnabled, &p.FeeEnabled, validateFeeEnabled)` be added to the `paramtypes.ParamSetPairs` slice? Although `ParamSetPairs` method is not used...

x/interchaintxs/keeper/msg_server_test.go
```diff
@@ -305,38 +309,6 @@ func TestMsgSubmitTXValidate(t *testing.T) {
 			},
 			sdkerrors.ErrInvalidCoins,
 		},
-		{
```
joldie777, line 308: Why were these test cases removed?

x/transfer/keeper/keeper_test.go
```diff
@@ -249,38 +249,6 @@ func TestMsgTransferValidate(t *testing.T) {
 			},
 			errors.ErrInvalidCoins,
 		},
-		{
-			"zero ack fee",
```
joldie777, line 253: Why were these test cases removed?

x/feerefunder/keeper/keeper.go
```diff
@@ -60,14 +60,20 @@ func (k Keeper) LockFees(ctx context.Context, payer sdk.AccAddress, packetID typ
...
-	if _, ok := k.channelKeeper.GetChannel(c, packetID.PortId, packetID.ChannelId); !ok {
-		return errors.Wrapf(channeltypes.ErrChannelNotFound, "channel with id %s and port %s not found", packetID.ChannelId, packetID.PortId)
+	params := k.GetParams(c)
+	if !params.FeeEnabled {
+		k.Logger(c).Debug("Skip locking fees because fee disabled", "packetID", packetID, "fee", fee)
+		return nil
 	}
 
 	if err := k.checkFees(c, fee); err != nil {
 		return errors.Wrapf(err, "failed to lock fees")
```
sotnikov-s: not your code but still would be good to rephrase
```suggestion
		return errors.Wrapf(err, "fees check failed")
```

x/feerefunder/keeper/keeper_test.go
```diff
@@ -191,6 +195,30 @@ func TestKeeperLockFees(t *testing.T) {
...
+	})
+	require.Error(t, err)
+
+	// timeout fee is empty
+	err = k.LockFees(ctx, payer, packet, types.Fee{
+		RecvFee:    nil,
+		AckFee:     sdk.NewCoins(sdk.NewCoin("denom1", math.NewInt(1))),
+		TimeoutFee: nil,
+	})
+	require.Error(t, err)
```
sotnikov-s: would be nice to assert the result errors are of the "ack fee or timeout fee is zero" kind

x/feerefunder/keeper/keeper_test.go
```diff
@@ -225,10 +253,10 @@ func TestDistributeAcknowledgementFee(t *testing.T) {
 		PortId:    "transfer",
 		Sequence:  1,
 	}
-	panicErrorToCatch := errors.Wrapf(errors.Wrapf(sdkerrors.ErrKeyNotFound, "no fee info for the given channelID = %s, portID = %s and sequence = %d", invalidPacket.ChannelId, invalidPacket.PortId, invalidPacket.Sequence), "no fee info")
-	assert.PanicsWithError(t, panicErrorToCatch.Error(), func() { k.DistributeAcknowledgementFee(ctx, receiver, invalidPacket) })
+	k.DistributeAcknowledgementFee(ctx, receiver, invalidPacket)
+	assert.NotPanics(t, func() { k.DistributeAcknowledgementFee(ctx, receiver, invalidPacket) })
```
sotnikov-s: why do the same `k.DistributeAcknowledgementFee(ctx, receiver, invalidPacket)` operation twice?

x/feerefunder/keeper/keeper_test.go
```diff
@@ -254,9 +282,18 @@ func TestDistributeAcknowledgementFee(t *testing.T) {
 		),
 	}, ctx.EventManager().Events())
 
-	feeInfo, err := k.GetFeeInfo(ctx, packet)
+	feeInfo, found := k.GetFeeInfo(ctx, packet)
 	require.Nil(t, feeInfo)
-	require.ErrorContains(t, err, "no fee info")
+	require.False(t, found, "no fee info")
```
sotnikov-s: same at the line 361
```suggestion
	require.False(t, found, "no fee is expected to be info")
```

x/feerefunder/keeper/keeper.go
```diff
@@ -100,10 +106,10 @@ func (k Keeper) DistributeAcknowledgementFee(ctx context.Context, receiver sdk.A
 	c := sdk.UnwrapSDKContext(ctx)
 
 	k.Logger(c).Debug("Trying to distribute ack fee", "packetID", packetID)
```
sotnikov-s, line 112: would be useful to mention in the `DistributeAcknowledgementFee` and `DistributeTimeoutFee` descriptions that if there's no fee stored for the packedID these are no-ops

x/feerefunder/keeper/keeper_test.go
sotnikov-s: why do these checks at all? prior to this PR it made sense because we asserted that there were no fees locked yet for the packet. now we do this NotPanics check at the end of this test case at the line 296. same at the line 331 in the `TestDistributeTimeoutFee`

x/feerefunder/types/params.go
```diff
@@ -44,7 +48,15 @@ func (p *Params) ParamSetPairs() paramtypes.ParamSetPairs {
 
 // Validate validates the set of params
 func (p Params) Validate() error {
-	return p.MinFee.Validate()
+	if err := validateFee(p.MinFee); err != nil {
+		return err
```
sotnikov-s:
```suggestion
		return fmt.Errorf("invalid minFee: %w", err)
```

x/feerefunder/types/fee.go
NeverHappened: you're correct, this was for previous version of the code, now it makes sense to return it back

x/transfer/keeper/keeper_test.go
NeverHappened: You're right, I can get these tests back with latest code.

x/feerefunder/keeper/keeper.go
NeverHappened: wdym? i have log like this below:
`k.Logger(c).Debug("Nothing to distribute for ack fee", "packetID", packetID)`

x/feerefunder/keeper/keeper_test.go
NeverHappened: i thought it would be nice to check that missing fee info behaviour is handled correctly, since it would be the case sometimes. Not sure if need to delete this, but i can

joldie777 (dismissed):

x/feerefunder/keeper/keeper.go
sotnikov-s: in the descriptions = in the comments to the methods

x/feerefunder/keeper/keeper_test.go
sotnikov-s: > it would be the case sometimes

would it? could you please elaborate?

if leaving as is, could you please add a comment that describes the check then? so it's clear what this check is about. otherwise it looks unexpected: we never sent a packet yet but try to distribute some fee

sotnikov-s (dismissed):

joldie777 (dismissed):

x/transfer/keeper/keeper.go
mmikev, line 1: You still have fees validation on L41, meaning a contract won't be able to send a transfer message with 0 fees. Please, check it by writing an integration test + fix the issue if i'm right. Also please write a test that would check that a contract is able to send a native standard IBC transfer message with fees disabled, not our custom transfer message

pr0n00gler (changes requested):

proto/neutron/feerefunder/params.proto
```diff
@@ -11,4 +11,5 @@ message Params {
   option (gogoproto.goproto_stringer) = false;
 
   Fee min_fee = 1 [(gogoproto.nullable) = false];
+  bool fee_enabled = 2;
```
mmikev, line 14: Since you are introducing a new param, you are basically changing the storage layout. So it must be a module's migration not the upgrade handler.
What do you think?

pr0n00gler (changes requested):

proto/neutron/feerefunder/params.proto
mmikev: Like here: https://github.com/neutron-org/neutron/pull/671/files#diff-ccca7b6b13d95c27524a44a3c3931ea113765a5c393d1348a86948819cc840c5

pr0n00gler (dismissed):

pr0n00gler:
Attach test run, please

x/feerefunder/types/tx.go
```diff
@@ -32,7 +32,11 @@ func (msg *MsgUpdateParams) Validate() error {
 		return errorsmod.Wrap(err, "authority is invalid")
 	}
 
-	if err := msg.Params.MinFee.Validate(); err != nil {
+	if err := msg.Params.Validate(); err != nil {
+		return err
+	}
+
+	if err := msg.Params.MinFee.ValidateFeePresent(); err != nil {
```
sotnikov-s, line 39: why isn't Params.MinFee.ValidateFeePresent a part of the Params.Validate method? I think it makes sense to do it this way

NeverHappened: Well, it messes up validation when fee are disabled. And I didn't figure out better way to do this

sotnikov-s (dismissed):

NeverHappened:
https://github.com/neutron-org/neutron-tests/actions/runs/16222021301

PR #934, opened 2025-06-26
refactor: use slices.Sort to simplify code

friendlyping:
use slices.Sort to simplify code

### DIFF
--- a/utils/slice_helper.go
+++ b/utils/slice_helper.go
@@ -1,6 +1,7 @@
 package utils
 
 import (
+	"slices"
 	"sort"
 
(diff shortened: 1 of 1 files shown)


utils/slice_helper.go
```diff
@@ -1,6 +1,7 @@
 package utils
 
 import (
+	"slices"
 	"sort"
```
mmikev, line 5: Please, remove that unused import, because it breaks the build

pr0n00gler (changes requested):

pr0n00gler:
Closing due to inactivity

PR #936, opened 2025-07-02
Add better accounting checks and fix tranche key generation

jcompagni10:
Switch to more reliable method for trancheKey generation.
Add more redundant accounting checks to ensure tranche value cannot be manipulated.


jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/16027410742

app/upgrades/v6.0.3/constants.go
mmikev, line 1: We don't need v6 upgrade in the main branch, remove it, please. We already have v7 upgrade handler there

pr0n00gler (changes requested):

pr0n00gler (approved):

PR #945, opened 2025-07-17
Update broken IBC transfer handover doc link in README

MozirDmitriy:
Replaced the outdated link to the Neutron IBC transfer results handover documentation in x/ibc-hooks/README.md with the current, working URL. This ensures that users can access up-to-date information about the sudo callback mechanism in the Transfer module.


pr0n00gler:
Thanks! But if was fixed here: https://github.com/neutron-org/neutron/pull/944

PR #947, opened 2025-07-28
Fix: incorrect panic message for price lookup

jcompagni10:
Sign on panic message is backwards


jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/16578589127

pr0n00gler (approved):

PR #978, opened 2025-09-30
Feat: proxygasmeter for tokenfactory

pr0n00gler:
Backport of https://github.com/MANTRA-Chain/mantrachain/pull/438

### DIFF
--- a/utils/proxygasmeter.go
+++ b/utils/proxygasmeter.go
@@ -1,0 +1,81 @@
+// https://github.com/MANTRA-Chain/mantrachain/pull/438
+
+package utils
+
+import (
+	"fmt"
+	"math"
+
+	storetypes "cosmossdk.io/store/types"
+)
+
+var _ storetypes.GasMeter = &ProxyGasMeter{}
+
+// ProxyGasMeter wraps another GasMeter, but enforces a lower gas limit.
+// Gas consumption is delegated to the wrapped GasMeter, so it won't risk losing gas accounting compared to standalone
+// gas meter.
+type ProxyGasMeter struct {
+	storetypes.GasMeter
+	limit storetypes.Gas
+}
+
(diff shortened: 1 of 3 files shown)


utils/proxygasmeter.go
```diff
@@ -0,0 +1,81 @@
...
+	consumed, overflow := addUint64Overflow(pgm.GasMeter.GasConsumed(), amount)
+	if overflow {
+		panic(storetypes.ErrorGasOverflow{Descriptor: descriptor})
+	}
+
+	if consumed > pgm.limit {
+		panic(storetypes.ErrorOutOfGas{Descriptor: descriptor})
+	}
+
+	pgm.GasMeter.ConsumeGas(amount, descriptor)
```
swelf19: It looks like there is a small bug here.
If consumed exceeds the new limit, then:

We will get a panic at line L73.

The consumed gas will not be recorded in the gas meter (L76).

As a result, both methods - IsPastLimit and IsOutOfGas - will return `false` after recovering from the panic, which doesn't seem correct.
You can verify this with a test.

```
func TestProxyGasMeter2(t *testing.T) {
	baseGas := uint64(1000)
	limit := uint64(100)

	bgm := storetypes.NewGasMeter(baseGas)
	bgm.ConsumeGas(500, "ff")
	pgm := NewProxyGasMeter(bgm, limit)

	require.Panics(t, func() {
		pgm.ConsumeGas(101, "")
	})
	require.True(t, pgm.IsOutOfGas())
}

```

Suggestion: change the execution order and remove addUint64Overflow, since overflow checking is already performed in the parent gas meter.

```suggestion
    pgm.GasMeter.ConsumeGas(amount, descriptor)

	if pgm.GasConsumed() > pgm.limit {
		panic(storetypes.ErrorOutOfGas{Descriptor: descriptor})
	}
```
As in the parent method, we first record the consumption and then check the limits.

PR #986, opened 2025-10-11
refactor: use reflect.TypeFor

rocksload:
Try to use better api reflect.TypeFor instead of reflect.TypeOf when we have known the type.

More info https://github.com/golang/go/issues/60088

### DIFF
--- a/testutil/common/nullify/nullify.go
+++ b/testutil/common/nullify/nullify.go
@@ -9,8 +9,8 @@ import (
 )
 
 var (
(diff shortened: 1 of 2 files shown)


rocksload:
The CI error look like no relationship with this PR.

rocksload:
@pr0n00gler Hi, Could you please review this PR at your convenience? Thank you very much.

pr0n00gler (approved):

PR #989, opened 2025-10-27
Fix: balance iter

jcompagni10:
Fix BalanceMap iteration in fractional banker to ensure determinism.

### DIFF
--- a/x/dex/keeper/fractional_banker.go
+++ b/x/dex/keeper/fractional_banker.go
@@ -1,6 +1,7 @@
 package keeper
 
 import (
+	"sort"
 	"strings"
 
 	math "cosmossdk.io/math"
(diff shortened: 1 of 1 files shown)


jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/18854848695

x/dex/keeper/fractional_banker.go
```diff
@@ -72,7 +73,14 @@ func (k *FractionalBanker) GetAllFractionalBalances(ctx sdk.Context) (types.Prec
...
 	store := prefix.NewStore(ctx.KVStore(k.storeKey), types.KeyPrefix(types.FractionalBalanceKeyPrefix))
-	for denom, amount := range balances {
+	sortedBalances := make([]string, 0, len(balances))
+	for denom := range balances {
+		sortedBalances = append(sortedBalances, denom)
+	}
+	sort.Strings(sortedBalances)
+
+	for _, denom := range sortedBalances {
+		amount := balances[denom]
```
mmikev, line 83: Is it possible to get rid of the map completely? Because it's not the first time when we are thinking (or just missing during the review) that working with maps are safe and it's not the last time when a map shoots us in the foot.

I can imagine where this `BalanceMap` will be reused somewhere and will cause the same problems

pr0n00gler (changes requested):

x/dex/keeper/fractional_banker.go
mmikev: There will be a map linter instead for now: https://github.com/neutron-org/neutron/pull/990

pr0n00gler (approved):

swelf19 (approved):

PR #990, opened 2025-10-28
Feat: map iteration linter

jcompagni10:


pr0n00gler:
Could you add this to CI as well, so it would be run on each PR?

internal/analyzers/mapiter/analyzer.go
```diff
@@ -0,0 +1,57 @@
...
+		return nil, nil
+	},
+}
+
+func hasIgnoreComment(pass *analysis.Pass, file *ast.File, rangeStmt *ast.RangeStmt) bool {
+	forLine := pass.Fset.Position(rangeStmt.For).Line
+	for _, group := range file.Comments {
+		for _, comment := range group.List {
+			text := comment.Text
+			if !strings.Contains(text, "nolint:mapiter") && !strings.Contains(text, "mapiter:ignore") {
```
sotnikov-s: the "//nolint:something" is the golang-ci linter instruction format. because of that, there's a WARN occurring when running `make lint`:
```
make lint

golangci-lint run --fix --config .golangci.yml
>>>> WARN [runner/nolint_filter] Found unknown linters in //nolint directives: mapiter 
0 issues.
find . -name '*.go' -not -name "*.pb.go" -type f -not -path "./vendor*" -not -path "*.git*" -not -path "*_test.go" | xargs gofmt -d -s
/Library/Developer/CommandLineTools/usr/bin/make lint-mapiter
go run ./internal/analyzers/mapiter -- ./...
```

why even declaring `"nolint:mapiter"` here as a possible ignore instruction? why don't just keep the `"mapiter:ignore"`?

PR #991, opened 2025-10-28
fix price range error msg

jcompagni10:


jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/18881497087

pr0n00gler (approved):

PR #996, opened 2025-11-06
Feat: Withdraw with shares

jcompagni10:
Add a new Message to be able to perform withdraws by only passing shares and not tick data

**related**
https://github.com/neutron-org/neutron-std/pull/19
https://github.com/neutron-org/neutron-test-tube/pull/10


x/dex/keeper/withdraw.go
```diff
@@ -77,28 +111,22 @@ func (k Keeper) ExecuteWithdraw(
...
 		poolDenom := pool.GetPoolDenom()
 
-		// TODO: this is a bit hacky. Since it is possible to have multiple withdrawals from the same pool we have to artificially update the bank balance
-		// In the future we should enforce only one withdraw operation per pool in the message validation
-		alreadyWithdrawnOfDenom := coinsToBurn.AmountOf(poolDenom)
-		sharesOwned := k.bankKeeper.GetBalance(ctx, callerAddr, poolDenom).Amount.Sub(alreadyWithdrawnOfDenom)
+		sharesOwned := k.bankKeeper.GetBalance(ctx, callerAddr, poolDenom).Amount
+
+		fmt.Printf("sharesOwned: %s\n", sharesOwned)
+		fmt.Printf("sharesToRemove: %s\n", sharesToRemove)
```
mmikev: I guess these are some leftover logs from debugging stage, please remove them

x/dex/types/errors.go
```diff
@@ -262,4 +262,14 @@ var (
...
 	)
+	ErrDuplicatePoolWithdraw = sdkerrors.Register(
+		ModuleName,
+		1172,
+		"Cannot withdraw from the same pool multiple times:", // tickIndex: "%d", fee: "%d"
+	)
+	ErrCanOnlyWithdrawFromSamePair = sdkerrors.Register(
+		ModuleName,
+		1173,
+		"Can only withdraw from the same pair:", // poolId: "%d", pairID: "%s"
```
mmikev: I think ':' and formatting (which is implied by the comment) are redundant here. All the formatting must happen when the error itself is wrapped and returned in a module's logic

x/dex/keeper/msg_server_test.go
```diff
@@ -1843,176 +1861,74 @@ func TestMsgDepositValidate(t *testing.T) {
 	}
 }
 
-func TestMsgWithdrawalValidate(t *testing.T) {
+func TestMsgWithdrawalWithSharesValidate(t *testing.T) {
```
mmikev, line 2040: Why have you completely removed the test for `TestMsgWithdrawal`?

pr0n00gler (changes requested):

pr0n00gler (changes requested):
There are some conflicts and attach an integration tests run, please

x/dex/keeper/withdraw.go
```diff
@@ -136,3 +164,66 @@ func (k Keeper) ExecuteWithdraw(
...
+	tickIndicesNormalized []int64,
+	fees []uint64,
+) ([]*types.Pool, error) {
+	poolsToRemoveFrom := make([]*types.Pool, len(tickIndicesNormalized))
+	for i, tickIndex := range tickIndicesNormalized {
+		pool, found := k.GetPool(ctx, pairID, tickIndex, fees[i])
+		if !found {
+			return nil, sdkerrors.Wrapf(
+				types.ErrInsufficientShares,
+				"User does not have shares for in Pool tickIndex %d, fee %d",
```
swelf19: The error sounds like pool exists but user does not have shares, that is not  the case, right?
The same below in `SharesToPools` method

x/dex/types/message_withdrawal_with_shares.go
```diff
@@ -0,0 +1,78 @@
...
+	for _, share := range msg.SharesToRemove {
+
+		if !share.Amount.IsPositive() {
+			return sdkerrors.Wrapf(ErrZeroWithdraw, "share amount is zero for share %s", share.Denom)
+		}
+
+		if err := ValidatePoolDenom(share.Denom); err != nil {
+			return sdkerrors.Wrapf(ErrInvalidPoolDenom, "invalid share denom (%s)", err)
+		}
+		if _, ok := poolsSeen[share.Denom]; ok {
```
zavgorodnii, line 71: why does it check for denom here, but for denom+fee in message_withdrawal.go?

poolStr := fmt.Sprintf("%d-%d", tickIndex, fee)
		if _, ok := poolsSeen[poolStr]; ok {

x/dex/keeper/grpc_query_simulate_withdrawal.go
```diff
@@ -30,14 +30,63 @@ func (k Keeper) SimulateWithdrawal(
...
+	}
+
+	callerAddr := sdk.MustAccAddressFromBech32(msg.Creator)
+	receiverAddr := sdk.MustAccAddressFromBech32(msg.Receiver)
+	poolsToRemoveFrom, shareAmountsToRemove, err := k.SharesToPools(ctx, msg.SharesToRemove)
+	if err != nil {
+		return nil, err
+	}
+
+	pairID := poolsToRemoveFrom[0].MustPairID()
```
zavgorodnii, line 86: is this safe?

x/dex/types/message_withdrawal_with_shares.go
jcompagni10: Not totally sure I understand the question. The goal is to ensure that there are no duplicate deposits. But if you withdraw with tickindex + pair + fee you can't get the denom without doing a kvstore lookup. We don't want to do a lookup at validation stage so we use tickIndex+fee. The result is the same.

x/dex/keeper/grpc_query_simulate_withdrawal.go
jcompagni10: There is no risk of it throwing because we just retrieved the pools from the kvstore. 
If the concern is that we are assuming all pools will have the same pairID this is explicity checked for in `SharesToPools`

jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/19553071138

x/dex/types/message_withdrawal_with_shares.go
swelf19: What is the reason dont do lookup at validation?
Since messages no longer need to have a `Validate` method, the `HasValidate` interface is deprecated, and it’s needed to call Validate explicitly. Because of this, it seems we can call `Validate(ctx)`.
I’m not sure how good a practice it is.

jcompagni10: Maybe I am operating off of outdated cosmos sdk standards. But given that the end result is the same it still seems cleaner to separate stateless and stateful validation. Not sure what advantage we get out of doing an actual lookup if we don't have to

pr0n00gler (changes requested):
Regenerate swagger schema as well, please

swelf19 (dismissed):

pr0n00gler (dismissed):

swelf19 (approved):

pr0n00gler (approved):

PR #1029, opened 2026-01-21
feat: enabled pebble build flag, add snapshot subcommand

swelf19:
PR adds:
- added `snapshot` subcommand to manage local snapshots
- enabled pebble db. Now you can run you node with pebbledb backend by setting `db_backend` field in the config.toml to `pebbledb`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,6 +51,7 @@ ifeq ($(WITH_CLEVELDB),yes)
   build_tags += gcc
 endif
 build_tags += $(BUILD_TAGS)
+build_tags += pebbledb
 build_tags := $(strip $(build_tags))
 
 build_tags_test_binary = $(build_tags)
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -51,6 +51,7 @@ ifeq ($(WITH_CLEVELDB),yes)
   build_tags += gcc
 endif
 build_tags += $(BUILD_TAGS)
+build_tags += pebbledb
```
mmikev, line 54: I guess we need to add this only some kinda flag? We don't want all binaries to be built with pebble

keylogic1: Build tags enable capability, config.toml or runtime flags control behavior. 

Adding the tag just registers pebbledb as an option to use but the default is still goleveldb unless someone explicitly changes `db_backend` in their config/runtime.

Hence it's safe to simply have one binary with these capabilities enabled.

mmikev: Ah, ok

pr0n00gler (approved):

keylogic1 (approved):

PR #1057, opened 2026-03-20
Chore: Improve Accuracy of PriceToTick Calculation

jcompagni10:
Fix the PriceToTickCalculation to ensure that the closest tick is always returned. Current code sometimes has an off-by-one error.


jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/23357641580

swelf19 (approved):

PR #1081, opened 2026-05-07
feat: fix tranche key ordering

### LINKED ISSUE #1072: issue about a newer expiring order can be filled before an older one at the same price because the DEX compares tranche IDs as plain text.
the problem here is that when two expiring orders are sitting at the same price, then the older one should be filled first. But here that can fail because the DEX is sorting tranche IDs as plain text, not by real order age. So a newer tranche like tk-10 can be treated as coming before an older one like tk-2. and the result is simple: a later attacker order can cut in line, and get the fill first, and take the taker-side proceeds, while the earlier user is left without the execution they should have received

sotnikov-s:
# this PR:
- resolves https://github.com/neutron-org/neutron/issues/1072 by using Uint64ToSortableString func on tranche key indices: tk-100500 becomes tk-Uint64ToSortableString(100500). this ensures proper lexicographic ordering
- adds dex storage migration for limit order tranches, inactive limit order tranches, limit order tranche users, and limit order expirations. affected storage leafs are the ones that contain "tk-" prefix.

# how to verify this code:
https://www.notion.so/hadron/dex-tk-0-migration-verification-35f85d6b9b1080e28044f4d4f793a285?source=copy_link#36e85d6b9b1080a996c4ef23fe9c7832


app/upgrades/nextupgrade/upgrades.go
```diff
@@ -0,0 +1,220 @@
...
+	}
+
+	if err := reconstructLoTrancheUserLists(ctx, k); err != nil {
+		return fmt.Errorf("failed to reconstruct LO tranche user lists: %w", err)
+	}
+
+	return nil
+}
+
+func reconstructLoExpirations(ctx sdk.Context, k dexkeeper.Keeper) error {
```
jcompagni10: Why does this need to be a separate process? Why can't we just iterate through all limit order tranches and then check the tranche directly if it is expiring?

x/dex/types/limit_order_tranche_key.go
```diff
@@ -42,3 +44,8 @@ func (p LimitOrderTrancheKey) MustPrice() (priceTakerToMaker math_utils.PrecDec)
 	}
 	return price
 }
+
+// NewTrancheKey returns a new tranche key based on the tranche index.
+func NewTrancheKey(trancheIdx uint64) string {
+	return fmt.Sprintf("tk-%020d", trancheIdx)
```
jcompagni10: Probably not a big deal, but why not use `Uint64ToSortableString`? It's a bit more space-efficient. Also, from a UX perspective, trying to count/type a bunch of leading zeros is a PITA.

jcompagni10 (dismissed):

app/upgrades/nextupgrade/upgrades.go
sotnikov-s: I found explicit per-storage-branch functions easier to understand and safer: we iterate over the whole bunch of leafs at a time and leave no room for a mistake/misunderstanding. I could do as you suggested but in my opinion it would be a more complex migration code with no perceptible practical benefits like speed or efficiency

x/dex/types/limit_order_tranche_key.go
sotnikov-s: no strong point against this approach, I also considered it, but preferred the zero-padded format for easier testing and debugging. re UX perspective: 1) I don't think anyone (but me and you in scope of this PR haha) would ever count the zeroes since it's easier and more useful to count the non-zero digits 2) if someone ever needs to get to know their tranche index, would it be easier for them to parse the base36 number prefixed with length?

sotnikov-s: space efficiency is a good point. you suggest to use `tk-Uint64ToSortableString(trancheIdx)` instead of `tk-020%d`, right? do you guys think it worths refactoring?

swelf19: agree Uint64ToSortableString function, I would even get rid of the `tk` prefix - it doesnt seem to carry any useful information

swelf19:
It seems to me that this kind of migration should be performed in a numbered module migration rather than in an upgrade handler, just like any storage migration required for the correct operation of the new version of the code.

x/dex/types/limit_order_tranche_key.go
sotnikov-s: moved from zero padded tranche keys to Uint64ToSortableString usage. didn't remove the tk- prefix since it ruined proper lexicographical ordering: there are obsolete keys build with height + block gas usage that also have that base36 format build with Uint64ToSortableString and they all start with 5 as well as the tk-plain-decimal keys do. e.g. there is a tranche user record with tranche key 5ak8d741gqz, and after migration a typical tranche key reconstructed from plain decimal looks like 52ojj9. the former key is older than the latter whereas lexicographic sorting puts the latter one before the former, and this is incorrect and it should be mitigated: older tranche keys must be put before newer ones in sorting/iterating.

also addressed this comment:

> It seems to me that this kind of migration should be performed in a numbered module migration rather than in an upgrade handler, just like any storage migration required for the correct operation of the new version of the code.

read about the results and dex module exported states here https://www.notion.so/hadron/dex-tk-0-migration-verification-35f85d6b9b1080e28044f4d4f793a285?source=copy_link#36e85d6b9b1080a996c4ef23fe9c7832

jcompagni10 (dismissed):

jcompagni10:
This all looks good. Maybe add a test to confirm that the 3 types of tranches keys (sortableString, tk-[n], and tk-[sortable-string] all get stored in the correct order. I think it is correct. But a test would be nice.

swelf19 (dismissed):
Am I understanding correctly that these are breaking changes? If an application stored the tk when it was created, it will unexpectedly receive an error when trying to work with the old tk later?
Does it make sense to add a note or comment about this in the code and mention it in a future release?

TrancheUserKey has a back-link to tk. It seems those need to be updated as well?

sotnikov-s:
> This all looks good. Maybe add a test to confirm that the 3 types of tranches keys (sortableString, tk-[n], and tk-[sortable-string] all get stored in the correct order. I think it is correct. But a test would be nice.

done here: https://github.com/neutron-org/neutron/pull/1081/changes/9c837101b3366cc6d12eeb877254088d98d1fb4b

sotnikov-s:
> TrancheUserKey has a back-link to tk. It seems those need to be updated as well?

doesn't this part of the migration address this? https://github.com/neutron-org/neutron/blob/9c837101b3366cc6d12eeb877254088d98d1fb4b/x/dex/migrations/v9/store.go#L195-L229

sotnikov-s:
> Am I understanding correctly that these are breaking changes? If an application stored the tk when it was created, it will unexpectedly receive an error when trying to work with the old tk later? Does it make sense to add a note or comment about this in the code and mention it in a future release?

I guess so: some msgs and queries have tranche key as a parameter, so if e.g. a smart contract memorised a plain decimal tranche key e.g. tk-100 and it then tries to do something with that tranche using that stored key, it won't work because the tranche key will become tk-22S after migration. don't think it makes sense adding a comment in a code but a note in a future release makes sense to me

sotnikov-s:
https://github.com/neutron-org/neutron-tests/actions/runs/26770936506

swelf19 (approved):

